PR #60, opened 2024-09-11
adds more tests for flexarray etc.

ringabout:


tests/nifc/hello.nif
```diff
@@ -37,11 +37,67 @@
...
+
+  (type :MyObject.ptr.object .
+  (object . (fld :a1 . (i +32 atomic))
+    (fld :a2 . (i +64 atomic))
+    (fld :a3 . (bool .))
+    (fld :a4 . (ptr MyObject3.m .))
+  )
+  )
+
+  (type :MyObject.sequence.base .
```
Araq, line 82: Make these test cases harder so that things are used before they are defined.

PR #70, opened 2024-09-17
implements linedir (#line) for debugging

ringabout:
generates `#line 5 "test.nim"`


Araq:
Nah, sorry, you missed the point entirely. NIF itself already has the line information in its AST. You can just query it and produce it before every function call and assignment statement.

ringabout:
Okay, I will generate `#line` from LineInfo instead

PR #84, opened 2024-09-23
prints stacktraces when built with `-d:debug`

ringabout:


src/nifc/codegen.nim
```diff
@@ -134,6 +134,8 @@ proc error(m: Module; msg: string; tree: PackedTree[NifcKind]; n: NodePos) =
   write stdout, "[Error] "
   write stdout, msg
   writeLine stdout, toString(tree, n, m)
+  when defined(nifcdebug):
```
Araq, line 137: Other tools like c2nim simply use `defined(debug)` so that's what we should do here too.

ringabout, line 137:
```suggestion
  when defined(debug):
```

ringabout: okay

PR #104, opened 2024-10-16
implements `InfC, NegInfC, NanC` for nifc

ringabout:


Araq:
Yep, that's how to do it.

PR #113, opened 2024-11-09
nimony: progress

Araq:


src/gear3/expander.nim
```diff
@@ -230,6 +229,7 @@ proc traverseParams(e: var EContext; c: var Cursor) =
       if c.substructureKind != ParamS:
         error e, "expected (param) but got: ", c
       traverseLocal(e, c, "param", TraverseSig)
+  traverseType e, c
```
ringabout, line 232: This is the output from nifler. ` (param :x . . (i +32) .)` has already handled by `traverseLocal`.
```
(params
   (param :x . . (i +32) .)
)
```
What's the `traverseType` call here for?

PR #118, opened 2024-11-14
improve gear3

ringabout:
- [ ] Types
- [x] Cases
- [x] Discard
- [x] Import Procs 
- [ ] etc.

### DIFF
--- a/src/gear3/expander.nim
+++ b/src/gear3/expander.nim
@@ -402,7 +402,10 @@ proc traverseExpr(e: var EContext; c: var Cursor) =
     case c.kind
     of EofToken: break
     of ParLe:
-      e.dest.add c
+      if pool.tags[c.tag] == "curlyConstr":
+        e.add "ranges", c.info
+      else:
+        e.dest.add c
       inc nested
     of ParRi: # TODO: refactoring: take the whole statement into consideration
       if nested == 0:
@@ -553,7 +556,6 @@ proc traverseCase(e: var EContext; c: var Cursor) =
       wantParRi e, c
     else:
       error e, "expected (of) or (else) but got: ", c
-  traverseStmt e, c
   wantParRi e, c
 
 proc traverseStmt(e: var EContext; c: var Cursor; mode = TraverseAll) =
--- a/src/gear3/expander.nim
+++ b/src/gear3/expander.nim
@@ -103,6 +103,18 @@ proc expectSymdef(e: var EContext; c: var Cursor) =
   if c.kind != SymbolDef:
     error e, "expected symbol definition, but got: ", c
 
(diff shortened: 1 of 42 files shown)


Araq:
The most important feature of gear3 is the graph expansion btw, a NIF module can refer to symbols in a different NIF module without any explicit `imp` (import)  declaration but NIFC does not support that, so gear3 needs to copy things into the current NIFC module.

Please write test cases for this!

tests/gear3/gear3_helloworld.nif
```diff
@@ -4,11 +4,12 @@
   (var :x . . (i +32) +12)
   (var :y . . (i +32) +12)
 
+  (import mod1)
```
Araq, line 7: Not required in gear3! You can simply use `foo.c12.mod1` later and then it is picked up from `mod1`.

ringabout: I see. I added code to skip `import` because the result from `nifler parse` retains the `import mod1`.

Done

src/gear3/expander.nim
```diff
@@ -103,6 +103,18 @@ proc expectSymdef(e: var EContext; c: var Cursor) =
   if c.kind != SymbolDef:
     error e, "expected symbol definition, but got: ", c
 
+proc getSymDef(e: var EContext; c: var Cursor): (SymId, PackedLineInfo) =
+  let isQuoted = c.kind == ParLe and pool.tags[c.tag] == "quoted"
```
Araq, line 107: Gear3 does not accept `quoted` as it won't be emitted by the frontend.

src/gear3/expander.nim
```diff
@@ -402,8 +409,21 @@ proc traverseExpr(e: var EContext; c: var Cursor) =
     case c.kind
     of EofToken: break
     of ParLe:
-      e.dest.add c
-      inc nested
+      if pool.tags[c.tag] == "curlyConstr":
```
Araq, line 412: "curlyConstr" is not in gear3's input language either. (I need to write a spec...)

src/gear3/expander.nim
```diff
@@ -402,8 +409,21 @@ proc traverseExpr(e: var EContext; c: var Cursor) =
...
+        e.add "ranges", c.info
+        inc nested
+      elif pool.tags[c.tag] == "quoted":
```
Araq, line 415: Likewise this cannot happen.

Araq:
See my remarks, the rest is awesome.

src/gear3/expander.nim
ringabout: Sure, I will assume the input to be `ranges`

PR #120, opened 2024-11-14
nifc: add alignof & offsetof, make sizeof only take type

metagn:
Unlike `sizeof`, `alignof` and `offsetof` only expect a type in C, so `sizeof` also now accepts only types instead of only expressions.


Araq:
I think it's ok to change the spec so that `sizeof` always and only takes a type. `sizeof(value)` makes no sense in an optimized-for-codegens language. :-)

metagn:
Done

PR #124, opened 2024-11-18
improves gear3

ringabout:
- [x] import vars/consts
- [x] importc (printf, assert)
- [ ] Types

### DIFF
--- a/src/nifc/codegen.nim
+++ b/src/nifc/codegen.nim
@@ -82,7 +82,6 @@ type
     headerFile: seq[Token]
     generatedTypes: IntSet
     requestedSyms: HashSet[string]
-    toExtern: bool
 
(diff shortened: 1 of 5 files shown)


src/nifc/codegen.nim
```diff
@@ -82,6 +82,7 @@ type
     headerFile: seq[Token]
     generatedTypes: IntSet
     requestedSyms: HashSet[string]
+    toExtern: bool
```
Araq, line 85: Instead of this solution, why not call `genVar` instead of `genStmt`?

ringabout: Sure, I will extract them into a `genVar` function to avoid duplication and pass `toExtern` by parameters

PR #138, opened 2024-11-27
improves gear3 with atom expression

ringabout:


Araq:
`traverseExpr` will get more custom logic soon anyway so this change doesn't matter.

ringabout:
It was an expedient that I used
```nim
if nested == 0:
  break
```
to identify an atom expression without parenthesis. Probably it had some side effects for malformed input.

PR #148, opened 2024-11-29
nimony: implements blocks

ringabout:


src/nimony/sem.nim
```diff
@@ -1790,6 +1815,8 @@ proc semExprSym(c: var SemContext; it: var Item; s: Sym; flags: set[SemFlag]) =
         skipToLocalType n
       elif s.kind.isRoutine:
         skipToParams n
+      elif s.kind == LabelY:
+        skip n # skip label
```
Araq, line 1819: dead code, instead use `discard` here.

PR #151, opened 2024-11-29
Nimony: Default float type should be -1

SirOlaf:


Araq:
I think in Nim `float` is always mapped to `float64` so `-1` would be an incompatibility.

SirOlaf:
Thought it would be decided by the backend, but it does look like NF at least is mapped to `double`

SirOlaf:
Also for context, literals get assigned -1 so assigning a literal to float32 variable doesn't match https://github.com/nim-lang/nif/blob/master/src/nimony/builtintypes.nim#L39-L40

Araq:
Integers are not floats, `int` is its own type and `float` is merely an alias.

SirOlaf:
I mean like `let x: float32 = 1.0` which currently has mismatched types because literals are -1 and variables get 64, but if it is supposed to be like that the issue is in type checks

Araq:
The type checking is WIP like generics, sets, case statements ... ;-)

PR #159, opened 2024-12-03
[wip] implements lowerer

ringabout:
- [ ] test against nimony output
- [ ] improve code shapes


src/lowerer/transformer.nim
```diff
@@ -0,0 +1,497 @@
...
+    block forStmtLabel:
+      let a = 1
+      let b = sideEffect()
+      var x = 0
+      while x <= b:
+        block inner:
+          loopBodyStart()
+          use x
+          inc x
+          if condA:
```
blackmius, line 417: it looks like inc x should be after block inner

blackmius: but then how multiple yield will work?
it should dublicate block inner between yields?

what such iterator will look like?

```
iterator cross(a, b: int): tuple[int, int] =
  var i = 0
  while i <= a:
    var j = 0
    while j <= b:
      yield (i, j)
      inc j
    inc i
```

blackmius: and interesting to see multiple yield iterators

```
iterator powers(n: int): int =
  var i = 0
  while i <= n:
    yield i
    yield i*i
    yield i*i*i
    inc i
```

ringabout: > it looks like inc x should be after block inner

fixed in https://github.com/nim-lang/nimony/pull/349

> and interesting to see multiple yield iterators

I have added it to tests. It works fine.

ringabout:
Succeeded by https://github.com/nim-lang/nimony/pull/349

PR #164, opened 2024-12-03
nifc: use _PTR macros for proc type call conventions

metagn:
This is what Nim uses, it might cause issues when adding attributes (currently it does `typedef void (*__attribute(foo)__ bar)` unilke no call convention which does `typedef void (__attribute__(foo) *bar)`) but this can be addressed when attributes are used for proc types, I can't find an example of either version working.

### DIFF
--- a/src/nifc/codegen.nim
+++ b/src/nifc/codegen.nim
@@ -179,6 +179,19 @@ proc genUIntLit(c: var GeneratedCode; litId: StrId) =
 
 # Type graph
 
+const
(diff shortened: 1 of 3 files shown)


Araq:
Tests are red...

metagn:
Changed to use `_PTR` macros same as Nim, this changes the `__attribute__` placement but I can't tell what the correct placement for `__attribute__` in proc typedefs is.

PR #171, opened 2024-12-06
gear3: implements enumT

ringabout:
implements `enumT`: adds an extra parameter `dest: var TokenBuf` for `traverseExpr`, `loop`, `traverseType` etc. Because we need to evaluate fields in advance for later use to get the enum types, the result of fields need to be stored in a temp `TokenBuf`.

### DIFF
--- a/src/gear3/expander.nim
+++ b/src/gear3/expander.nim
@@ -176,6 +176,28 @@ proc traverseTupleField(e: var EContext; c: var Cursor; flags: set[TypeFlag] = {
   inc c # skips value
   wantParRi e, c
 
+proc traverseEnumField(e: var EContext; c: var Cursor; flags: set[TypeFlag] = {}): TokenBuf =
+  e.dest.add c # efld
+  inc c
+
+  expectSymdef(e, c)
+  let (s, sinfo) = getSymDef(e, c)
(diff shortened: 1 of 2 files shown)


Araq:
> Because we need to evaluate fields in advance for later use to get the enum types, the result of fields need to be stored in a temp TokenBuf.

Nah, learn about the `swap tmp, c.dest` idiom.

ringabout:
I see, I didn't notice it

src/gear3/expander.nim
```diff
@@ -176,6 +176,28 @@ proc traverseTupleField(e: var EContext; c: var Cursor; flags: set[TypeFlag] = {
...
+  inc c
+
+  expectSymdef(e, c)
+  let (s, sinfo) = getSymDef(e, c)
+  e.dest.add toToken(SymbolDef, s, sinfo)
+  e.offer s
+
+  skipExportMarker e, c
+
+  inc c # pragmas: must be empty
```
Araq, line 190: At least add an `assert c.kind == DotToken` here or better use `skip c` which cannot crash.

Araq, line 190:
```suggestion
  skip c # pragmas: must be empty
```

PR #173, opened 2024-12-07
nimony: `if` expressions, make commonType work with noreturn

metagn:
`semStmtBranch` should also work for `case`/`try` as well

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -708,6 +708,114 @@ proc combineType(c: var SemContext; info: PackedLineInfo; dest: var Cursor; src:
   else:
     c.typeMismatch info, src, dest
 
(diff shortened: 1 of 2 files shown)


src/nimony/sem.nim
```diff
@@ -708,6 +708,114 @@ proc combineType(c: var SemContext; info: PackedLineInfo; dest: var Cursor; src:
...
+  #      checkBranch(branch[^1])
+  #  # all branches are discardable
+  #  result = true
+  of CallS, CmdS:
+    inc it
+    if it.kind == Symbol:
+      let sym = tryLoadSym(it.symId)
+      if sym.status == LacksNothing:
+        var decl = sym.decl
+        if isRoutine(symKind(decl)):
```
Araq, line 793: btw learn about decls.nim

PR #174, opened 2024-12-08
nimony: tuple constructors

metagn:
Typedesc tuples aren't implemented (i.e. `(int, string)`), they could be but it would hinge on type expressions being accepted as regular expressions, which is not the case currently but probably should be eventually. Then we would check for `TypedescT`.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -2429,7 +2429,7 @@ proc semArrayConstr(c: var SemContext, it: var Item) =
   c.dest.buildTree ArrayT, it.n.info:
     c.dest.add toToken(IntLit, count, it.n.info)
     c.dest.addSubtree elem.typ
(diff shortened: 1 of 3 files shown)


Araq:
`combineType` needs to be replaced by `commonType` almost everywhere.

src/nimony/sem.nim
```diff
@@ -2498,6 +2498,58 @@ proc semSuf(c: var SemContext, it: var Item) =
...
+      wantParRi c, it.n
+    typ.addSubtree elem.typ # type
+    typ.addDotToken() # value
+    typ.addParRi() # end field
+  wantParRi c, it.n
+  typ.addParRi()
+  let start = c.dest.len
+  var t = typ.cursorAt(0)
+  semTupleType(c, t)
+  combineType c, it.n.info, it.typ, typeToCursor(c, start)
```
Araq, line 2550: You need `commonType` here.

PR #182, opened 2024-12-10
nimony: allow any type aliases in sigmatch

metagn:
`asTypeAlias` only allows type symbols with the implementation either being other type symbols or a generic invocation, this needed tuple types to be special cased in #179. However this will need other types to be whitelisted as well, so the restriction is removed. Maybe there is still a reason why these were whitelisted though?


Araq:
That's better but the initial idea was to resolve type aliases early on in semLocalType.

PR #187, opened 2024-12-10
nimony: expression evaluator for simple constant expressions

metagn:
This acts as a primitive version of the compile time VM so that we can handle the most common constant expressions in Nim code. I had thought `defined`/`declared` would also be evaluated here, but currently they are immediately resolved in semchecking, which could be sufficient overall including for things like `sizeof` (though not when they depend on an unresolved generic parameter i.e. `sizeof(T)`, but this would be resolved in semchecking during instantiation too).

Might be ugly and low on comments, couldn't think of how to clean it up. `semConstBoolExpr` is unused for now but will likely be used.

### DIFF
--- a/src/nimony/expreval.nim
+++ b/src/nimony/expreval.nim
@@ -1,40 +1,48 @@
 ## expression evaluator for simple constant expressions, not meant to be complete
-## 
-## included in sem.nim
 
-type EvalContext = object
-  c: ptr SemContext # used the same way as `var SemContext`
+include nifprelude
+import nimony_model, decls, programs, sigmatch # sigmatch just for addStrLit
+
+type EvalContext* = object
   values: seq[TokenBuf]
   trueValue, falseValue: Cursor
 
-proc isConstBoolValue(n: Cursor): bool =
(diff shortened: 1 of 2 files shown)


Araq:
This will inevitably grow to a point where it becomes both annoying&slow and at the same time hard to replace. We need some upfront design here. We could generate NIFC and then add an `eval` feature to NIFC. This requires a new component that can do:

1. Map NIF tokens to a blob.
2. Map blob (plus a type descriptor) back to NIF tokens.

This is much more work but inevitably we need this for macros/plugin support anyway.

metagn:
The idea was that this would be a placeholder VM *until* we can generate NIFC from semchecked code, so that we can typecheck things like `when defined(foo) and not defined(bar)` beforehand. It's not meant to grow but I guess this is hard to guarantee, I thought not being feature complete would force us to replace it. We could maybe make it opt in just for testing.

The mapping sounds like it could just be serialization of NIFC types between NIF and their native values. So if we don't want to depend on a specific native representation, we could also generate both serializers on the backend and just pass and receive NIF. I don't know if this would be harder.

PR #192, opened 2024-12-11
gear3: fixes arrayT

ringabout:
`nimony` gives `(array +3 (i -1))` for array types: `traverseExpr` -> `traverseType`


Araq:
Nimony is wrong though, I think. It should use `(array Type Expr)`.

PR #194, opened 2024-12-12
nimony: subscripts sans generic instantiations

metagn:
Code is messy for now, will move out the bugfix `addParLe`s and do the `ArrayT` reorder as mentioned in #192 first. Also a bug that `let arrIndex1: int = x[0]` gives `got: (i -1) but wanted: (u -1)`

### DIFF
--- a/src/nimony/expreval.nim
+++ b/src/nimony/expreval.nim
@@ -1,7 +1,7 @@
 ## expression evaluator for simple constant expressions, not meant to be complete
 
 include nifprelude
-import nimony_model, decls, programs, sigmatch # sigmatch just for addStrLit
+import nimony_model, decls, programs, sigmatch # sigmatch just for addStrLit and addParLe
 
 type EvalContext* = object
   values: seq[TokenBuf]
@@ -43,7 +43,7 @@ proc getTrueValue(c: var EvalContext): Cursor =
   if c.trueValue == default(Cursor):
     let i = c.values.len
     c.values.add createTokenBuf(2)
-    c.values[i].add toToken(ParLe, TrueX, NoLineInfo)
+    c.values[i].addParLe(TrueX, NoLineInfo)
     c.values[i].addParRi()
     c.trueValue = cursorAt(c.values[i], 0)
   result = c.trueValue
@@ -52,7 +52,7 @@ proc getFalseValue(c: var EvalContext): Cursor =
   if c.falseValue == default(Cursor):
     let i = c.values.len
     c.values.add createTokenBuf(2)
-    c.values[i].add toToken(ParLe, FalseX, NoLineInfo)
+    c.values[i].addParLe(FalseX, NoLineInfo)
     c.values[i].addParRi()
     c.falseValue = cursorAt(c.values[i], 0)
   result = c.falseValue
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -2339,7 +2339,7 @@ proc semArrayConstr(c: var SemContext, it: var Item) =
   wantParRi c, it.n
   let typeStart = c.dest.len
   c.dest.buildTree ArrayT, it.n.info:
-    c.dest.add toToken(IntLit, count, it.n.info)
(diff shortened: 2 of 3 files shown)


Araq:
I would have done it this way:

```nim

proc buildReadAccessor(c: var CheckContext; dest: var Tree; tree: Tree, n: NodePos, acc: string) =
  copyIntoKind(dest, Call, n.info):
    let accessor = c.m.strings.getOrIncl(acc)
    buildSymChoice(c, dest, accessor, n.info)
    for ch0 in sonsReadonly(tree, n):
      copyTree(dest, tree, ch0)

proc semBracketExpr(c: var SemContext; dest: var Tree; tree: Tree; n: NodePos;
                    preferredType: FullTypeId): FullTypeId =
  var tmp = createTempTree(c.thisModule)
  discard sem(c, tmp, tree, n.firstSon, autoType)
  var routines: seq[FullSymId] = @[]
  if isTypeExpr(c, tmp, StartPos):
    rollbackTo c.p, tmp, StartPos
    result = borrowPosition(c, dest)
    semDeclType(c, dest, tree, n, inLocalDecl)
  elif isGenericRoutine(c, tmp, StartPos, routines):
    rollbackTo c.p, tmp, StartPos
    result = semExplicitGenericRoutineInst(c, dest, tree, n, routines)
  else:
    rollbackTo c.p, tmp, StartPos
    buildReadAccessor(c, tmp, tree, n, "[]")
    result = sem(c, dest, tmp, StartPos, preferredType)

proc buildWriteAccessor(c: var CheckContext; dest: var Tree; tree: Tree; n: NodePos, acc: string) =
  copyIntoKind(dest, Call, n.info):
    let accessor = c.m.strings.getOrIncl(acc)
    buildSymChoice(c, dest, accessor, n.info)
    let (le, ri) = sons2(tree, n)
    # flatten the 'BracketExpr' away:
    for ch1 in sonsReadonly(tree, le):
      copyTree(dest, tree, ch1)
    # but not the 'value' of 'a[i] = value':
    copyTree(dest, tree, ri)


  of AsgnS:
    case n.firstSon.kind
    of BracketExpr:
      var tmp = createTempTree(c.thisModule)
      buildWriteAccessor(c, tmp, tree, n, "[]=")
      result = sem(c, dest, tmp, StartPos, preferredType)
    of CurlyExpr:
      var tmp = createTempTree(c.thisModule)
      buildWriteAccessor(c, tmp, tree, n, "{}=")
      result = sem(c, dest, tmp, StartPos, preferredType)
    else:
      semAsgn(c, dest, tree, n)

```

Araq:
And then the system.nim adds [] and []= etc for arrays which are mapped to the magic: mArrPut etc.

metagn:
This is ported from the old compiler, I looked into moving `[]` into overloads with it and there were some incompatibilities as outlined [here](https://github.com/nim-lang/Nim/issues/22970#issuecomment-2452654268) but nothing that prevents it from working. Tuples would still have to be special cased I'm guessing.

So `AtX` would be the parsed expression tag, and then gets semchecked into `ArrGetX`/`ArrPutX`/maybe `StrGetX`/`StrPutX` etc? Does `PatX` still exist then?

Araq:
Alright, your approach wins then.

metagn:
I wasn't disagreeing with your approach, the incompatibilities are fine imo (namely arbitrary pointer dereference & lax index int type). The backends will probably still need the different magics (except maybe the assign ones), NIFC wouldn't correctly handle `AtX` with a string operand. Another option would be to change the parsed tag to something other than `AtX`.

Araq:
> NIFC wouldn't correctly handle AtX with a string operand. 

Sure but it's not its job anyway. The lowerer translates it to `(at (dot stringObject s +0) i)` or something like that, you get the idea.

> Another option would be to change the parsed tag to something other than AtX.

Well we need more than just `AtX`, `AtX` is nifler's way of saying "a[i]", we should have our specialized variants indeed.

Araq:
It would be sad to lose this so we should move it to a `semcompat.nim` file that will do things the Nim v2 way. Prepass for generics & templates, etc yada yada.

PR #199, opened 2024-12-12
nimony: patches types of enum fields 

ringabout:
patches types of enum fields instead of using enum identifier

before:

```
(efld ~8 :false.0.tesg7afhq1 . . :bool.0.tesg7afhq1 +0)
```

after:

```
(efld ~8 :false.0.tesg7afhq1 . . (i -1) +0)
```


Araq:
No, the enum's field type is always the enum itself in order to simplify type checking. The type of the value should be stored as `(suf expr T)` if necessary.

PR #200, opened 2024-12-13
nimony: make commonType always give expected type

metagn:
As mentioned in #197. The cases I can think of where the received type can subsume the expected type would be generic typeclasses or `auto`, but these aren't encountered here since typevars aren't set (`auto` is handled as a special case first). And these would probably need a separate instantiation mechanism if they were encountered. So unless we can think of other cases where this happens, it's simpler to always use the expected type as a general rule.


Araq:
Fine for now but I think sigmatch should expose which type to use properly.

PR #202, opened 2024-12-13
nimony: subscripts with invocations as special case

metagn:
As outlined [here](https://github.com/nim-lang/nif/pull/194/#issuecomment-2538356468), would also have a special case for generic routines but there is nothing to call for them yet. Anything that is not an invocation is treated as an overloaded call to `[]`, which needs to have magic overloads.

Note for the future: https://github.com/nim-lang/Nim/issues/9997, standalone instantiations of symchoices shouldn't generate a symchoice of the instantiated symbols as Nim currently does, instead it should probably be treated as an "instantiation choice" either remaining as a bracket over a symchoice or its own node kind that can later be type-fit to a specific instantiation

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -2616,6 +2616,36 @@ proc semDeclared(c: var SemContext; it: var Item) =
   it.typ = c.types.boolType
   commonType c, it, beforeExpr, expected
 
+proc semSubscript(c: var SemContext; it: var Item) =
+  var n = it.n
+  inc n # tag
+  var lhsBuf = createTokenBuf(4)
+  swap c.dest, lhsBuf
+  var lhs = Item(n: n, typ: c.types.autoType)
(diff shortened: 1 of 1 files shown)


src/nimony/sem.nim
```diff
@@ -2616,6 +2616,36 @@ proc semDeclared(c: var SemContext; it: var Item) =
...
+
+  # build call:
+  var callBuf = createTokenBuf(16)
+  callBuf.addParLe(CallX, it.n.info)
+  callBuf.add toToken(Ident, pool.strings.getOrIncl("[]"), it.n.info)
+  callBuf.add lhsBuf
+  it.n = lhs.n
+  while it.n.kind != ParRi:
+    callBuf.addSubtree it.n
+    skip it.n
```
Araq, line 2643:
```suggestion
    callBuf.takeTree it.n
```

PR #220, opened 2024-12-16
nimony: implement `when`

metagn:
The conditions are evaluated as constant expressions, if they are true or false then only the passing branch is semchecked. However if one of the conditions fail to compile, the entire `when` statement is left unresolved.

There is an important limitation in that the semchecking of the chosen branch has to be delayed until the body phase of semchecking for it to be possible to compile the condition (in the system module at least). This means symbols declared in `when` branches can't be used before they're declared. We could maybe relax this if the condition is constant over phases (and we aren't in the system module) i.e. `true` or `defined(abc)` but expressions like `declared(abc)` could go either way.

Test for `defined` is disabled until untyped args/magics after overloading work.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1740,7 +1740,7 @@ proc semEnumField(c: var SemContext; n: var Cursor; state: var EnumTypeState) =
     if explicitValue != state.thisValue:
       state.hasHole = true
       state.thisValue = explicitValue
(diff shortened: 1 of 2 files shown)


Araq:
Fine for now but in principle within a generic you must not evaluate the `when` statement directly, but after instantiation. Well it is not clear how it interacts with checked generics.

PR #221, opened 2024-12-16
add original AST to recoverable error tokens

metagn:
I'm not sure if this is the right way to do this, but the idea is that matching `untyped` arguments would remove all error tokens from semchecked AST so that it can be semchecked again in another context. For this to be possible and ideally to prevent cascading errors, the last child of error token trees is now either a dot for when it's meant to be completely removed, and any other tree if it's meant to be replaced by that tree. It's not clear if this would work, I haven't implemented untyped arguments with this yet.

### DIFF
--- a/src/gear3/expander.nim
+++ b/src/gear3/expander.nim
@@ -214,7 +214,7 @@ proc traverseType(e: var EContext; c: var Cursor; flags: set[TypeFlag] = {}) =
     inc c
   of ParLe:
     case c.typeKind
-    of NoType, OrT, AndT, NotT, TypedescT:
+    of NoType, OrT, AndT, NotT, TypedescT, UntypedT, TypedT:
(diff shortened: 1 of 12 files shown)


Araq:
We could also give "undeclared identifier" and "undeclared module.foobar" errors their own dedicated NIF tags and then `defined` can "patch" over these more easily.

metagn:
Apparently the current error generation is broken regardless, an undeclared identifier generates `abc (error "undeclared identifier")`, and since the argument cursor is to `abc` it's a complete tree by itself and the error token is completely ignored and not added to the final call arguments. So `defined` would already (unintentionally) work for undeclared identifiers, but probably not more complex cases like `defined(abc.def)` which I'll try with this approach.

Araq:
Feel free to rebase and reopen if I missed something.

PR #230, opened 2024-12-17
semcheck some magics after overloading

metagn:
Magic expressions produced by overload resolution may need to be semchecked. However this doesn't make sense for all magics, so a whitelist is used.

`semCmp` and `semTypedArithmetic` are not included in the semchecked magics but they are fixed so that semchecking those magics works regardless: The comparison magics are now typed so `semCmp` is updated accordingly, also it erroneously recursed over `it` rather than using a new item for the operands which expect a different type than the comparison expression itself (which expects `bool`). Typed arithmetic magics are also adapted to `commonType`.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -666,9 +666,9 @@ proc addFn(c: var SemContext; fn: FnCandidate; fnOrig: Cursor; args: openArray[I
       if n.kind == SymbolDef:
         inc n # skip the SymbolDef
         if n.kind == ParLe:
-          if n.exprKind in {DefinedX, DeclaredX}:
-            # magics that opt in to semchecking after overloading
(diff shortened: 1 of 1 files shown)


Araq:
> Magic expressions produced by overload resolution may need to be semchecked.

Why?

metagn:
Calls to magics do get transformed into `(magic ...)` etc rather than `(call magic ...)`, but this produced expression still needs to be evaluated for `defined`, `sizeof` etc. which would happen during semchecking. `magicsAfterOverloadResolution` does something similar in the old compiler except that magics like `defined` are folded before overloading.

metagn:
Now magics have to opt-in to be semchecked again, `unpack` gave an error for example

PR #241, opened 2024-12-19
implement `isMainModule` as magic, fix `not` and expreval

metagn:
Hastur also adds `--isMain` to the arguments for every test, maybe this could be restricted to files starting with `t`.

Also adds disabled behavior that makes `const` definitions evaluate their values.

### DIFF
--- a/src/nimony/expreval.nim
+++ b/src/nimony/expreval.nim
@@ -70,9 +70,9 @@ proc eval*(c: var EvalContext, n: var Cursor): Cursor =
   template propagateError(r: Cursor): Cursor =
     let val = r
     if val.kind == ParLe and val.tagId == ErrT:
-      val
-    else:
       return val
+    else:
+      val
   case n.kind
   of Ident:
     error "cannot evaluate undeclared ident: " & pool.strings[n.litId], n.info
--- a/src/nimony/expreval.nim
+++ b/src/nimony/expreval.nim
@@ -7,10 +7,11 @@
 ## expression evaluator for simple constant expressions, not meant to be complete
 
 include nifprelude
-import nimony_model, decls, programs, xints
+import nimony_model, decls, programs, xints, semdata
 
 type
   EvalContext* = object
+    c: ptr SemContext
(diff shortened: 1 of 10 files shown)


src/nimony/sem.nim
```diff
@@ -1386,12 +1386,23 @@ proc maybeInlineMagic(c: var SemContext; res: LoadResult) =
       inc n # skip the SymbolDef
       if n.kind == ParLe:
         # ^ export marker position has a `(`? If so, it is a magic!
-        c.dest[c.dest.len-1] = n.load
+        let exprStart = c.dest.len-1
+        c.dest[exprStart] = n.load
+        let magicKind = n.exprKind
```
Araq, line 1391: Not required.

src/nimony/sem.nim
```diff
@@ -1386,12 +1386,23 @@ proc maybeInlineMagic(c: var SemContext; res: LoadResult) =
...
         inc n
         while true:
           c.dest.add n
           if n.kind == ParRi: break
           inc n
+        if magicKind in {IsMainModuleX}:
```
Araq, line 1397: Not required. Inlining a magic doesn't have to cause its evaluation!

src/nimony/expreval.nim
```diff
@@ -70,9 +70,9 @@ proc eval*(c: var EvalContext, n: var Cursor): Cursor =
   template propagateError(r: Cursor): Cursor =
     let val = r
     if val.kind == ParLe and val.tagId == ErrT:
-      val
-    else:
       return val
+    else:
+      val
```
Araq, line 75: Teach the evalution engine here about `isMainModule`.

metagn:
Had to disable the change that made `const`s evaluate their values as `tcstring` uses a cstring conversion which is not implemented in `expreval` yet. Will do in another PR.

Araq:
> Or we could just not test for it until we have a proper VM.

Yes, we will revisit this logic when there some kind of VM.

PR #243, opened 2024-12-19
gear3: map all borrow procs to inline

ringabout:
mapping

```
(call ~3 foo.1.tesg7afhq1 1 x.2 4 x.2)
```

to

```
(dconf VarId.0.tesg7afhq1 
   (call foo.0.tesg7afhq1 
    (dconf
     (i -1) x.2) 
    (dconf 
     (i -1) x.2)))
```

# TODO

- [ ] supports external symbols: replace them after `importSymbol` is finished


Araq:
What is this? Just treat a `borrow` pragma like `InlineC`.

PR #248, opened 2024-12-20
optional original AST after error tokens, make `defined` work with typed args

### LINKED ISSUE #242: Basic cases of structured error AST in typed code
The errors currently generated by the compiler are not structured, which is a problem for magics like `defined` that are meant to work on parsed AST but receive typed arguments after overloading. For now only these cases have to be handled:

- Undeclared identifiers, generate `(err Ident ... "undeclared identifier")` instead of `Ident (err ... "undeclared identifier")`
- Dot fields and calls that fail to compile due to overload mismatches/undeclared fields/undeclared LHSes become `(err <original expression> ... <error message>)`, currently generate something like `(call) (err ... "error message")`
- Stuff like missing imports etc. need to at least be caught by `compiles`, i.e. the error token shouldn't get swallowed at some point during overload resolution due to not being in the same tree

In general this suggests a format of `(err <original expression> ... "error message")`, the original expression part could be `.` if it is not available. #221 mostly implements this except that the expression is at the end of the tree, and `untyped` is treated separately from `typed` in that the error tokens are removed where possible in favor of the original expressions, but it's enough that `untyped` acts the same as `typed` for now.

metagn:
closes #242

Similar to #221 except that `untyped` now acts the same as `typed` (which is now added), the original AST is at the start of the error tokens and not at the end, and only the errors described in #242 keep the original AST for now.

`defined` and for now `declared` manually strip all errors from the AST they receive and build strings from them, with `defined` now also supporting dotted identifiers. In the future `declared` could fully type its argument and check for specific errors instead of looking up an identifier based off of the raw AST.

`AllowUndeclared` is added that keeps identifier tokens as is instead of wrapping them in error tokens when they are undeclared. `semCall` uses this for the callee part so that we don't have to specifically strip undeclared identifier errors for now.

Unfortunately the code is not very pretty between the intermediate buffers/cursors and recursion in `takeWithoutErrors`/`getDottedIdent`, will apply any suggestions.

### DIFF
--- a/src/gear3/expander.nim
+++ b/src/gear3/expander.nim
@@ -214,7 +214,7 @@ proc traverseType(e: var EContext; c: var Cursor; flags: set[TypeFlag] = {}) =
     inc c
   of ParLe:
     case c.typeKind
-    of NoType, OrT, AndT, NotT, TypedescT, UntypedT:
+    of NoType, OrT, AndT, NotT, TypedescT, UntypedT, TypedT:
       error e, "type expected but got: ", c
     of IntT, UIntT, FloatT, CharT, BoolT, AutoT, SymKindT:
       e.loop c:
--- a/src/nimony/expreval.nim
+++ b/src/nimony/expreval.nim
@@ -40,8 +40,9 @@ proc skipParRi(n: var Cursor) =
 
 proc error(c: var EvalContext, msg: string, info: PackedLineInfo): Cursor =
   let i = c.values.len
-  c.values.add createTokenBuf(3)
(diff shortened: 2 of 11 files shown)


Araq:
What's up with `takeWithoutErrors`? Just reprocess the resulting `(err...)` in the `declared`/`defined` logic, that's why we did it this way.

metagn:
The `(err)` nodes are skipped while reading now (assuming they are not stacked like `(err (err ...) ...)`? if so `isError` needs to be a counter), for some reason I thought rebuilding it was the most general way of doing it.

PR #257, opened 2024-12-22
module symbols, qualified identifiers & `import as` support

### LINKED ISSUE #258: the `module.name` access syntax

metagn:
closes #258

`filenameVal` now retrieves the name of the imported module as well as the full path expression (except when the import is a string literal like `"foo/bar.nim"` but this can be done later), then a `module` symbol is defined by this name when importing it. The symbols imported from this module are added to its interface table which is attached to the module symbol's ID, then the interface table is merged into the imported symbol table for the module (hopefully this will allow `from import` to be easily implemented later). A dot expression where the LHS is a module symbol looks up the module's interface table for identifiers with the same name as the field and builds a symchoice (as in `buildSymChoiceForForeignModule`).

Since it was easy to implement in the rewrite of `filenameVal`, `import as` syntax is now supported, as well as general infix/prefix operators.


src/nimony/sem.nim
```diff
@@ -222,15 +222,28 @@ proc semInclude(c: var SemContext; it: var Item) =
...
+    c.currentScope.addOverloadable(moduleName, s)
+    var moduleDecl = createTokenBuf(2)
+    moduleDecl.addParLe(ModuleY, info)
+    moduleDecl.addParRi()
+    publish moduleSym, moduleDecl
+    var module = ImportedModule()
+    loadInterface suffix, module.iface
+    # merge module symbols into import table:
+    for strId, symIds in module.iface:
+      c.importTab.mgetOrPut(strId, @[]).add(symIds)
```
Araq, line 245: The code was carefully crafted to avoid things like these. Can it be avoided?

metagn: Maybe `importTab` can change to a mapping of identifiers to a list of modules that contain that identifier, then instead of iterating over `importTab` for the symbols directly, we iterate over the modules then over their symbols with the same name as the identifier. Would this avoid problems?

src/nimony/sem.nim
```diff
@@ -222,15 +222,28 @@ proc semInclude(c: var SemContext; it: var Item) =
...
+    c.currentScope.addOverloadable(moduleName, s)
+    var moduleDecl = createTokenBuf(2)
+    moduleDecl.addParLe(ModuleY, info)
+    moduleDecl.addParRi()
+    publish moduleSym, moduleDecl
+    var module = ImportedModule()
+    loadInterface suffix, module.iface
+    for strId, _ in module.iface:
+      # mark that this module contains the identifier `strId`:
+      c.importTab.mgetOrPut(strId, @[]).add(moduleSym)
```
Araq, line 245: I think we can make `loadInterface` do this step too and thus avoid another iteration over the iface (which could be large).

src/nimony/sem.nim
Araq: Fine, and that's how my other compiler did it.

Araq:
I think we can `semDot` on the happy path where **someObject.field** is much more common than `module.name` if we give the `ModuleY` entries a `ModuleT` dummy type. Then semDot can always inspect the type and handle the ObjectT case first.

metagn:
> I think we can keep semDot on the happy path where someObject.field is much more common than module.name if we give the ModuleY entries a ModuleT dummy type. 

Seems to not be needed since `skipModifier` ignores `NoType`. I also just made it check the item `kind`, the "module in symchoice takes over" behavior isn't necessary, and if we want it we can set the `kind` field while building the symchoice and only do this on `AllowModuleSym`.

PR #259, opened 2024-12-22
add range types to use for array index types

metagn:
Range types are added so that we can implement custom index types for arrays as well as the `array[0..2, ...]` syntax. Now, fully resolved array types always have a range type as their index type, but unresolved array types do not, as the unresolved index type can either resolve to a range type or might have to be converted to one. 

Outside of array indexes, range types themselves don't have much functionality, in sigmatch they match their base type but their base types don't match them yet. Float ranges also aren't implemented.

Range types have the representation `(rangetype Type IntLit IntLit)` where its children are the base type followed by the bounds which are always integer literals. They may need to allow unsigned integer literals in the future as well but for array indexes this isn't a concern. In the expander, array types will convert their index range type to a length integer to pass to NIFC.

Followup PRs would implement explicit/inferred index types in array constructors, test/implement array index type inference in generics (`proc foo[I, T](x: array[I, T])`), and implement indexing for custom index types.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1617,7 +1617,9 @@ proc maybeInlineMagic(c: var SemContext; res: LoadResult) =
           if n.kind == ParRi: break
           inc n
 
-proc semTypeSym(c: var SemContext; s: Sym; info: PackedLineInfo; context: TypeDeclContext) =
+proc semExprSym(c: var SemContext; it: var Item; s: Sym; start: int; flags: set[SemFlag])
+
+proc semTypeSym(c: var SemContext; s: Sym; info: PackedLineInfo; start: int; context: TypeDeclContext) =
   if s.kind in {TypeY, TypevarY}:
     let res = tryLoadSym(s.name)
     let beforeMagic = c.dest.len
@@ -1640,6 +1642,11 @@ proc semTypeSym(c: var SemContext; s: Sym; info: PackedLineInfo; context: TypeDe
         c.dest.shrink c.dest.len-1
         var t = typ.body
         semLocalTypeImpl c, t, context
+  elif context == AllowValues:
+    var dummyBuf = createTokenBuf(1)
+    dummyBuf.add dotToken(info)
+    var it = Item(n: cursorAt(dummyBuf, 0), typ: c.types.autoType)
+    semExprSym c, it, s, start, {}
   elif s.kind != NoSym:
     var orig = createTokenBuf(1)
     orig.add c.dest[c.dest.len-1]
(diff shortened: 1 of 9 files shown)


Araq:
> Need to move very carefully here, there is lots of potential for bugs.

Correct but it gets worse. I think for Nim 3 we should essentially remove `range` types and embrace `.requires` and `.ensures` annotations. I have the core of a proof engine ready for this (no more foreign dependencies).

For backwards compat we can transform `range[M..N] of int` to `(i -1 M N)`, maybe. But I don't really know.

metagn:
Mostly added them here for array indices, we could treat them the same as their base type or disallow them when they're used on their own. We could also put array indices into the array type representation but it might be harder to implement stuff like `proc foo[I, T](x: array[I, T])` otherwise.

src/nimony/sem.nim
```diff
@@ -1952,7 +2025,67 @@ proc semLocalTypeImpl(c: var SemContext; n: var Cursor; context: TypeDeclContext
     of ArrayT:
       takeToken c, n
       semLocalTypeImpl c, n, context
+      # index type, possibilities are:
```
Araq, line 2028: Use a helper proc for this `semArrayType`.

src/nimony/sem.nim
```diff
@@ -1952,7 +2025,67 @@ proc semLocalTypeImpl(c: var SemContext; n: var Cursor; context: TypeDeclContext
...
+            c.buildErr index.info, "invalid array index: " & typeToString(index)
+          else:
+            c.dest.addParLe(RangeT, info)
+            c.dest.addSubtree c.types.intType
+            c.dest.addIntLit 0, info
+            c.dest.addIntLit length - 1, info
+            c.dest.addParRi()
+      wantParRi c, n
+    of RangeT:
+      takeToken c, n
```
Araq, line 2080: Use a helper proc for this `semRangeType`.

Araq:
I wonder if things become simpler if we split up the array type into `(corearr Type INTVAL)` and `(array Type IndexType)` where the later is in the stdlib and converts IndexType to the one INTVAL array length.

metagn:
As long as they don't coexist probably, so any time after semchecking. Getting the length from the abstract array type shouldn't be too much of a problem though. Index offsets should be compiled away anyway.

```nim
type Enum = enum
  a, b, c, d
var x: array[b..c, int]
x[c]
```

would probably compile to something like `(arrat x +1)` instead of `(at x c)`, unless semchecking is too early for this (old compiler does it in codegen).

Araq:
Well they would coexist in the sense that `array` is a library implementation like `type array[Idx, T] = corearr[T, len(idx)]`. This would also do the order reversal for us.

PR #260, opened 2024-12-22
add ordinal type functionality, use for set element errors

metagn:


src/nimony/sembasics.nim
```diff
@@ -347,6 +347,199 @@ proc declToCursor*(c: var SemContext; s: Sym): LoadResult =
   else:
     result = LoadResult(status: LacksPosition)
 
+const
```
Araq, line 350: Create a new module `typeprops.nim` for these.

src/nimony/typeprops.nim
```diff
@@ -0,0 +1,205 @@
...
+    if s.status != LacksNothing:
+      result = createNaN()
+      return
+    if s.decl.symKind != TypeY:
+      result = createNaN()
+      return
+    let decl = asTypeDecl(s.decl)
+    case decl.body.typeKind
+    of EnumT:
+      # check for holes
```
Araq, line 146: Fine for now but we should just introduce `HolyEnumT` as that reflects reality better IMO.

PR #261, opened 2024-12-22
implement `from import`/`import except`, fix declared

metagn:
My best attempt at good enough looking code, open to suggestions


metagn:
Merging to avoid conflicts, worst case can be cleaned up later

PR #262, opened 2024-12-22
fix & test import/include lists

metagn:
The `import a, b, c` syntax was not implemented, although `import foo/[a, b, c]` was, both are tested


Araq:
`include` is also affected, `include a, b, c` is a thing in Nim.

metagn:
Done

PR #265, opened 2024-12-23
nimony: fixes enum fields

ringabout:
```nim
type
  Color = enum
    Red = "r1", Green = "r12", Blue = "r3"
```
Transform into `(0, "r1")` pairs.

TODO：

- [x] fixes enums with tuple fields: (int, string)
- [x] fixes `ord(enums)` and case statements with enums


metagn:
Maybe the custom string can be another child of `(efld)`, before or after the ordinal value? So we wouldn't have to check for a tuple constructor to skip to access the ordinal value every time.

PR #268, opened 2024-12-24
nimony: basic object constructors

### LINKED ISSUE #267: basic object constructors

metagn:
closes #267

Does not check if every field was assigned/insert default values etc. Assuming this is enough for now

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1299,6 +1299,7 @@ proc findObjField(t: Cursor; name: StrId; level = 0): ObjField =
     skip n # pragmas
     skip n # type
     skip n # value
+    skipParRi n
   if baseType.kind == Symbol:
     result = findObjField(objtypeImpl(baseType.symId), name, level+1)
   else:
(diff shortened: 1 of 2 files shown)


Araq:
Fine for now, but we need:

- Insert `default(T)` values for missing object fields.
- On the fly reorder so that the `field: value` pairs are reordered to match the object declaration order. 

This means it's better to iterate primarily over the object type rather than the object constructor.

With the help of some minor new syntax, `default` can be implemented entirely in system like so:

```nim
template default*(x: typedesc[bool]): bool = false
template default*(x: typedesc[char]): char = '\0'
template default*(x: typedesc[int8]): int8 = 0'i8
template default*(x: typedesc[uint8]): uint8 = 0'u8
template default*(x: typedesc[int16]): int16 = 0'i16
template default*(x: typedesc[uint16]): uint16 = 0'u16
template default*(x: typedesc[int32]): int32 = 0'i32
template default*(x: typedesc[uint32]): uint32 = 0'u32
template default*(x: typedesc[int64]): int64 = 0'i64
template default*(x: typedesc[uint64]): uint64 = 0'u64
template default*(x: typedesc[float32]): float32 = 0.0'f32
template default*(x: typedesc[float64]): float64 = 0.0'f64
template default*(x: typedesc[string]): string = ""
template default*[T: enum](x: typedesc[T]): T = T(0)

template default*[T: ptr](x: typedesc[T]): T = nil
template default*[T: ref](x: typedesc[T]): T = nil

proc default*[I, T](x: typedesc[array[I, T]]): array[I, T] {.noinit, nodestroy.} =
  var i = 0
  while i < result.len:
    result[I(i)] = default(T)
    inc i

template default*[T](x: typedesc[set[T]]): set[T] = {}

proc default*[T: object](x: typedesc[T]): T {.inline.} = T(_: default)
proc default*[T: tuple](x: typedesc[T]): T {.inline.} = (_: default)
```

So `ObjType(_: procHereNotAcall)` would lead to an auto expansion. Alternatively we can do:

```nim

proc default*[T: object](x: typedesc[T]): T {.noinit, nodestroy.} =
  for f in fields(result): result.f = default(typeof(result.f))
```

But then we need the builtin fields + typeof.

metagn:
I wanted to do `default` next, maybe we can make just the `default` overloads for `object` and `tuple` magics, which would then call invoke the overloads of `default` for the fields or use their default values if they exist?

Araq:
Sure, whatever keeps you in the flow.

PR #271, opened 2024-12-24
nimony: implement dot/subscript setters

### LINKED ISSUE #270: dot/subscript setter overloads
```nim
proc `[]=`(a, b, c) = ...
a[b] = c # becomes `[]=`(a, b, c)
proc `foo=`(a, b) = ...
a.foo = b # becomes `foo=`(a, b)
```

Likely the subscript version has more priority but they should have similar implementations, calling `tryBuiltinDot`/`tryBuiltinSubscript` (doesn't exist yet, has to be split out of `semBuiltinSubscript`) first before looking up the overloads.

Note: this means any subscript magic overloads added in #269 would need `[]=` variants. Unless we add some behavior that falls back to `[]` overloads if `[]=` isn't found, but this adds complexity and isn't in original Nim.

```nim
type Foo = object
  x: int

proc `[]`(foo: var Foo, i: int): var int = foo.x

var foo = Foo(x: 123)
foo[5] = 456 # Error: type mismatch
```

Edit: Weirdly enough field assignments do fall back to dotcalls for the LHS if a `foo=` overload isn't found. Maybe we can do both by checking for error tokens after `semCall` using `TransformedCallSource`?

metagn:
closes #270

Field setters falling back to dotcalls is left unimplemented, if it's needed then we can disable field setters until it's implemented. Curly subscript assignments are also unimplemented but are much simpler since they always build `{}=`.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1061,6 +1061,55 @@ proc semCast(c: var SemContext; it: var Item) =
     c.dest.shrink beforeExpr
     c.buildErr info, "cannot `cast` between types " & typeToString(srcType) & " and " & typeToString(destType)
 
+proc buildCallSource(buf: var TokenBuf; cs: CallState) =
+  case cs.source
+  of RegularCall:
+    buf.add cs.callNode
+    buf.addSubtree cs.fn.n
+    for a in cs.args:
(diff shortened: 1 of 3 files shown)


Araq:
As for the other PR, please use helper procs.

PR #274, opened 2024-12-24
fix issues with generic proc matching

metagn:
1. Instantiated procs were only mapped to the types of their parameters, not which routine symbols they were instantiated from
2. Magic type invocations (i.e. `array[I, T]`) were left as `(at (array) I T)` in generics
3. `inc inGeneric` leaked in `semGenericParams` for type sections since they did not push a new routine
4. `linearMatch` matching a generic parameter only skipped a single token after doing so, i.e. `(array T I)` matching `(array (i -1) ...)` would match `T` to `(i -1)` then move to the next token after `(i`, matching `I` to `-1`

`inc inGeneric` leaking means procs were not instantiated in some of the tests, the `HSlice[T, U]` return type of `..` remains as `InvokeT` after instantiation rather becoming a symbol which causes issues, probably `semLocalType` isn't called on the return type during instantiation or whatever has to be done

### DIFF
--- a/src/nimony/semdata.nim
+++ b/src/nimony/semdata.nim
@@ -79,7 +79,7 @@ type
     types*: BuiltinTypes
     typeMem*: Table[string, TokenBuf]
     instantiatedTypes*: OrderedTable[string, SymId]
-    instantiatedProcs*: OrderedTable[string, SymId]
+    instantiatedProcs*: OrderedTable[SymId, OrderedTable[string, SymId]]
     thisModuleSuffix*: string
     moduleFlags*: set[ModuleFlag]
     processedModules*: HashSet[string]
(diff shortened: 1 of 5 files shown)


src/nimony/semdata.nim
```diff
@@ -79,7 +79,7 @@ type
     types*: BuiltinTypes
     typeMem*: Table[string, TokenBuf]
     instantiatedTypes*: OrderedTable[string, SymId]
-    instantiatedProcs*: OrderedTable[string, SymId]
+    instantiatedProcs*: OrderedTable[SymId, OrderedTable[string, SymId]]
```
Araq, line 82: Shouldn't instead the "comes from" symbol simply be part of the lookup key?

metagn: As a tuple or in the string?

PR #279, opened 2024-12-26
implement type kind typeclass, nil & enum conversions

metagn:
Instead of leaving `(object)` `(enum)` etc as typeclasses, they instead become `(typekind (object))`, and this is also done for other type kinds. The `elif f.kind == ParLe and a.kind == ParLe` branch in sigmatch used to handle the old representation and is removed.

### DIFF
--- a/src/gear3/expander.nim
+++ b/src/gear3/expander.nim
@@ -217,7 +217,7 @@ proc traverseType(e: var EContext; c: var Cursor; flags: set[TypeFlag] = {}) =
     inc c
   of ParLe:
     case c.typeKind
-    of NoType, OrT, AndT, NotT, TypedescT, UntypedT, TypedT:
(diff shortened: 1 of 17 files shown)


metagn:
Didn't catch that `EnumType(0)` and `PtrType(nil)` weren't implemented, so had to do those to enable the tests as well.

Araq:
But the lack of enum was a feature as `MyEnum(x)` can violate safety.

metagn:
Switched to cast, it was broken too because `skipDistinct` skips to the type body and so `isEnumType` didn't work

Araq:
On the other hand, we can also prove it at compile-time...

metagn:
Ok, I'll split the PR for any further changes

Araq:
Dude, don't prove it here in sem, there will be a phase 3 responsible for it.

PR #283, opened 2024-12-26
inferle engine for proving things

Araq:


Araq:
Plan: Not only variables but also important expressions like `x.len` are mapped to unique `VarIds`.

PR #285, opened 2024-12-26
implement default, object & tuple as magics, full object constructors

### LINKED ISSUE #276: the `default` builtin
Implementation is https://github.com/nim-lang/nimony/pull/268#issuecomment-2560641327 and the comment below it but overloading on `array[I, T]` is blocked by #274, `object`/`tuple` constraints aren't implemented and typedesc overloading isn't tested (#279)

metagn:
closes #276

Object constructors and the default object magic both build an object constructor containing every field in the object constructor: first starting with the fields of the most shallow base type of the object type (if it exists), then of its base type and so on, and finally adding the fields from the original object type. This order is the same one used in the VM in the old compiler, could be changed.

Object constructors build a table of the given field assignments first, then when the output object constructor is being built, the table is looked up for each field to see if it has been set and the assignment is used if it has. I'm not sure what the optimal way to do this is. Currently it looks up the object fields when initially going through the object constructor and maps the assignment to the field symbol ID. We could also map to the field name strings as well which would save iterations over the fields, this would require that we get the basename from each field and look that up when building the output object constructor (and don't bother when the field table is empty). But maybe it has pitfalls, it's not 1-to-1 with the field access syntax so I didn't start with it. I'm guessing we should optimize for most of the object fields being set in the constructor since this is going to be the case for typed AST and is reasonable enough for user code as well. ~~I would guess the second option is better for this but~~ maybe there's some method I'm not seeing.

Edit: Mapping to the symbol IDs is probably better for typed AST if we add the symbol IDs directly for symbol tokens. Now, it checks if the symbol token is a field symbol, and uses it directly for the mapping if it is (otherwise converts it to an identifier and looks it up). If we want to protect against field symbols belonging to other objects (after macro manipulation for example) then maybe we can build the set of field symbol IDs by iterating over the object type first.

Typevar symbols are also declared globally now for generic procs in system to work, it could be restricted to typevars in exported routines/types but maybe this would be too complex or incorrect.

### DIFF
--- a/src/nifler/bridge.nim
+++ b/src/nifler/bridge.nim
@@ -525,7 +525,8 @@ proc toNif*(n, parent: PNode; c: var TranslationContext) =
       toNif(n[i], n, c)
     # n.len-3: pragmas: must be empty (it is deprecated anyway)
     if n.len == 0:
-      c.b.addEmpty 2 # pragmas, body
+      # object typeclass, has no children
+      discard
     else:
       if n[n.len-3].kind != nkEmpty:
         c.b.addTree "err"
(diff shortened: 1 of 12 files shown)


Araq:
Please rebase.

PR #295, opened 2024-12-27
implement low/high as magic

metagn:
Currently entirely a magic unlike `default` for a couple reasons:

* The overload for either `int` and `uint` would be hard to represent at this point
* Enums and range types require the use of a magic
* The original overloads also use a magic

### DIFF
--- a/src/gear3/expander.nim
+++ b/src/gear3/expander.nim
@@ -217,7 +217,7 @@ proc traverseType(e: var EContext; c: var Cursor; flags: set[TypeFlag] = {}) =
     inc c
   of ParLe:
     case c.typeKind
-    of NoType, OrT, AndT, NotT, TypedescT, UntypedT, TypedT, TypeKindT:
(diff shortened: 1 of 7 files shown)


Araq:
Nah, use the signatures from Nim v2's system.nim, they shouldn't be too hard to implement and have proven to be decently overloadable.

metagn:
Needed some patches for the original signatures:

* Added `Ordinal[T]` with no functionality other than matching all ordinal types (generic argument is ignored)
* The special case for `|` in type contexts is implemented (which exists in original nim, see [this dummy declaration](https://github.com/nim-lang/Nim/blob/86d6f71f5afd03ce372ee89b348797f43899579a/lib/system.nim#L1432)), maybe we could come up with a proper overload later
* `array`/`range`/`set` etc are regular symbols so `tryTypeClass` didn't work for them, now they are checked in `semTypeSym` to see if they are a typeclass

PR #301, opened 2024-12-29
add back base types to enums

metagn:
refs #162 and #284

The issue before was that nifler didn't push an empty token for them

Guessing sem needs to set the base type, could do this now or in a followup PR


Araq:
> Guessing sem needs to set the base type, could do this now or in a followup PR

Yes, in a followup PR.

PR #305, opened 2024-12-29
gear3: consider mangled syms with different owners

metagn:
i.e.

```nim
proc foo() {.importc: "foo".}
proc bar() =
  foo()
```

Previously `foo` wasn't mangled.

This is probably a slow way of doing this since it's on every symbol use, maybe each owner can have its own mangle table instead that copies every mangled symbol from the previous owner. Or maybe the key can be changed or there is some other different solution, in which case this PR can be closed.

### DIFF
--- a/src/gear3/expander.nim
+++ b/src/gear3/expander.nim
@@ -1108,9 +1108,14 @@ proc writeOutput(e: var EContext) =
     of Ident:
       b.addIdent(pool.strings[c.litId])
     of Symbol:
-      let owner = ownerStack[^1][0]
-      let key = (c.symId, owner)
-      let val = e.toMangle.getOrDefault(key)
+      var val = ""
+      var ownerPos = ownerStack.len - 1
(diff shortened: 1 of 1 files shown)


Araq:
This owner stack is a hack, we're better off modelling the scopes properly. A bit annoying so we might also always generate the "full" names `name.N.mod` in sem.nim instead.

SirOlaf:
Maybe I'm misunderstanding the algorithm but maybe this should adjust `nextIsOwner` https://github.com/nim-lang/nimony/blob/8185c52d763ea64e8997b7f4e976b87892ccaeef/src/gear3/expander.nim#L1148

metagn:
`nextIsOwner` seems to be for immediately after `ParLe`, to mark that the next token is a symbol token that should be added as the new owner. It wouldn't be used in the closing paren token.

Closing so it doesn't stall, can reopen if deemed good enough for now

PR #308, opened 2024-12-30
gear3: transform right parts of gvar into asgn if not literals

ringabout:
Transforms `gvar` with expressions into a `(gvar; asgn)` pair. Literals are allowed to be assigned in the global sphere.

### DIFF
--- a/src/gear3/expander.nim
+++ b/src/gear3/expander.nim
@@ -896,7 +896,20 @@ proc traverseLocal(e: var EContext; c: var Cursor; tag: string; mode: TraverseMo
     traverseType e, c
 
   if mode != TraverseSig:
-    traverseExpr e, c
+    if tag == "gvar":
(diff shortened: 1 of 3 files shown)


Araq:
NIFC already has some logic for this, iirc. Not sure if that's NIFC's job though.

ringabout:
Yeah, this PR relies on the logic in nifc to move `asgn` to InitSection. Though nifc doesn't have logic to transform `(gvar x expr)` into `(gvar x); (asgn x expr)`

ringabout:
I think `gear3` needs to generate a `main` function and collects top level statements (except decls) into a main function

Araq:
Ok.

Araq:
Now obsolete, I think.

ringabout:
I will fix it in the nifc parts

PR #309, opened 2024-12-30
nifbuilder: fixes LineInfo for filenames

ringabout:
According to the `nif` spec, there are 3 forms for line infos:
```
1. `<column-diff>`
2. `<column-diff, line-diff>`
3. `<column, line, filename>`
```

If the `line-diff` is 0 but `filename` is different, It should generate `1;;filename` or `;;filename` instead of `1;filename` which causes problems for parsing

### DIFF
--- a/src/lib/nifbuilder.nim
+++ b/src/lib/nifbuilder.nim
@@ -19,6 +19,9 @@ type
     f: File
     nesting: int
 
+  LineInfoFormat = enum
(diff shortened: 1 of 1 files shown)


Araq:
Sure but your patch doesn't do that, does it?

Araq:
Make this so clear it's obviously true, instead of counting separators use an enum about which format of the line information is emitted.

ringabout:
Sure thing

Araq:
Thanks, now it's "obviously correct". :-)

PR #311, opened 2024-12-31
try open calls to `[]` and `[]=` as builtin subscripts first

metagn:
The idea is that untyped templates and generics produce calls to `[]` for all subscript expressions to capture the overloads of `[]` in scope at the time. To handle the cases where these subscript expressions are meant as builtin subscripts (e.g. generic instantiations) and not as behavior that any overload of `[]` would provide, we check for these types of call nodes before triggering overload resolution on the call and try them as builtins.

For clarity and maybe performance, we could give calls to `[]` and `[]=` produced from untyped subscript expressions different representation, e.g. `(atcall (ochoice `[]` ...) a b c ...)` or `(call (atchoice `[]` ...) ...)`. This would stop things like `` `[]`(seq, int) `` from working.


Araq:
Why is any of this important when we cannot yet write a binary tree or any program of mild complexity...

Issue #269, opened 2024-12-24
builtin subscripts & required magics for indexing


metagn:
From the ideas in #194

The subscript expression has the following builtin behaviors before delegating to overloads of `[]`:

- [x] generic type invocation
- [ ] generic routine invocation
  - [ ] including symchoices, probably should be left as `(at (choice ...) ...)`
- [x] array indexing (in overload)
- [ ] openarray/uncheckedarray/varargs indexing
- [x] cstring indexing (in overload)
- [x] string/seq indexing - done for seq/string implementations
- [x] tuple indexing

We could maybe move the indexing operations to overloads rather than handling them in the compiler, but this has some incompatibilities as outlined [here](https://github.com/nim-lang/Nim/issues/22970#issuecomment-2452654268), namely:

* they accept arbitrary pointer dereferences,
* any integer type is allowed for the index.

Tuple indexing may need a wacky signature like:

```nim
proc `[]`[T: tuple](tup: T; index: static int): typed {.magic: TupleIndex.}
```

which might not work as expected any time soon, so it might be better to keep it builtin.

In any case, it would be nice to not have the super-overload of `ArrGet` /`ArrPut` in the old Nim that needs constant special casing. Hopefully we won't need it.

We also might need different tags (magics) than `(at)` for different behaviors of the operator, especially given `(at)` is the tag for parsed, unsemchecked subscripts. Probably enough to separate indexing between array-likes and seq-likes but idk.

PR #313, opened 2024-12-31
implement array/cstring indexing

metagn:
refs #269

### DIFF
--- a/lib/std/system.nim
+++ b/lib/std/system.nim
@@ -76,9 +76,11 @@ proc high*[T: Ordinal|enum|range](x: typedesc[T]): T {.magic: "High", noSideEffe
 proc high*[I, T](x: typedesc[array[I, T]]): I {.magic: "High", noSideEffect.}
 
 proc `[]`*[I, T](x: array[I, T], i: I): T {.magic: "ArrAt".}
-proc `[]`*(x: string, i: int): char {.magic: "StrAt".}
-proc `[]`*(x: cstring, i: int): char {.magic: "CstrAt".}
-proc `[]=`*() {.magic: "ArrPut", varargs.}
+proc `[]`*(x: cstring, i: int): char {.magic: "Pat".}
+template `[]=`*[I, T](x: array[I, T], i: I; elem: T) =
+  (x[i]) = elem
+template `[]=`*(x: cstring, i: int; elem: char) =
+  (x[i]) = elem
 
 # integer calculations:
 proc `+`*(x: int8): int8 {.magic: "UnaryPlusI", noSideEffect.}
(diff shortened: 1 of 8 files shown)


lib/std/system.nim
```diff
@@ -75,6 +75,11 @@ proc low*[I, T](x: typedesc[array[I, T]]): I {.magic: "Low", noSideEffect.}
 proc high*[T: Ordinal|enum|range](x: typedesc[T]): T {.magic: "High", noSideEffect.}
 proc high*[I, T](x: typedesc[array[I, T]]): I {.magic: "High", noSideEffect.}
 
+proc `[]`*[I, T](x: array[I, T], i: I): T {.magic: "ArrAt".}
+proc `[]`*(x: string, i: int): char {.magic: "StrAt".}
+proc `[]`*(x: cstring, i: int): char {.magic: "CstrAt".}
```
Araq, line 80: That would just be `PatX`.

lib/std/system.nim
```diff
@@ -75,6 +75,11 @@ proc low*[I, T](x: typedesc[array[I, T]]): I {.magic: "Low", noSideEffect.}
 proc high*[T: Ordinal|enum|range](x: typedesc[T]): T {.magic: "High", noSideEffect.}
 proc high*[I, T](x: typedesc[array[I, T]]): I {.magic: "High", noSideEffect.}
 
+proc `[]`*[I, T](x: array[I, T], i: I): T {.magic: "ArrAt".}
+proc `[]`*(x: string, i: int): char {.magic: "StrAt".}
```
Araq, line 79: That is just some `.inline` proc, we need to be able to acces the `string` as its `(ptr, len)` pair though inside the proc body.

lib/std/system.nim
```diff
@@ -75,6 +75,11 @@ proc low*[I, T](x: typedesc[array[I, T]]): I {.magic: "Low", noSideEffect.}
...
+proc `[]`*(x: cstring, i: int): char {.magic: "CstrAt".}
+proc `[]=`*() {.magic: "ArrPut", varargs.}
```
Araq, line 81: Why is this varargs?

lib/std/system.nim
```diff
@@ -75,6 +75,11 @@ proc low*[I, T](x: typedesc[array[I, T]]): I {.magic: "Low", noSideEffect.}
 proc high*[T: Ordinal|enum|range](x: typedesc[T]): T {.magic: "High", noSideEffect.}
 proc high*[I, T](x: typedesc[array[I, T]]): I {.magic: "High", noSideEffect.}
 
+proc `[]`*[I, T](x: array[I, T], i: I): T {.magic: "ArrAt".}
```
Araq, line 78: That returns a `var T` btw.

lib/std/system.nim
Araq: Also this one is just:

```nim

proc `[]`*[I, T](x: array[I, T], i: I; elem: sink T) {.inline.} =
  (a[i]) = elem

```

Araq: Or the same as a `template`, then the `sink` can be omitted too.

metagn: Made into template for now, I thought maybe we would need a generic version of this (hence varargs for more than 1 subscript argument) but it wouldn't change much from existing Nim which only has this behavior for the builtins anyway

lib/std/system.nim
metagn: Excluded string for now

PR #317, opened 2025-01-02
plugin support: first implementation

Araq:


Araq:
Merging it now already because the refactorings must cause conflicts otherwise.

PR #322, opened 2025-01-02
nifc: fixes top level var decls with non-literals

ringabout:
fixes top level var decls

### DIFF
--- a/src/nifc/codegen.nim
+++ b/src/nifc/codegen.nim
@@ -351,7 +351,7 @@ proc genVarDecl(c: var GeneratedCode; t: Tree; n: NodePos; vk: VarKind; toExtern
       c.code.insert(Token(StaticKeyword), beforeDecl)
     if t[d.value].kind != Empty:
       if vk == IsGlobal and
-        t[d.value].kind notin {IntLit, UIntLit, FloatLit, CharLit, StrLit, FalseC, TrueC}:
+        t[d.value].kind notin {IntLit, UIntLit, FloatLit, CharLit, StrLit, FalseC, TrueC, AconstrC, OconstrC}:
         c.add Semicolon
         moveToInitSection:
           c.add name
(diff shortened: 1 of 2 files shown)


src/nifc/codegen.nim
```diff
@@ -350,11 +350,24 @@ proc genVarDecl(c: var GeneratedCode; t: Tree; n: NodePos; vk: VarKind; toExtern
     if vis == StaticC:
       c.code.insert(Token(StaticKeyword), beforeDecl)
     if t[d.value].kind != Empty:
-      c.add AsgnOpr
-      if vk != IsLocal: inc c.inSimpleInit
-      genx c, t, d.value
-      if vk != IsLocal: dec c.inSimpleInit
-    c.add Semicolon
+      if vk == IsGlobal and
+        t[d.value].kind notin {IntLit, UIntLit, FloatLit, CharLit, StrLit, FalseC, TrueC}:
```
Araq, line 354: But "bracket expressions" for arrays and objects are fine too and it would be wrong to move them!

src/nifc/codegen.nim
```diff
@@ -350,11 +350,26 @@ proc genVarDecl(c: var GeneratedCode; t: Tree; n: NodePos; vk: VarKind; toExtern
...
     if t[d.value].kind != Empty:
-      c.add AsgnOpr
-      if vk != IsLocal: inc c.inSimpleInit
-      genx c, t, d.value
-      if vk != IsLocal: dec c.inSimpleInit
-    c.add Semicolon
+      if vk == IsGlobal and
+        t[d.value].kind notin {IntLit, UIntLit, FloatLit,
+          CharLit, StrLit, FalseC, TrueC, AconstrC,
+          OconstrC, InfC, NegInfC, NanC, SufC}:
```
Araq, line 356: You need some `isLiteral` deep check for OconstrC and AconstrC as `[1, 2]` is very different from `[myVar, 3]`.

ringabout: I have no clues how to deal with `Tree` iteratively. Do I need to use `hasNext` to query the end of the tree or something?

Araq: You use the simple `while n.kind != ParRi:...` loop but your way is fine too. Also NIFC doesn't yet use the newer APIs so that's probably why you were confused.

PR #335, opened 2025-01-06
inline structural/alias invocations even with generic arguments, broken

metagn:
In:

```nim
type
  NodeObj[T] = object
    x: Node[T]
  Node[T] = ref NodeObj[T]
```

first `SemcheckTopLevelSyms` will save the untyped AST of the types to their symbols, however they will not define typevars or look them up in the body. This means when `NodeObj` is being semchecked, attempting to inline `Node[T]` by instantiating it will not work because there will be no typevars to replace.

This is a general problem for forwarded generic types:

```nim
type
  Foo[T] = object
    x: Bar[int]
  Bar[T] = object
```

A solution might be to define the typevars in the `SemcheckTopLevelSyms` pass and use an untyped prepass on the body so that they can be instantiated. But this is not done yet as an untyped prepass is not implemented.


metagn:
Can reopen when ready

PR #336, opened 2025-01-06
recursively skip invocations for generic ref object types

metagn:
The alternative to #335 for just object primitives, but this is a more deep rooted problem for `ref` etc in general, for example `nil` does not match `Node[T]`. Unless it's a priority, we can wait until #335 can work.


metagn:
Can reopen when needed

PR #337, opened 2025-01-06
gear3: adds a type for arrayConstr

ringabout:
`(aconstr Type Expr*)` requires a type

### DIFF
--- a/src/gear3/expander.nim
+++ b/src/gear3/expander.nim
@@ -873,8 +873,8 @@ proc traverseExpr(e: var EContext; c: var Cursor) =
         skipParRi(e, c)
       of AconstrX:
         e.dest.add tagToken("aconstr", c.info)
-        inc c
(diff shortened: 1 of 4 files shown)


Araq:
Yes but that is NIFC, not Nimony's model. Gear3 get use the typenavigator to add a type to the construct.

src/gear3/expander.nim
```diff
@@ -871,6 +871,12 @@ proc traverseExpr(e: var EContext; c: var Cursor) =
         e.dest.shrink beforeType
         traverseExpr(e, c)
         skipParRi(e, c)
+      of AconstrX:
+        e.dest.add tagToken("aconstr", c.info)
+        inc c
+        var arrayType = e.typeCache.getType(c)
```
Araq, line 877:
```suggestion
        var arrayType = e.typeCache.getType(c)
        inc c
```

As you want the type of the array construction, not of a single element.

PR #338, opened 2025-01-06
nimony: implements hook index

ringabout:
- [x] createIndex
- [x] readIndex

image

### DIFF
--- a/src/lib/nifindexes.nim
+++ b/src/lib/nifindexes.nim
@@ -167,10 +167,10 @@ proc createIndex*(infile: string; buildChecksum: bool;
 
     for value in values:
       let (obj, sym) = value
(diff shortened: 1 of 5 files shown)


src/lib/nifindexes.nim
```diff
@@ -155,17 +158,39 @@ proc createIndex*(infile: string; buildChecksum: bool) =
...
+  content.add "\n"
+
+  for (key, values) in hookIndexMap.pairs:
+    var hookSectionBuf = default(TokenBuf)
+    let tag = registerTag(key)
+    hookSectionBuf.addParLe tag
+
+    for value in values:
+      let (obj, sym) = value
+      let diff = symToOffsetMap[sym] - symToOffsetMap[obj]
```
Araq, line 170: That's not how these diffs work. But you can just store this offset as it is without any diffing this time.

PR #342, opened 2025-01-08
remove unused cyclic import of sigmatch in semdata

metagn:
Sigmatch imports semdata and semdata imports sigmatch but semdata doesn't use sigmatch, causes problems when compiling semdata or importing it by itself without sigmatch but not when compiling sigmatch or importing it by itself


metagn:
Should be harmless, merging

PR #343, opened 2025-01-08
fixes some issues

ringabout:
- `inSimpleInit` means a global var initialization. Since the value parts are moved to `InitSection`, `inSimpleInit` should zero.

- `low(int64)` cannot be negated

### DIFF
--- a/src/nimony/xints.nim
+++ b/src/nimony/xints.nim
@@ -116,7 +116,9 @@ proc `$`*(a: xint): string =
   else: $a.val
 
 proc createXint*(x: int64): xint =
-  if x < 0:
+  if x == low(int64):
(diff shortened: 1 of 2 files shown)


src/nimony/xints.nim
```diff
@@ -116,7 +116,9 @@ proc `$`*(a: xint): string =
   else: $a.val
 
 proc createXint*(x: int64): xint =
-  if x < 0:
+  if x == low(int64):
+    xint(neg: true, val: cast[uint64](x))
```
Araq, line 120: Obscure bit logic here, better use:

```suggestion
    xint(neg: true, val: uint64(high(int64)) + 1'u64)
```

PR #345, opened 2025-01-09
implement field visibility

metagn:
Code is a bit ugly for now

Currently checks for the line info of the field access to compare modules. An alternative might be to track the current instantiated routine symbol in a stack which would include templates

Tests need some workarounds for/tweaks in hastur, otherwise they work for the most part. Imported modules don't seem to compile to C yet so will try to see how to test just sem compilation of `tfieldvis`. `tfieldviserr` also has mismatching output with the `make` call

The type conversion in:

```nim
template getPrivateTempl*[T](x: Generic[T]): T =
  T(x.private)
```

is a workaround, the type of `x.private` is the `T` in the original definition of `Generic`. This only errors for templates because other routines use `commonType` since #320, which matches anything to typevar expected types. Maybe `x.private` can substitute the type of `private` according to the `InvokeT` params of the type of `x`, this would have to be done for inherited types too.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1349,7 +1349,15 @@ proc semCall(c: var SemContext; it: var Item; source: TransformedCallSource = Re
   else:
     resolveOverloads c, it, cs
 
-proc findObjField(t: Cursor; name: StrId; level = 0): ObjField =
+proc genericRootSym(td: TypeDecl): SymId =
+  result = td.name.symId
+  if td.typevars.typeKind == InvokeT:
+    var root = td.typevars
+    inc root
+    assert root.kind == Symbol
+    result = root.symId
+
+proc findObjFieldAux(t: Cursor; name: StrId; level = 0): ObjField =
   assert t == "object"
   var n = t
   inc n # skip `(object` token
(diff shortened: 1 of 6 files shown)


src/nimony/sem.nim
```diff
@@ -1767,7 +1813,10 @@ proc semObjectType(c: var SemContext; n: var Cursor) =
     takeToken c, n
   else:
     # object fields:
+    let oldScopeKind = c.currentScope.kind
     withNewScope c:
+      # copy toplevel scope status for exported fields
```
Araq, line 1818: Wait ... what? Why?

src/nimony/sem.nim
```diff
@@ -3273,8 +3322,11 @@ proc semTypeSection(c: var SemContext; n: var Cursor) =
       takeToken c, n
       isGeneric = false
     else:
+      let oldScopeKind = c.currentScope.kind
```
Araq, line 3325: Why?

Araq:
> Code is a bit ugly, can be cleaned up if necessary

We rarely if ever find the time to clean up things so it's better to do it right now. That said, I don't find it particularly ugly.

src/nimony/sem.nim
metagn: Exporting a symbol is only allowed if the current scope has kind `ToplevelScope`, so this is propagated from object declarations in toplevel scopes to the scope of their fields so that the fields can be exported. The only other use of scope kinds seems to be for whether or not to make global or local syms in `identToSym` which already unconditionally makes global syms for field symbols, so I thought it was sufficient to change it.

Generics in general also open a new scope though (the other place where this is done) which is a bit more murky, this is supposed to work:

```nim
type Foo[T] = object
  field*: T
```

But this would compile even though it doesn't add the symbols to the top level scope:

```nim
type Foo[T] = (; const field* = 123; T)
```

Maybe use a `ToplevelFieldScope` kind instead?

Araq: Ok, but can be done later.

PR #349, opened 2025-01-10
gear3: support basic for loops inlining

ringabout:


Araq:
Move the implementation to its own module please. If that requires more refactorings you can do that as a follow-up PR.

blackmius:
add test for combined inline iterators?

```
iterator countup(n: int): int =
  var i = 0
  while i <= n:
    yield i
    inc i

iterator powers(n: int): int =
  for i in countup(n):
    yield i
    yield i*i
    yield i*i*i
```

blackmius:
how break and continue behave with outer block tags?

```
block OUTER:
  for i in countup(10):
    if i % 2 == 0:
      continue
    if i == 6:
      break OUTER
```

PR #359, opened 2025-01-15
disable expected NIF output of tdefault

metagn:
Keeps needing to be updated whenever something is added to system due to line info, already tests that it compiles anyway, the NIF output can be manually tested when necessary


Araq:
We could simply remove the "expected NIF outcome" instead.

metagn:
The `.nif` file is the one I disabled, sorry for ambiguous wording

Araq:
Yeah, just remove it.

Issue #304, opened 2024-12-29
Nimony:  ref T and ptr T support plus nil


Araq:
- [x] needs a .compilerproc mechanism and allocator support
- [x] support `nil` for these - tested by #319 at least
- [x] support `(ref T)(fields: values...)` construction - accepted in frontend
- [x] lower ref object construction (`NewOconstrX`)
- [x] derefs for ref object dot fields
- [x] make deref operator `[]` part of system.nim: a `.magic` for pointers/refs
- [x] nominal ref/ptr objects - possibly as own type kinds
- [x] Atomic RC ops for `ref`

Araq:
@metagn Work on the frontend aspects of this please.

PR #360, opened 2025-01-15
implement ref objects by generating inner object decl

metagn:
refs #304

This is the simplest implementation for the time being, another implementation would take longer so I did this first.

When a `ref object` type is declared, a new symbol is created for the inner object type and a type declaration is built for it, copying the typevars, pragmas and object AST from the `ref object` type. Then the body of the `ref object` type declaration is replaced with either `ref ObjSym` if the type is not generic and `ref ObjSym[T1, T2, ...]` if it is. Finally the type declaration built for the inner object is processed (a forward type pass is also run on it before finishing the `ref object` type).

This works with the existing code for `ref ObjType` skipping and gives a proper type to expressions like `foo[]` where `foo` has `ref object` type.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -3347,6 +3347,7 @@ proc semTypePragmas(c: var SemContext; n: var Cursor; beforeExportMarker: int) =
     exportMarkerBecomesNifTag c, beforeExportMarker, crucial
 
 proc semTypeSection(c: var SemContext; n: var Cursor) =
+  let startCursor = n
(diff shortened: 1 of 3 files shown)


Araq:
I still like `RefObjectT` better than these anon symbols.

metagn:
Then we might still need this in the compatibility switch for stuff like dereferences of foo objects

Araq:
You mean for `typeof(myref[])`?

metagn:
I've seen direct uses of `myref[]` as well

PR #365, opened 2025-01-15
Fix compile errors from nifc when asm enabled

demotomohiro:
I tried to compile nifc with `nim c -d:enableAsm src/nifc/nifc.nim` to try nifc's native target. But I got compile errors.

In `proc matchAny(c: var Context): bool` in `src/nifc/amd64/emitter.nim`, it seems `result` should be set in `of ParLe:` branch.
But I don't know how `result` should be set, so I just initialized `result` with `false`.

In `proc allocParamWin64*(a: var RegAllocator; param: AsmSlot): Location =` proc in `src/nifc/amd64/machine.nim`,
it seems `result` is assigned in all branches and always initialized but compiler says "Error: Cannot prove that 'result' is initialized.".

### DIFF
--- a/src/nifc/amd64/emitter.nim
+++ b/src/nifc/amd64/emitter.nim
@@ -120,6 +120,8 @@ proc matchAndEmitTag(c: var Context; tag: TagId; asStr: string): bool =
     result = false
 
 proc matchAny(c: var Context): bool =
+  result = false
+
   while true:
     case c.current.kind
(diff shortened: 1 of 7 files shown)


Araq:
> to try nifc's native target.

Don't bother, it's not ready at all. You can develop it though, Port https://github.com/rui314/chibicc/blob/main/codegen.c from C to Nim and then patch it to use my register allocation strategy.

demotomohiro:
Thank you!
I will read that C file and try to develop it.

Issue #357, opened 2025-01-14
Implement .compile pragma


Araq:
- [ ] Make hastur download mimalloc as part of the build process. Maybe depend on it via a git submodule.
- [ ] Add `.compile: "foo.c"` to Nimony so that we can also depend on C code like mimalloc.
- [ ] Make use of `.compile` in system.nim so we can use a decent memory allocator which offers `allocatedSize()` so that our new super efficient strings and seq can use it.

PR #367, opened 2025-01-15
implements `{.compile.}`

ringabout:
fixes https://github.com/nim-lang/nimony/issues/357

ref https://github.com/microsoft/mimalloc

- [x] Make hastur download mimalloc as part of the build process. Maybe depend on it via a git submodule.
- [x] Add `.compile: "foo.c"` to Nimony so that we can also depend on C code like mimalloc.
- [x] Make use of `.compile` in system.nim so we can use a decent memory allocator which offers `allocatedSize()` so that our new super efficient strings and seq can use it.

### DIFF
--- a/tests/nimony/sysbasics/tcompile.nim
+++ b/tests/nimony/sysbasics/tcompile.nim
@@ -1,5 +1,4 @@
-{.compile: "tests/nimony/sysbasics/foo.c".}
-# TODO: stores filename in the deps so that relative paths work
+{.compile: "foo.c".}
 
 proc myFunc(): int {.importc: "myFunc".}
 
--- a/tests/nimony/sysbasics/tcompile.nim
+++ b/tests/nimony/sysbasics/tcompile.nim
@@ -1,5 +1,8 @@
 {.compile: "foo.c".}
+{.compile("foo.c", "-fno-strict-aliasing")
 
 proc myFunc(): int {.importc: "myFunc".}
+proc myFunc2(): int {.importc: "myFunc2".}
 
 let s = myFunc()
+let s2 = myFunc2()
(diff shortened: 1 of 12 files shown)


tests/nimony/sysbasics/tcompile.nim
```diff
@@ -0,0 +1,6 @@
+{.compile: "tests/nimony/sysbasics/foo.c".}
```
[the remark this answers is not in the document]
Araq: Once that feature is complete, yes as we are producing a makefile for every full build already.

ringabout:
`{.compile.}` is almost finished, but it requires
- [ ] [`importc` mangling support](https://github.com/nim-lang/nimony/pull/305) 
- [ ] `currentSourcePath`: so `{.compile("$path", "-I$includedPath").}` can get the source path for parameters at the compile time

Araq:
Have you considered to make Nimony emit `(compile "foobar.c")` information to the generated index file. Then it would be precise for `when x: {.compile ...}` and it would not require this ugly Nifler addition.

tests/nimony/sysbasics/tcompile.nim
```diff
@@ -0,0 +1,8 @@
+{.compile: "foo.c".}
+{.compile("foo.c", "-fno-strict-aliasing")
```
SirOlaf, line 2:
```suggestion
{.compile("foo1.c", "-fno-strict-aliasing").}
```

ringabout:
Succeeded by https://github.com/nim-lang/nimony/pull/383

tests/nimony/sysbasics/tcompile.nim
ringabout: done

PR #369, opened 2025-01-16
fixes for symbol defs in generics and templates

metagn:
Probably no high priority bugs here, just wanted to lay this out before I forgot about it. If the fixes aren't good enough then this can be closed for now.

Type definitions only semcheck the type bodies if the symbol is freshly declared (an identifier). However in generics and templates, definitions define symbols even if their bodies are not fully checked yet. To deal with this, every symboldef introduced by a template/generic instantiation is marked as "fresh" in the sem context so that type definitions can typecheck their bodies. The way this is done I'm not sure about, it's a set of sym IDs that removes a sym when it's been processed via `handleSymDef`, I don't know if this would leave any symbols remaining. It's not scope sensitive either.

The test does not pass hexer (which is not tested) due to the `tryLoadSym` call in `traverseType` breaking for local type symbols, and doesn't pass nifc due to types not being lifted yet, so for now it's a semcheck test.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -379,8 +379,8 @@ proc declareResult(c: var SemContext; info: PackedLineInfo): SymId =
 
 # -------------------- generics ---------------------------------
 
-proc newSymId(c: var SemContext; s: SymId; overrideGlobal = false): SymId =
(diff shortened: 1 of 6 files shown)


metagn:
Removed global symbol part of the PR as it just made the test pass hexer which isn't tested anyway

PR #383, opened 2025-01-18
nimony: implement `build` + mimalloc support

ringabout:
fixes https://github.com/nim-lang/nimony/issues/357

ref https://github.com/microsoft/mimalloc

- [x] Make hastur download mimalloc as part of the build process. Maybe depend on it via a git submodule.
- [x] Add `.build: "foo.c"` to Nimony so that we can also depend on C code like mimalloc.
- [x] Make use of `.build` in system.nim so we can use a decent memory allocator which offers `allocatedSize()` so that our new super efficient strings and seq can use it.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -8,7 +8,7 @@
 ## Most important task is to turn identifiers into symbols and to perform
 ## type checking.
 
-import std / [tables, sets, syncio, formatfloat, assertions, packedsets, strutils, os]
+import std / [tables, sets, syncio, formatfloat, assertions, packedsets, strutils]
 include nifprelude
 import nimony_model, symtabs, builtintypes, decls, symparser, asthelpers,
   programs, sigmatch, magics, reporters, nifconfig, nifindexes,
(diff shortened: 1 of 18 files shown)


src/nimony/sem.nim
```diff
@@ -8,7 +8,7 @@
 ## Most important task is to turn identifiers into symbols and to perform
 ## type checking.
 
-import std / [tables, sets, syncio, formatfloat, assertions, packedsets, strutils]
+import std / [tables, sets, syncio, formatfloat, assertions, packedsets, strutils, os]
```
Araq, line 11: Don't import os here. Instead move the logic that requires it to `semos.nim`.

ringabout:
It seems that the mingw-64 (gcc  11.1.0) shipped by Nim doesn't have `ERROR_COMMITMENT_MINIMUM` as added in `https://sourceforge.net/p/mingw-w64/mailman/message/37409807/`. It works with my local installation (gcc 14)

src/nimony/sem.nim
ringabout: Sure

Araq:
Fork the project, patch it to use the integer value for ERROR_COMMITMENT_MINIMUM.

src/nimony/semos.nim
```diff
@@ -35,6 +35,18 @@ proc binDir*(): string =
...
+
+proc fileExists2*(f: string): bool =
+  result = os.fileExists(f)
+
+proc toAbsolutePath*(f: string, dir: string): string =
+  if f.isAbsolute:
+    result = f
+  else:
+    result = dir / f
+
```
Araq, line 49: Not like that... But fine for now.

Issue #381, opened 2025-01-17
Nimony: add Converters


Araq:
Add support for Nim's converters to Nimony. This should use some effective caching logic and attach converters to canonicalized types so that overload resolution doesn't slow down too much. This feature needs to work with generics out of the box so that something like

```nim

converter toOpenArray*[I, T](x {.byref.}: array[I, T]): openArray[T] {.inline.} =
  if len(x) == 0:
    openArray[T](a: nil, len: 0)
  else:
    openArray[T](a: addr(x), len: len(x))

```

becomes possible.

PR #384, opened 2025-01-18
implement converters in sigmatch

metagn:
refs #381

Implemented with a converter section in the index like `(converter (kv <type> <converter>))`, was just my first guess for what to do, maybe there is a better way. The converter symbol is stored rather than the offset which I'm hoping is fine, I tried offsets first and couldn't figure out what to do with them.

Attaches to nominal return types for now skipping ref/ptr like type bound ops, both because the algorithm already exists and just storing symbol IDs in the index is simpler. Unlike type bound ops, attachment is required to use converters, so this means only converters to nominal types work. An idea for structural types would be to store the root magic but still skip magics like ref/ptr, so `array[I, T]` and `ref array[3, int]` both attach to `array` but `T` or `ref T` cannot be attached to anything. For converters that can't be attached to anything we could also attach them to `.` and always consider these.

Only generics are left but looks like a can of worms

### DIFF
--- a/src/hexer/expander.nim
+++ b/src/hexer/expander.nim
@@ -814,7 +814,7 @@ proc traverseExpr(e: var EContext; c: var Cursor) =
         inc nested
       of TupleConstrX:
         traverseTupleConstr e, c
-      of CmdX, CallStrLitX, InfixX, PrefixX:
+      of CmdX, CallStrLitX, InfixX, PrefixX, HcallX:
         e.dest.add tagToken("call", c.info)
         inc c
         inc nested
--- a/src/lib/nifindexes.nim
+++ b/src/lib/nifindexes.nim
@@ -99,8 +99,40 @@ proc processForChecksum(dest: var Sha1State; content: var TokenBuf) =
     else:
       inc n
 
-proc createIndex*(infile: string; buildChecksum: bool;
-    hookIndexMap: Table[string, seq[(SymId, SymId)]]) =
+type IndexSections* = object
+  hooks*: Table[string, seq[(SymId, SymId)]]
+  converters*: seq[(SymId, SymId)]
+
+proc getSection(tag: TagId; values: seq[(SymId, SymId)]; symToOffsetMap: Table[SymId, int]): TokenBuf =
+  let KvT = registerTag "kv"
+
+  result = default(TokenBuf)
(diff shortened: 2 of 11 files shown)


Araq:
Maybe it's better not to do that in sigmatch but in sem.nim instead. Outline:

```nim

const
  ConverterSecretName = "`converter`"

proc converterKey(c: var SemContext; n: TypeCursor): string =
  var n = n
  while n.kind in {MutT, SinkT, PtrT, RefT, OutT, InvokeT}:
    inc n
  result = ConverterSecretName & toString(n, false)

proc converterMatch(c: var SemContext; m: var Match; f, a: TypeCursor;
                    apos: int; info: PackedLineInfo): bool =
  # we speed this up by checking if type `a` can have any converter attached to it:
  let k = converterKey(c, a)
  var converters: seq[SymId] = @[]

  let localConverters = c.m.iface.getOrDefault(k)
  for defId in items localConverters:
    converters.add defId

  for imported in items c.importedModules:
    let candidates = c.p[imported].iface.getOrDefault(k)
    for defId in candidates:
      converters.add defId

  var orig = NoSymId
  result = false
  for conv in items converters:
    # f(a)
    # --> f(conv(a)) ?
    # conv's return type must match `f`.
    # conv's input type must match `a`.
    let retType = returnType(conv)
    let inputType = firstParam(conv)

    # first match the input argument of `conv` so that the unification algorithm works as expected:
    let res = typeRel(inputType, a)
    var resB = Different
    case res
    of Same, Subtype:
      resB = typeRel(f, retType)
    of Generic:
      let retTypeInst = instReturnType(retType, context.bindings)
      resB = typeRel(f, retTypeInst)
    of Different:
      discard
    if resB != Different:
      if orig == NoSymId:
        orig = conv

        m.hiddenCalls[apos] = r
        inc m.usedConverters

        m.argTypes[apos] = f

        result = true
      else:
        # ambiguous match. For now don't report:
        result = false
        break
```

To be called from `resolveOverloads if idx < 0`.

metagn:
Going to see if I can rearrange it

metagn:
There is an off chance this will come up again but #399 is fairly ahead to the point that it's probably better to redo it after rather than continue this, so closing

PR #393, opened 2025-01-20
nimony: integrate derefs

### LINKED ISSUE #386: Integrate derefs.nim as the last step of the pipeline of Nimony
- [ ] sigmatch.nim must not produce the `HaddrX` nodes, that job is done by derefs.nim
- [ ] Debug the logic.
- [ ] Write basic tests: 
  - cannot assign value to `const`
  - cannot pass a call `f()` to a `var` parameter
  - `proc g(): var int; let x = g()` must deref the `g()` call and x's type must skip the `var` modifier, `x`'s type must be `int`

ringabout:
fixes #386

- [x] sigmatch.nim must not produce the `HaddrX` nodes, that job is done by derefs.nim
- [x] Debug the logic.
- [x] Write basic tests: 
  - [x] cannot assign value to `const`
  - [x] cannot pass a call `f()` to a `var` parameter
  - [x] `proc g(): var int; let x = g()` must deref the `g()` call and x's type must skip the `var` modifier, `x`'s type must be `int`


src/nimony/sem.nim
```diff
@@ -4949,6 +4949,14 @@ proc semcheck*(infile, outfile: string; config: sink NifConfig; moduleFlags: set
       c.dest.copyTree res.decl
   instantiateGenericHooks c
   wantParRi c, n
+
+  if reportErrors(c) == 0:
+    var final = move c.dest
+    var finalBuf = beginRead(final)
+    c.dest = injectDerefs(finalBuf)
```
Araq, line 4956: You need to report errors for this one as well... right?

ringabout: Yeah, It is. Errors from `deref` should be caught by the `reportErrors` below. `derefs` doesn't work well with erroneous code from the preceding phase. This stops injecting derefs if there are already errors.

PR #394, opened 2025-01-21
nimony: implements `emit`

ringabout:
Supports

```nim
{.emit: "int s = 12;".}
```

and
```nim
{.emit: [s, " = ", 3, ";"].}
```

### DIFF
--- a/src/hexer/nifcgen.nim
+++ b/src/hexer/nifcgen.nim
@@ -14,6 +14,10 @@ import typekeys
 import ".." / nimony / [nimony_model, programs, typenav, expreval, xints, decls]
 import basics, pipeline
 
+type
+  ExprFlag = enum
+    efKeepStringLit
+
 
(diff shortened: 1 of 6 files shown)


src/hexer/nifcgen.nim
```diff
@@ -1078,7 +1085,12 @@ proc traverseStmt(e: var EContext; c: var Cursor; mode = TraverseAll) =
       inc c
       e.loop c:
         traverseExpr e, c
-    of EmitS, AsgnS, RetS:
+    of EmitS:
+      e.dest.add c
+      inc c
+      e.loop c:
+        traverseExpr e, c, {efKeepStringLit}
```
Araq, line 1092: If you special case `c.kind == StringLit` here in this loop you can avoid the efKeepStringLit flag.

ringabout: okay

Issue #398, opened 2025-01-21
Integrate duplifier.nim into hexer


Araq:
Must be run after iterator inlining and before `xelim`.

PR #403, opened 2025-01-22
Integrate duplifier.nim into hexer

ringabout:
ref #398


ringabout:
This PR makes existing tests work with `duplifier` phase. It still needs `ref` runtime support to work, which shall be solved in the future PRs

Issue #408, opened 2025-01-23
Integrate destroyer.nim into hexer


Araq:

Araq:
Must run after "xelim" so that it is easy for the destroyer.nim to inject destructors.

PR #409, opened 2025-01-24
Integrate destroyer.nim into hexer

ringabout:
ref #408

### DIFF
--- a/src/hexer/destroyer.nim
+++ b/src/hexer/destroyer.nim
@@ -163,7 +163,6 @@ proc trScope(c: var Context; body: var Cursor) =
       inc body
       while body.kind != ParRi:
         tr c, body
-      c.dest.addParRi()
       inc body
     else:
       tr c, body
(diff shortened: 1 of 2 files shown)


src/hexer/destroyer.nim
```diff
@@ -204,10 +204,10 @@ proc trProcDecl(c: var Context; n: var Cursor) =
   c.dest.addParRi()
 
 proc trNestedScope(c: var Context; body: var Cursor; kind = Other) =
-  var bodyScope = createNestedScope(kind, c.currentScope, body.info)
-  swap c.currentScope, bodyScope
```
ringabout, line 208: Because `Scope` is a value, it formed a self-cycle because `c.parent` always points to `c.currentScope`

src/hexer/pipeline.nim
```diff
@@ -23,10 +23,14 @@ proc transform*(c: var EContext; n: Cursor; moduleSuffix: string): TokenBuf =
   endRead(n0)
 
   shrink(n1, n1.len-1)
-  n1.add ctx[].dest
+  n1.add move(ctx[].dest)
```
Araq, line 26: I think you need to do this after `injectDestructors`

PR #412, opened 2025-01-27
transform unoverloadable magic calls without typechecking args

metagn:
This is so we can have more control over the typechecking of `typeof`, `compiles` etc for their special cases in the future (edit: `compiles` doesn't seem to affect semchecking except maybe `compiles do: import ...`?). The analog in the old compiler is the [early evaluation of some magics](https://github.com/nim-lang/Nim/blob/95b1dda1db86881f0119bd319380790fbd594ac5/compiler/semexprs.nim#L2577) before overloading. Maybe we could still walk back error tokens for these magics as `declared`/`defined` already do though.

This is pre-emptive so we could also close it if we don't need it for now.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1499,7 +1499,7 @@ proc semCall(c: var SemContext; it: var Item; flags: set[SemFlag]; source: Trans
     while true:
       case syms.kind
       of Symbol:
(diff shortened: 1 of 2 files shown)


Araq:
- Extract it to a helper proc.
- Disable it unless --compat is used (or whatever we called that switch)

PR #415, opened 2025-01-27
generate set ops

metagn:
Seems like it will need a temp generation/`cpsStmts` style mechanism to implement properly


Araq:
> Seems like it will need a temp generation/cpsStmts style mechanism to implement properly

You can produce `ExprX` nodes and rely on the xelim pass to transform it properly. Well you need to transform it earlier in the pipeline then but we need something like that for `MyObjectRef(field....)` already.

PR #421, opened 2025-01-28
patches for stringimpl

metagn:
1. `[]` for `ptr UncheckedArray`, just maps to `Pat`, not sure about `UncheckedArray` in general
2. `min`/`max` implementation, needs 2 bugfixes:
3. `not`/`or`/`and` produce `bool` type for their `Item`
4. xelim fixes: trivial `or`/`and` copy to the target and not the dest, locals build their target on the dest and not in their own declaration

Procvar fixes:

5. Generate proc types as named types in nifcgen
6. Allow proc types where `params` types are expected across the codebase and skip to the params (`skipProcTypeToParams`)

Other codegen fixes:

7. Generate call convention pragma for procs and proc types
8. Generate `suf` cstring literals
9. `importc`/`exportc` etc without arguments emit their symbol name as an argument since codegen expects an argument (`CrucialPragma` changes)

Sem changes, possibly more impactful:

10. Calls generate a full symchoice for their callee symbols instead of just the innermost symbol so a shadowed proc can still be called (`FindOverloads`, `semIdent`/`semQuoted` changes) 
11. `rootOf` when used to check addressability in `addr` allows pointer indirection

Can split if necessary

### DIFF
--- a/lib/std/system/stringimpl.nim
+++ b/lib/std/system/stringimpl.nim
@@ -134,7 +134,6 @@ proc makeAllocated(s: var string; newLen: int) =
 
 proc add*(s: var string; part: string) =
   let len = s.len
-  discard part.len
   let newLen = len + part.len
   if not isAllocated(s):
     makeAllocated s, newLen
--- a/lib/std/system/stringimpl.nim
+++ b/lib/std/system/stringimpl.nim
@@ -29,7 +29,7 @@ proc `=destroy`*(s: string) {.exportc: "nimStrDestroy", inline.} =
   if isAllocated(s): dealloc(s.a)
 
 template safeCopyMem(dest: var string; src: string; len, allocated: int) =
-  if dest.a != StrData(nil):
+  if dest.a != nil:
     copyMem dest.a, src.a, len
     dest.i = len shl LenShift
   else:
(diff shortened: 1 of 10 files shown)


lib/std/system/stringimpl.nim
```diff
@@ -29,7 +29,7 @@ proc `=destroy`*(s: string) {.exportc: "nimStrDestroy", inline.} =
   if isAllocated(s): dealloc(s.a)
 
 template safeCopyMem(dest: var string; src: string; len, allocated: int) =
-  if dest.a != nil:
+  if dest.a != StrData(nil):
```
Araq, line 32: The compiler must accept this without the type conversion.

metagn: Already does apparently, oops

metagn:
The only remaining problem is that `ashr` doesn't exist in nifc, could be transformed to `shr` as in: https://github.com/nim-lang/Nim/blob/cab3342a2dbe5bbd62aeed0b0436e34a646fa741/compiler/ccgexprs.nim#L757. Manually removing the ashrs gives `undefined reference to alloc/copyMem` at link time which is expected.

Araq:
This ProcT hack is getting out of hand. How about we fix Nifler that it produces `(Proctype (params ...) retType pragmas)` and then for type matching purposes it always points to the `(params ...)` part so that ordinary procs can be matched as easily. And then we make `ParamsS` not a "substructure kind" but a typekind as it indicates a proctype. And while we are at it, parameters should not be able to be dotTokens anymore...

metagn:
The only caveat I can think of is that the routine symbol pragmas i.e. `importc` may need to be differentiated from the type pragmas, maybe the original parsed pragmas can start in the routine declaration pragmas then the proc type pragmas can be populated based on it. Type declarations also have separate pragmas from proc type pragmas.

Araq:
It's mildy annoying but we already have the logic to extract the calling convention out of a more general pragma section and I think we should keep it this way.

Issue #351, opened 2025-01-11
Implement bitsets


Araq:
Size computation already available in evalexprs.nim.

One way to implement these rather easily:

```nim

func `[]`[T](a: set[T]; i: int): var uint8 {.magic: "AccessSetAsByteArray".}

func `+`*[T](a, b: set[T]): set[T] {.inline, noinit.} =
  for i in 0 ..< sizeof(a): result[i] = a[i] or b[i]

func `*`*[T](a, b: set[T]): set[T] {.inline, noinit.} =
  for i in 0 ..< sizeof(a): result[i] = a[i] and b[i]

func `-`*[T](a, b: set[T]): set[T] {.inline, noinit.} =
  for i in 0 ..< sizeof(a): result[i] = a[i] and not b[i]

func `==`*[T](a, b: set[T]): bool {.inline.} =
  for i in 0 ..< sizeof(a):
    if a[i] != b[i]: return false
  return true

func `<=`*[T](a, b: set[T]): bool {.inline.} =
  for i in 0 ..< sizeof(a):
    if (a[i] and not b[i]) != 0: return false
  return true

func `<`*[T](a, b: set[T]): bool = (a <= b) and not (a == b)
```

PR #428, opened 2025-01-29
generate set ops in desugar pass

metagn:
refs #351, succeeds #415 as described in https://github.com/nim-lang/nimony/pull/415#issuecomment-2615788335

Set types are still only transformed in nifcgen, maybe could generate `dconv` too

* [x] move operands to temps if they have side effects
* [ ] implement set ops for array sets (can be done later, some might need compilerprocs)


Araq:
Compiler procs are just symbols you can access like `pool.syms.getOrIncl("procname.0." & SystemModuleSuffix)`. We then need to add these to system.nim, of course.

metagn:
Undrafting, remaining unimplemented stuff could be implemented later, might still introduce bugs

PR #440, opened 2025-01-31
implement inc/dec as desugared magic

metagn:
Tried to implement as template, doesn't work because the use of `+`/`-` needs a concept and concepts still seem to have some issues, even `x = x` does not work (will still look at these issues). Alternatively it could be an untyped template. In any case this is more faithful, could be revisited later.

A problem with this approach is that mutability checking has to be special cased for these magics as well as `incl`/`excl`, which is not done yet

### DIFF
--- a/lib/std/system.nim
+++ b/lib/std/system.nim
@@ -104,6 +104,20 @@ template `[]=`*[T](x: ptr T, val: T) =
 template `[]=`*[T](x: ref T, val: T) =
   (x[]) = val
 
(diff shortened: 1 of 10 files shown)


Araq:
> A problem with this approach is that mutability checking has to be special cased for these magics as well as incl/excl, which is not done yet

Yeah, this solution is messy as too many steps of the pipeline then need to understand the `IncX` special case. We need a template for this as well as some way to say "eval once".

metagn:
Trying inline procs

Issue #431, opened 2025-01-31
Backend C compiler fails to find the include file


demotomohiro:
test.nim:
```nim
type Foo {.importc: "Foo", header: "test.h".} = object
```
test.h:
```c
struct Foo {};
```

```console
$ nimony c test.nim 
C compiler says "test.h" is not found.
```

In generated .makefile, there is no `-I` option in the C compiler command line.

Araq:
Not sure that's a bug, does it work with Nim 2 this way?

demotomohiro:
> Not sure that's a bug, does it work with Nim 2 this way?
Yes. Nim 2 simply always passes the project directory as include directory:
https://github.com/nim-lang/Nim/blob/0861dabfa70f40f00c5c95d58c344ad5e0fd19ec/compiler/extccomp.nim#L612

nimony and NIFC should work in the same way to keep compatibility with Nim 2?
Or only include directories specified by the user with the command line options or pragma are passed to backend C compiler?

Araq:
In **this case**, if Nim 2 does it, so should Nimony.

demotomohiro:
I found related PR and issue:
https://github.com/nim-lang/Nim/pull/10012
https://github.com/nim-lang/Nim/issues/8408

Araq:
@demotomohiro Should probably be fixed in nimony/deps.nim

mrgaturus:
i think to make relative `{.header.}` include more effective it should have `$path` support like `{.build.}` has, right now the `-I` workaround only works for current project but doesn't work when the header is on another `--path` (nimble module for example).
the `-I` should be keep to ensure compatibility with Nim 2.x

this workaround make it works when the header is on another `--path` that is not the current project:
```nim
func includePath(): string {.compileTime.} =
  result = currentSourcePath()
  var l = result.len
  # Remove File Name
  while l > 0:
    let c = result[l - 1]
    if c == '/' or c == '\\':
      break
    # Next Char
    dec(l)
  # Remove Chars
  result.setLen(l)

{.push header: includePath() & "foo.h".}
# include stuff from header
{.pop.}
```

PR #448, opened 2025-02-01
nimony: add basic `${path}` to `{.header.}` for relative c include

mrgaturus:
- add `replaceSubs` helper to `semos.nim` to replace a pattern by an absolute path
- add basic `${path}` and `${nifcache}`support to `{.header.}` + some tests
- add a `-I` of root module path relative to nifcache in final makefile for nim 2.x backwards compatibility (see https://github.com/nim-lang/Nim/pull/10012)

fixes https://github.com/nim-lang/nimony/issues/431

### DIFF
--- a/src/nimony/semos.nim
+++ b/src/nimony/semos.nim
@@ -6,7 +6,7 @@
 
 ## Path handling and `exec` like features as `sem.nim` needs it.
 
-from strutils import replace, contains
+from std / strutils import multiReplace
 import std / [tables, sets, os, syncio, formatfloat, assertions]
 include nifprelude
 import ".." / lib / nifchecksums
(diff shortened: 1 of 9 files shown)


Araq:
- Use relative paths if possible so that if we add the produced makefiles to the testsuite they can actually work.
- The signature should be `replaceSubs*(f, currentFile, nifDir: string): string`
- Use strutils.multiReplace in replaceSubs's implementation.

src/nimony/semos.nim
```diff
@@ -6,6 +6,7 @@
 
 ## Path handling and `exec` like features as `sem.nim` needs it.
 
+from strutils import replace, contains
```
ringabout, line 9:
```suggestion
from std/strutils import replace, contains
```

PR #449, opened 2025-02-01
fixes #437

### LINKED ISSUE #437: Var return type with expression style return fails to compile
```nim
proc access(x: var int): var int =
  x # works with `result = x` or `return x` (producing strange nif code though), non-var works too
```

Deadlocks during nimsem stage

Araq:


Araq:
@SirOlaf wasn't that hard to fix :P

SirOlaf:
I guess it wasn't :)
Makes sense now that I see it

PR #458, opened 2025-02-03
hexer pipeline needs to be invoked for inlined routines coming from a different module

### LINKED ISSUE #443: hexer pipeline needs to be invoked for inlined routines coming from a different module

ringabout:
fixes #443

### DIFF
--- a/src/hexer/nifcgen.nim
+++ b/src/hexer/nifcgen.nim
@@ -1191,6 +1191,23 @@ proc importSymbol(e: var EContext; s: SymId) =
     if c.stmtKind == TypeS:
       traverseTypeDecl e, c
     else:
+      if isRoutine(c.symKind):
+        var pragmas = asRoutine(c).pragmas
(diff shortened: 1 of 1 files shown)


src/hexer/nifcgen.nim
```diff
@@ -1184,13 +1184,33 @@ proc traverseStmt(e: var EContext; c: var Cursor; mode = TraverseAll) =
...
+proc transformInlineRoutines(e: var EContext; c: var Cursor) =
+  var swapped = createTokenBuf()
+  swap e.dest, swapped
+
+  var toTransform = createTokenBuf()
+  toTransform.copyIntoKind StmtsS, c.info:
+    takeTree(toTransform, c)
+  var c0 = beginRead(toTransform)
+  var dest = transform(e, c0, e.main)
+  c = beginRead(dest)
```
Araq, line 1196: This is not correct, the underlying buffer does not live enough. I think you merely got lucky that the memory is not reused too soon.

PR #466, opened 2025-02-05
experiment: allow nimony compile projects outside compiler folder

mrgaturus:
- remove hardcoded nifcache path prefixes
- make nimony be able compile projects outside compiler folder
- allow locate a custom nifcache path using a `--nifcache:PATH` command line argument (`nimony` only)
- makefiles, nif files, and c files have paths relative to nifcache folder (ideal for tests)

this is an experiment so i don't think this would be a proper solution right now, at least this experiment could be useful to a proper implementation of this in a future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+bin/
 nimcache/
 nimblecache/
 htmldocs/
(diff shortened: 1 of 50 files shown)


Araq:
Bring this over the finish line please.

mrgaturus:
in this experiment you don't need cd to nifcache first, your example works as well. i mean the fact all paths inside nifcache files are relative to the nifcache path, so you will see `../tests/nimony` instead `tests/nimony` for example

Araq:
I did it my way, sorry.

mrgaturus:
no problem, it wasn't supposed to be merged just an experiment

Issue #450, opened 2025-02-01
integrate injectConstParamDerefs into hexer's pipeline


Araq:
- [ ] nifcgen also needs to understand `lent T` auto parameters.

PR #467, opened 2025-02-05
integrate injectConstParamDerefs into hexer's pipeline

ringabout:
ref #450

- [x] the codegen's `maybeByConstRef` logic

### DIFF
--- a/src/hexer/nifcgen.nim
+++ b/src/hexer/nifcgen.nim
@@ -509,7 +509,7 @@ proc maybeByConstRef(e: var EContext; c: var Cursor; pragmas: Cursor) =
     paramBuf.add param.pragmas
     copyIntoKind paramBuf, PtrT, param.typ.info:
       paramBuf.add param.typ
-    paramBuf.add param.val # TODO: default parameters haven't been implemented yet
+    paramBuf.add param.val
     paramBuf.addParRi()
     var paramCursor = beginRead(paramBuf)
     traverseLocal(e, paramCursor, "param", TraverseSig)
--- a/src/hexer/nifcgen.nim
+++ b/src/hexer/nifcgen.nim
@@ -500,7 +500,7 @@ proc traverseType(e: var EContext; c: var Cursor; flags: set[TypeFlag] = {}) =
 
 proc maybeByConstRef(e: var EContext; c: var Cursor) =
   let param = asLocal(c)
-  if passByConstRef(param.typ, param.pragmas, sizeof(int)):
+  if passByConstRef(param.typ, param.pragmas, e.bits):
     var paramBuf = createTokenBuf()
     paramBuf.add tagToken("param", c.info)
     paramBuf.add param.name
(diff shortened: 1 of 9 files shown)


src/hexer/nifcgen.nim
```diff
@@ -498,7 +499,26 @@ proc traverseType(e: var EContext; c: var Cursor; flags: set[TypeFlag] = {}) =
...
+  let param = asLocal(c)
+  if passByConstRef(param.typ, pragmas, sizeof(int)):
+    var paramBuf = createTokenBuf()
+    paramBuf.add tagToken("param", c.info)
+    paramBuf.add param.name
+    paramBuf.add param.exported
+    paramBuf.add param.pragmas
+    copyIntoKind paramBuf, PtrT, param.typ.info:
+      paramBuf.add param.typ
+    paramBuf.add param.val # TODO: default parameters haven't been implemented yet
```
Araq, line 512: There are no default parameters in NIFC!

src/hexer/nifcgen.nim
```diff
@@ -633,7 +656,7 @@ proc traverseProc(e: var EContext; c: var Cursor; mode: TraverseMode) =
     inc c
     skip c # skip return type
   else:
-    traverseParams e, c
+    traverseParams e, c, procDecl.pragmas
```
Araq, line 659: But what is in these pragmas that affect pass-by-const-ref?

src/hexer/nifcgen.nim
```diff
@@ -498,7 +499,26 @@ proc traverseType(e: var EContext; c: var Cursor; flags: set[TypeFlag] = {}) =
   else:
     error e, "type expected but got: ", c
 
-proc traverseParams(e: var EContext; c: var Cursor) =
+proc maybeByConstRef(e: var EContext; c: var Cursor; pragmas: Cursor) =
+  let param = asLocal(c)
+  if passByConstRef(param.typ, pragmas, sizeof(int)):
```
Araq, line 504: For cross-compiling `sizeof(int)` would be wrong. Hexer needs to understand the `--bits` command line option and Nimony needs to pass it.

src/hexer/pipeline.nim
```diff
@@ -9,7 +9,30 @@
 
 include nifprelude
 
-import basics, iterinliner, desugar, xelim, duplifier, lifter, destroyer
+import ".." / nimony / [nimony_model, programs, decls]
+import basics, iterinliner, desugar, xelim, duplifier, lifter, destroyer, constparams
+
+proc publishHooks*(n: var Cursor) =
```
Araq, line 15: Looks like an unrelated bugfix?

src/nimony/sizeof.nim
```diff
@@ -180,7 +184,9 @@ proc typeSectionMode(n: Cursor): PragmaKind =
 
 proc passByConstRef*(typ, pragmas: Cursor; ptrSize: int): bool =
   let k = typ.typeKind
-  if k in {SinkT, MutT, OutT}:
+  if k in {SinkT, MutT, OutT, VoidT, VarargsT, OrT, AndT, NotT,
```
Araq, line 187: Not quite, types like `OrT` should produce an internal error instead.

src/hexer/nifcgen.nim
ringabout: Yeah, it didn't handle default parameters from the frontend because that's implemented yet
```nim
proc foo(x: int = 1) =
  discard

foo()
```
codegen error

Araq: Just ignore `param.val` here, why pretend that NIFC has it?

src/hexer/nifcgen.nim
ringabout: I forgot `byref` is attached to types

src/hexer/pipeline.nim
ringabout: It is needed because hooks are not in the index files, which cause problems when loading the symbol

src/hexer/nifcgen.nim
ringabout: okay

PR #478, opened 2025-02-07
add separate int conv matches

### LINKED ISSUE #477: Comparing int32 value with int causes ambiguous call error
testints.nim:
```nim
var x: int32
var y: int

# Following code compiles without errors.
discard x == y
discard y == 123.int32
discard 123.int32 == y
discard 123.int == 123.int
discard 123.int32 == 123.int32

# Following code causes compile error: "Error: ambiguous call"
discard 123.int32 == 123
discard 123 == 123.int32
discard x == 123
```

```console
$ ./bin/nimony c mytests/testints.nim 
nif/bin/nimsem  --isSystem m nifcache/sys9azlf.1.nif nifcache/sys9azlf.2.nif nifcache/sys9azlf.2.idx.nif
nif/bin/nimsem  m nifcache/tes9vq2bh.1.nif nifcache/tes9vq2bh.2.nif nifcache/tes9vq2bh.2.idx.nif
nif/src/nimony/nimsem.nim(120) nimsem
nif/src/nimony/nimsem.nim(117) handleCmdLine
nif/src/nimony/nimsem.nim(50) singleModule
nif/src/nimony/sem.nim(5443) semcheck
nif/src/nimony/sem.nim(746) semStmt
nif/src/nimony/sem.nim(5159) semExpr
nif/src/nimony/sem.nim(3455) semDiscard
nif/src/nimony/sem.nim(5233) semExpr
nif/src/nimony/sem.nim(1609) semCall
nif/src/nimony/sem.nim(1441) resolveOverloads
nif/src/nimony/sembasics.nim(159) buildErr
testints.nim(12, 19) Error: ambiguous call
ambiguous call
make: *** [nifcache/tes9vq2bh.makefile:6: nifcache/tes9vq2bh.2.idx.nif] Error 1
FAILURE: make -f nifcache/tes9vq2bh.makefile
```

metagn:
fixes #477

Supposed to match the old compiler, which reuses `intConvCosts` instead of separate `intLitCosts`. `intConvCosts` seems to mostly exist so a `int32` matches `int` better than `float`. Could rename `intLitCosts` to `literalCosts` if it makes sense to generalize it.

The order of `cmpMatches` had to be reversed into checking the most expensive conversions first, unlike the order in the old compiler `cmpCandidates` because the `equalMatches`/`convMatches` etc in the old compiler complement/are negatives of the costs.

### DIFF
--- a/tests/nimony/sysbasics/tintlitmatch.nim
+++ b/tests/nimony/sysbasics/tintlitmatch.nim
@@ -5,3 +5,19 @@ bar(123)
 proc baz(x: int16) = discard
 proc baz(x: int) = discard
 baz(123)
+
+# issue #477
+var x: int32
+var y: int
+
+# Following code compiles without errors.
+discard x == y
+discard y == 123.int32
+discard 123.int32 == y
+discard 123.int == 123.int
+discard 123.int32 == 123.int32
+
+# Following code causes compile error: "Error: ambiguous call"
+discard 123.int32 == 123
+discard 123 == 123.int32
+discard x == 123
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -56,7 +56,7 @@ type
     skippedMod: TypeKind
     argInfo: PackedLineInfo
     pos, opened: int
-    inheritanceCosts, intCosts: int
+    inheritanceCosts, intLitCosts, intConvCosts, convCosts: int
     returnType*: Cursor
     context: ptr SemContext
     error: MatchError
@@ -513,7 +513,14 @@ proc matchIntegralType(m: var Match; f: var Cursor; arg: Item) =
     else:
       m.args.addParLe HconvX, m.argInfo
       m.args.addSubtree forig
-      inc m.intCosts
+      if isIntLit:
+        if f.typeKind == FloatT:
+          inc m.convCosts
+        else:
+          inc m.intLitCosts
+      else:
+        # sameKind is true
+        inc m.intConvCosts
       inc m.opened
   else:
     m.error InvalidMatch, f, a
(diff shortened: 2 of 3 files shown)


tests/nimony/sysbasics/tintlitmatch.nim
```diff
@@ -5,3 +5,19 @@ bar(123)
...
+discard x == y
+discard y == 123.int32
+discard 123.int32 == y
+discard 123.int == 123.int
+discard 123.int32 == 123.int32
+
+# Following code causes compile error: "Error: ambiguous call"
+discard 123.int32 == 123
+discard 123 == 123.int32
+discard x == 123
```
blackmius, line 23: these errors are bit annoying

cant compiler automatically convert literals without type provided to match the second operand?

metagn: There is no error with this, I just copied the code from the issue

blackmius: otherwise a lot of code doing bit manipulations is require adding postfixes to literals

``` nim
proc writeVarInt*(buf: Buffer, val: uint64) =
    if val <= 0x3F:
        write[uint8](buf, val.uint8)
    elif val <= 0x3FFF:
        write[uint16](buf, val.uint16 or 0x4000)
    elif val <= 0x3FFFFFFF:
        write[uint32](buf, val.uint32 or 0x80000000'u32)
    elif val <= 0x3FFFFFFFFFFFFFFF'u64:
        write[uint64](buf, val or 0xC000000000000000'u64)
    else:
        raise newException(ValueError, "Integer is too big")
```

metagn: Again, this PR would go toward making that work, that is a separate bug with the old compiler not considering those as int literals due to giving them `uint` types

Issue #470, opened 2025-02-06
make empty seq literals work


metagn:
Could use something like `--experimental:inferGenericParams` and depend on top down type inference, but this probably wouldn't be able to handle:

```nim
proc foo(x: seq[int]) = discard

foo(@[])
```

So we might still need to use an `Empty` type as in the old compiler but make sure it always gets checked.

This case already doesn't work in the old compiler:

```nim
proc foo(x: seq[int]) = discard
proc bar[T](x: seq[T]): seq[T] = x

foo(bar(@[]))
```

metagn:
Thought about it for a while, best I've come up with is:

* Call arguments whitelist empty array/set literals. Semchecking them gives the type `auto`
* If necessary, a call argument with type `auto` is checked to be an empty array/set literal and allowed
* Array/set types check for these in sigmatch and convert the empty literal argument, probably by adding `Hconv` as we should be able to handle stuff like `set[int8]({})` anyway

For seqs, provided they stay completely independent from the compiler:

* If an array/set type has an unbound generic element type and matches to an empty literal, a flag is set for the entire match that it is an empty literal call
  * Can be whitelisted by call arguments again, otherwise producing an error "cannot infer type of empty literal"
* Empty literal calls do not instantiate procs or expand templates, they stay as calls and have the type `auto`
* If a call argument has type `auto` after semchecking and is a call with an empty literal argument, the argument item is marked as an empty literal call
* When empty literal call arguments are matched to a parameter in sigmatch, a flipped match is performed between the return type of the call symbol and the (instantiated?) parameter type:
* If the full call match succeeds, the inferred parameters from the flipped match are used to call `sigmatch` on the empty literal call itself and the argument becomes the fully instantiated empty literal call (probably in `addArgsInstConverters`)

This does not have to stack with converters as it already doesn't in the old compiler.

What calls are allowed to be "empty literal calls" can be restricted as much as necessary given they match `@`. They might have to consider multiple overloads though.

Instantiating the parameter type to perform the flipped match imply that should be done in sem and not in sigmatch, but it doesn't really make sense to do this after the matches fail like converters given that the empty literal call arguments will always fail to match if not handled. Doing them in sigmatch would mean we still need `context.instantiateType` etc though, but maybe this is an OK exception.

[...]

PR #479, opened 2025-02-07
separate typed and untyped array/set literals, make case always use `ranges`

metagn:
Refs #470, empty array/set literals need this to work in hexer.

Not sure if there is a reason why `case of` branches didn't change their selectors from `set` to `ranges` until `nifcgen`, would have had to make them always `curly` here which requires `curly` to be traversed without erroring in the rest of hexer.


Araq:
In a follow-up PR feel free to also adapt Nifler to produce `ranges` instead of `set` for case statements. Or does it use these already?

metagn:
Yes, nifler generates `ranges` now

Issue #432, opened 2025-01-31
"Error: undeclared identifier" message doesn't show the identifier


demotomohiro:
testundec.nim
```nim
undeclaredproc()
```

```console
$ nimony c testundec.nim
testundec.nim(1, 15) Error: undeclared identifier
```

Expected output:
```console
$ nimony c testundec.nim
testundec.nim(1, 15) Error: undeclared identifier: undeclaredproc
```

When there is an undeclared procedure, nimony prints the error message without the identifier that causing the error.
I think that identifier should be printed so that you can easily find the mistake.

Araq:
Fixed.

PR #488, opened 2025-02-08
Print the identifier when an undeclared proc is called

demotomohiro:
This fixes a part of https://github.com/nim-lang/nimony/issues/432
This PR prints the identifier only when there is an undeclared proc call and an ambiguous call.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1423,11 +1423,13 @@ proc resolveOverloads(c: var SemContext; it: var Item; cs: var CallState) =
     var errorMsg: string
     if idx == -2:
       errorMsg = "ambiguous call: '"
-      errorMsg.add pool.strings[cs.fnName]
+      if cs.fnName != StrId(0):
+        errorMsg.add pool.strings[cs.fnName]
       errorMsg.add "'"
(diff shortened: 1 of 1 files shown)


src/nimony/sem.nim
```diff
@@ -1436,7 +1438,9 @@ proc resolveOverloads(c: var SemContext; it: var Item; cs: var CallState) =
         errorMsg.add "\n"
         addErrorMsg errorMsg, m[i]
     else:
-      errorMsg = "undeclared identifier"
+      errorMsg = "undeclared identifier: '"
+      errorMsg.add pool.strings[cs.fnName]
```
Araq, line 1442: Likewise.

src/nimony/sem.nim
```diff
@@ -1422,7 +1422,9 @@ proc resolveOverloads(c: var SemContext; it: var Item; cs: var CallState) =
     skipParRi it.n
     var errorMsg: string
     if idx == -2:
-      errorMsg = "ambiguous call"
+      errorMsg = "ambiguous call: '"
+      errorMsg.add pool.strings[cs.fnName]
```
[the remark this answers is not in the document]
demotomohiro: done

src/nimony/sem.nim
demotomohiro: done

PR #502, opened 2025-02-10
implement empty array/set literal matching

metagn:
refs #470

* When semchecking call arguments, empty array/set literals without an inferred element type are allowed to compile and produce the type `auto`
* Expressions with type `auto` are detected in sigmatch and checked for being empty literals. If so, the param type is checked to be a matching array/set type and the literal is updated to have the param type. If the param type is generic, the match is marked as containing a generic empty literal.
* Then when the call matches, it is checked if the match was marked to contain a generic empty literal, in which case empty literal arguments instantiate their type when being added to the final call node.

### DIFF
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -61,7 +61,7 @@ type
     context: ptr SemContext
     error: MatchError
     firstVarargPosition*: int
-    genericConverter*: bool
+    genericConverter*, genericEmpty*: bool
 
 proc createMatch*(context: ptr SemContext): Match = Match(context: context, firstVarargPosition: -1)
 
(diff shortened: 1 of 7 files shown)


src/nimony/sigmatch.nim
```diff
@@ -705,7 +705,51 @@ proc singleArgImpl(m: var Match; f: var Cursor; arg: Item) =
...
+proc isEmptyLiteral*(n: Cursor): bool =
+  result = n.exprKind in {AconstrX, SetConstrX}
+  if result:
+    var n = n
+    inc n # tag
+    skip n # type
+    result = n.kind == ParRi
+
 proc singleArg(m: var Match; f: var Cursor; arg: Item) =
+  if arg.typ.typeKind == AutoT and isEmptyLiteral(arg.n):
```
Araq, line 717: Move to a `matchEmptyContainer` helper proc.

PR #504, opened 2025-02-10
rename `add` for cursors, fix related bug

metagn:
Bug is in `maybeByConstRef`, `addSubtree` needed to be used.

A lot of these uses are `takeToken` analogs, could also add a version of it for `TokenBuf` to reduce the increased code length in followup.


Araq:
It doubt it's worth the extra typing, you can type `addToken` just as accidentically as `add` when you don't think things through.

metagn:
It's more noticeable when trying to figure out the problem though

metagn:
Going through the uses, almost all of them are correct so I guess there's no reason to change them. Will still add protections in another PR though along with the bugfix

PR #507, opened 2025-02-10
run xelim again after constparams if needed

### LINKED ISSUE #503: `constparams` runs last in hexer pipeline but generates `ExprX`
[here](https://github.com/nim-lang/nimony/blob/a768e9e7abd47c02f38995899778a415ced5e503/src/hexer/constparams.nim#L61)

Not eliminated in nifcgen, reaches nifc which doesn't parse it

metagn:
fixes #503

### DIFF
--- a/src/hexer/constparams.nim
+++ b/src/hexer/constparams.nim
@@ -26,7 +26,6 @@ type
     constRefParams: HashSet[SymId]
     ptrSize, tmpCounter: int
     typeCache: TypeCache
-    needsXelim: bool
 
 when not defined(nimony):
   proc tr(c: var Context; dest: var TokenBuf; n: var Cursor)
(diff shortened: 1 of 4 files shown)


Araq:
No, just run constparams after injectDestructors but before xelim. Always. No special casing required nor running of xelim twice.

metagn:
I was under the impression that injectDestructors depended on xelim

Araq:
No, but even if so, the pipeline is:   injectdestructors -> constparams -> xelim. 

So even if injectdestructors required xelim this order would still work.

metagn:
Depended as in required xelim to run before it, as implied by:

https://github.com/nim-lang/nimony/blob/e8503e5e011e764cce76dd5718f57a44d7e51f64/src/hexer/destroyer.nim#L12-L13

Araq:
I stand corrected. I misremembered.

PR #510, opened 2025-02-11
Fix link errors when include system/stringimpl

demotomohiro:
You still need to add `--passC:-Ivendor/mimalloc/include --passL:-Lvendor/mimalloc/out/release` so that GCC can find mimalloc include file and library file.


ringabout:
You can use functions from https://github.com/nim-lang/nimony/blob/e6e8df559c1e5f50e19c20074316dd0d9c3018d4/lib/std/system/mimalloc.nim

Araq:
You need to wrap the C procs using `size_t` and then offer Nim inline procs that call the C procs converting `int` to `size_t`.

demotomohiro:
@Araq Thank you!
I fixed `copyMem` and `cmpMem` procs.

demotomohiro:
I removed `tests/nimony/sysbasics/tmimalloc.nim` as compiling it causes assertion defect.
Content of the file is identials to `lib/std/system/mimalloc.nim` and it is now included from system module.

PR #521, opened 2025-02-12
Integrate the move analyser into duplifier.nim

### LINKED ISSUE #468: Integrate the move analyser into duplifier.nim

ringabout:
fixes #468


ringabout:
```nim
proc foo =
  var my = [1,2, 3, 4]
  var i = 0
  my[0] = 56

  discard ensureMove(my)
  my[1] = 2

foo()
```
I will check the test case

ringabout:
With the fix, the example above now gives `not the last usage of\3A \28emove my.0\29`

PR #531, opened 2025-02-13
Implement simple default parameters 

ringabout:

PR #536, opened 2025-02-13
test disallowing implicit conversion to generic types

metagn:
refs https://github.com/nim-lang/nimony/pull/531#discussion_r1954485562

Another option might be to disable generic param inference for these or a general rule to only infer generic params included in `m.tvars`.

Type conversions might not be the best way to do this due to the overlap with normal type conversions, same for casts.

### DIFF
--- a/tests/nimony/nosystem/tfib.nim
+++ b/tests/nimony/nosystem/tfib.nim
@@ -15,9 +15,9 @@ type
     proc `-`(x, y: Self): Self
 
 proc fib[T: Fibable](a: T): T =
-  if a <= 2:
-    result = 1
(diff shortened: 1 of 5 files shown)


tests/nimony/nosystem/tfib.nim
```diff
@@ -15,9 +15,9 @@ type
     proc `-`(x, y: Self): Self
 
 proc fib[T: Fibable](a: T): T =
-  if a <= 2:
-    result = 1
+  if a <= T(2):
```
Araq, line 18: I think this implies this feature would have been too annoying in practice. :-)

PR #547, opened 2025-02-14
instantiate and match inserted default values

### LINKED ISSUE #542: generic default parameters
As described [here](https://github.com/nim-lang/nimony/pull/536#issuecomment-2657258248)

metagn:
fixes #542, succeeds #544

The caveat is that inability to infer typevars has to be a post-match error for this to work in the case of `foo6()` rather than an overload match error. This might be better though, the old compiler already behaves this way:

```nim
proc foo[T, U](x: U) = discard
proc foo[T](x: T) = discard

foo(123) # error: ambiguous call
```

`insertedParam` could subsume `genericEmpty` if we add the full default value AST in sigmatch and keep track of which args are inserted without checking for dot tokens, probably by keeping a list of indexes. These modes are mutually exclusive with each other so we could also build a list of argument kinds like `[Normal, Inserted, Normal, Empty, Converter, ...]`.


Araq:
Why is so much special logic required at all? The original idea was "default parameters are simple, we simply emulate it as if they had been written out explicitly". Now it's turned into this mess...

metagn:
1. We need to instantiate generic default values after matching which is why we have any logic in `addArgsInstConverters`, could restrict this to default values that contain generic parameters.
2. We need to mark which arguments are default values so we can instantiate them. This is why dot tokens are added, could be done some other way as in PR description.
3. For the `foo6()` case in the test, i.e. when we have to infer generic parameters from default values of params:
  ```nim
  proc foo[T](x: T = 123) = discard
  foo()
  ```
  we match the instantiated default value to its param to infer the type. Doing this in sigmatch instead would be a problem because:
4. Matching types of default values to their param types in sigmatch would mean generic params could be inferred to themselves if the default value has a generic type depending on the generic params of the proc. We either match it to an instantiated type, which would need sigmatch to call `c.instantiateType`, or we selectively don't match it if it contains generic parameters or something, I don't know what the logic for this would be. We would probably need to distinguish between "generic parameters that belong to the routine" i.e. generic parameters that can have inferred values and "generic parameters in the current generic context" i.e. valid inferred type values but they can be the same if a routine is recursive.

We could do away with 3 as a language feature, but if we don't want to match the instantiated default value to the param type, we would need generic typechecking to generate an expression for the default value that ensures it has the type `T` when instantiated, i.e. `123` becomes `T(123)` or `(ensure 123 T)`, or be strict enough that it is sure the expression matches `T` as in #536. Or we instantiate the full proc type as in #544 which accomplishes this and take the default value from there.

I am not saying this solution is the best with these considered, these are just problems other solutions might have.

Araq:
Considering:

```nim

proc foo[T](x: T = 123) = discard
foo()
```

While we check if `foo()` matches the one foo in question we notice we got too few arguments so we take them from the default values producing `foo(123)` and so `T` is inferred to `int`. Easy peasy. I still don't get where the complexity comes from.

metagn:
If it was

```nim
proc foo[T](x: T = T(123)) = discard
foo()
```

instead, `T` would be inferred to `T`.

metagn:
```nim
proc foo[T](x: T, y: T = T(123)) = discard

foo(123)
```

The expression `T(123)` has type `T` when not instantiated, matching it to `T` which is inferred to `int` would fail in the current sigmatch. If `T` is not inferred yet then it becomes inferred to `T` (itself). If we handle the case where `T` matches itself to not infer anything here, the problem still exists when matching other generic parameters: 

```nim
proc foo[T, U](x: T, y: U = T(123)) = discard

foo(123)
```

Now `U` is being matched to `T`, which is inferred to `int`, but sigmatch does not consider this. Could also be:

```nim
type Obj[T] = object

proc foo[T, U](x: T, y: U = Obj[T]()) = discard

foo(123)
```

Now `U` is being matched to `Obj[T]`. This implies to me that an instantiation is needed, but I don't know.

Araq:
Well ok, so you need to take the default parameter and do a typevar substitution before adding it to the argument list.

metagn:
Then we also need to use `getType` as per @ringabout's comment in `collectDefaultValues`, which needs to be adapted to work in the sem phase, maybe by `TypeCache` having an optional `ptr SemContext` field to delegate to?

Araq:
Why? You have the already resolved typevars, you iterate over the default value's tokens replacing the typevars you know about. You can also disallow `complexParam = Obj[T]()` as it's complex shit that nobody in the real world uses anyway and if it comes up, you can write an overloaded proc instead.

metagn:
We need to get the type of the default value expression if we are inferring generic parameters to it, as in this case again.

```nim
proc foo[T](x: T = 123) = discard
foo()
```

We need the type of `123` to infer `T` to `int`. Hence `getType`. Again we could stop supporting this case but it was brought up.

If you are still not happy with this then I can move on to something else. The problems with the current situation are:

* generic parameters in default values do not get substituted
* even if they are substituted, the default value expression does not go through semchecking again, the extent to which this is an issue varies
* the actual type of the default value also does not get matched to the parameter type, which means it only matches as much as generic typechecking thinks it matches, i.e. `x: T = 123` being used when `T` is inferred to `string` does not currently complain

In my opinion this PR is not too complex considering it deals with these, with the exception of the use of dot tokens which has alternatives. Partially dealing with them doesn't seem like it would be much less complex in comparison either, but I don't know as I've been focusing on correctness. I can still try but I can't guarantee it will look any better.

PR #550, opened 2025-02-14
fixes #537; fixes invalid file causes crashing

### LINKED ISSUE #537: Simple wrong input causes NIFC to crash
```nif

(.nif24)

(stmts
  (proc :foo.escapedproc (params (param :x.p . unsigned\20char.c)) unsigned\20int.c . (stmts
    (ret x.p)
  ))
  4,4,"foo.nim"
  (oo ((())))
)
```

The problem here is that the NIF parser interprets "foo.nim" as an empty filename which is strange but according to the spec. This empty filename then causes a crash later on.

ringabout:
fixes #537

I'm not sure about the meaning of an empty filename in the line info. I suppose it is unintentional, instead of meaning the current filename.

It gives a proper error: `???(4, 5) [Error] expected top level construct but got: "foo.nim"`


src/nifc/codegen.nim
```diff
@@ -148,7 +148,10 @@ proc error(m: Module; msg: string; n: Cursor) {.noreturn.} =
   let info = n.info
   if info.isValid:
     let (file, line, col) = unpack(pool.man, info)
-    write stdout, pool.files[file]
+    if pool.files.hasId(file):
+      write stdout, pool.files[file]
+    else:
+      write stdout, "???"
```
Araq, line 154: No, if the `info.isValid` it should be valid line information, the error happens earlier.

PR #554, opened 2025-02-14
Make tvar and gvar part of Nimony's model

### LINKED ISSUE #551: Make tvar and gvar part of Nimony's model
These are generally very useful for all kinds of static analyses we need to do. NIFC already has them, Nimony needs them too. `tvar` is a threadlocal variable and `gvar` a global variable. But for consistency we also need `tlet` and `glet`, unfortunately.

ringabout:
fixes #551


Araq:
Well sem.nim also needs to produce these tags. Toplevel `var` becomes `gvar` (same for `let`)  and if annotated with `.threadvar` it becomes a tvar/tlet.

PR #555, opened 2025-02-14
nifc: fixes `genCLineDir` for call expressions

ringabout:
don't generate c lineinfo for call expressions


Araq:
Why? All you have to do is ensure that `#line` is on a line of its own, nested expressions cause no trouble otherwise.

PR #564, opened 2025-02-17
integrate seqs

### LINKED ISSUE #422: Integrate seq handling code into system.nim

metagn:
closes #422 (except openarray but works without it)


metagn:
Finally works, test using `assert.h` is just to make sure it works, can switch to `discard` in the short term if it causes problems

Araq:
Well `syncio` also has an `assert`.

metagn:
Did not know it was ready, seems to work

Issue #543, opened 2025-02-14
proper types for fields of invocation object types


metagn:
As described [here](https://github.com/nim-lang/nimony/pull/536#issuecomment-2657258248)

PR #580, opened 2025-02-19
split type inst cache and module type inst declaration list

metagn:
refs #543

The list of type instantiation syms which need their declarations appended to a module is now separated from the mapping of canonical invocations to their instantiated values. Only types which are fully instantiated i.e. without their generic arguments containing generic parameters themselves are appended to the module.

This allows us to cache partially instantiated types in `instantiatedTypes`, i.e. `Foo[U]` where `U` is a generic parameter of the current context now only substitutes the body of `Foo` with `U` once rather than every time `Foo[U]` is invoked. Maybe these shouldn't generate syms either but idk if it would be worth it to try to change that.

This would also allow us to repeatedly alternate over iterating the type instantiation requests and the proc instantiation requests so if one adds to the other they are all considered (the intended behavior of `instantiateGenerics`). But this is not done for now to not disturb the current behavior, it will be needed if instantiating generic hooks creates new type instantiations. `typeRequests` is still removed as it was not used.

`instantiatedTypes` also no longer needs to be ordered for this reason, `instantiatedProcs` already did not need to be ordered as it was never iterated. But could be changed back.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -2560,7 +2560,7 @@ proc semInvoke(c: var SemContext; n: var Cursor) =
       let targetSym = newSymId(c, headId)
       c.instantiatedTypes[key] = targetSym
       if genericArgs == 0:
-        c.typeInstantiations.add targetSym
+        c.typeInstDecls.add targetSym
       var sub = createTokenBuf(30)
       subsGenericTypeFromArgs c, sub, info, headId, targetSym, decl, args
       c.dest.endRead()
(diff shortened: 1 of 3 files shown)


Araq:
In the Nim compiler I've been bitten occassionally by the non-deterministic iteration order of Table and so I prefer OrderedTable just to be on the safe side and the new runtime will only offer OrderedTable anyway (comparable to Python's new dict impl). Just so that you know, this PR is still perfectly fine.

PR #586, opened 2025-02-20
hexer: for loop parameter instantiation

ringabout:
```nim
iterator countup3(a: int): int =
  yield a

iterator powers3(a: int): int =
  for j in countup3(a):
    yield j

proc printf(format: cstring) {.importc: "printf", varargs, header: "<stdio.h>", nodecl.}

for i in powers3(5):
  for j in countup3(4):
    printf("Hello, world: %ld\n", i+j)
```

With this PR, param `a` is instantiated with a fresh name each time for each `for loop` so that it won't conflict with nested for loops.


Araq:
But `inlineLoopBody` already takes a `mapping`. This solution seems way more complicated than necessary.

ringabout:
I'm trying to think of a better way. Perhaps I can merge two cases, though I failed to do that for some issues. I will try to figure out some other solutions

ringabout:
Two mappings are for different purposes. 

`yield` would inline loop body. It needs to replace symbols in it. The mapping is created for each `yield`.

Parameter mapping is used for iterator body. The mapping is created for the iterator, which is needed before `inlineLoopBody`.

src/hexer/iterinliner.nim
```diff
@@ -392,9 +423,7 @@ proc transformStmt(e: var EContext; c: var Cursor) =
       inc c
       while c.kind notin {EofToken, ParRi}:
         transformStmt(e, c)
-      takeParRi e, c # skipParRi
-    of VarS, LetS, CursorS, ResultS:
```
Araq, line 396: Still entirely correct comment. Handle locals correctly please.

ringabout: I removed them to delegate them to else branch so that each element will be transformed properly. Same as https://github.com/nim-lang/nimony/pull/630

PR #592, opened 2025-02-21
instantiate field types of invocation object types in exprs

### LINKED ISSUE #543: proper types for fields of invocation object types

metagn:
fixes #543

The aim with #580 was to instantiate these fully in `semInvoke` then look up the impls from `instantiatedTypes` with the canon type of the invoke types, but we can't do this yet because of the problems with instantiating forward declared types.

Instead `findObjField` and `buildDefaultObjConstr` now accept/build bindings based on invocation type args and instantiate the final field type with them. They rebuild bindings for parent types as well when iterating over them, substituting their arguments with the existing bindings.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1736,17 +1736,41 @@ proc genericRootSym(td: TypeDecl): SymId =
     assert root.kind == Symbol
     result = root.symId
 
-proc buildInvokeArgs(decl: TypeDecl, bindings: var Table[SymId, Cursor], invokeArgs: Cursor) =
-  var typevar = decl.typevars
-  assert typevar.substructureKind == TypevarsU
-  inc typevar
-  var arg = invokeArgs
-  while arg.kind != ParRi:
-    let tv = asLocal(typevar)
-    assert tv.kind == TypevarY
-    bindings[tv.name.symId] = arg
-    skip typevar
-    skip arg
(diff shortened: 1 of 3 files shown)


src/nimony/sem.nim
```diff
@@ -1745,27 +1777,30 @@ proc findObjFieldAux(t: Cursor; name: StrId; level = 0): ObjField =
...
+      var newBindingBuf = default(TokenBuf)
+      var newBindings = initTable[SymId, Cursor]()
+      if baseInvokeArgs != default(Cursor):
+        newBindingBuf = createTokenBuf(16)
+        var sc = SubsContext(params: addr bindings)
+        while baseInvokeArgs.kind != ParRi:
+          subs(c, newBindingBuf, sc, baseInvokeArgs)
+          skip baseInvokeArgs
+        newBindingBuf.addParRi()
+        buildInvokeArgs(decl, newBindings, beginRead(newBindingBuf))
```
Araq, line 1792: Move to a helper proc and while you're at it add a long comment explaining why all these shenigans are necessary.

metagn: Hopefully better now

Araq:
Fine for now but we need to rethink object fields and the complexity they cause for generic instantiations. Maybe if we keep them as identifier-like that get the type always via their parent structure?

PR #595, opened 2025-02-21
fixes #588; NIFC generates &buf.a[0] for arrays

### LINKED ISSUE #588: addReadLine proc in syncio module causes incompatible pointer type error from GCC
In syncio module, `fgets` is declared as:
```nim
proc fgets(str: out array[bufsize, char]; n: int32; f: File): cstring {.
  importc: "fgets", header: "<stdio.h>".}
```

And `addReadLine` proc passes `array[bufsize, char]` type variable to `fgets`:
```
  var buf: array[bufsize, char]
  while fgets(buf, bufsize.int32, f) != nil:
```

This code generates following C code:
```c
  AarrayAcS8ZS80_0_t buf_3;
  while ((!(((void*)fgets((&buf_3), ((NI32)bufsize_0_syn1lfpjv), f_22)) == NIM_NIL))){
```

This code result in compile error with GCC:
```console
nifcache/syn1lfpjv.c:446:28: error: passing argument 1 of ‘fgets’ from incompatible pointer type [-Wincompatible-pointer-types]
  446 |   while ((!(((void*)fgets((&buf_3), ((NI32)bufsize_0_syn1lfpjv), f_22)) == NIM_NIL))){
      |                           ~^~~~~~~
      |                            |
      |                            AarrayAcS8ZS80_0_t *
```

`fgets`'s first parameter type is `char *__restrict __`.

demotomohiro:
Fixes `getType` proc in `nifc/typenav.nim` to return the type of given variable correctly.
Adds `[0]` when generating the address of array.

### DIFF
--- a/src/nifc/genexprs.nim
+++ b/src/nifc/genexprs.nim
@@ -204,6 +204,9 @@ proc genAddr(c: var GeneratedCode; n: var Cursor) =
   genx c, n
   if arrType.typeKind == ArrayT and not c.m.isImportC(arrType):
     c.add ".a"
+    c.add BracketLe
+    c.add "0"
+    c.add BracketRi
   c.add ParRi
   if n.kind != ParRi and n.typeQual == CppRefQ:
     if c.m.config.backend == backendCpp:
--- a/src/nifc/typenav.nim
+++ b/src/nifc/typenav.nim
@@ -28,12 +28,16 @@ proc getType*(m: var Module; n: Cursor): Cursor =
     result = createIntegralType(m, "(err)")
   of Symbol:
     var it {.cursor.} = m.current
+    var res = default Cursor
     while it != nil:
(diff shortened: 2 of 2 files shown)


src/nifc/genexprs.nim
```diff
@@ -204,6 +204,9 @@ proc genAddr(c: var GeneratedCode; n: var Cursor) =
   genx c, n
   if arrType.typeKind == ArrayT and not c.m.isImportC(arrType):
     c.add ".a"
+    c.add BracketLe
+    c.add "0"
+    c.add BracketRi
```
Araq, line 209: What you need to do here is to detect if you're within a function call and that function is imported from C and then deduce from this fact that the address operator needs to be special cased.

In other words, inside `genCall` add special logic for `AddrC`.

PR #602, opened 2025-02-23
match directly to inferred value of generic param

metagn:
The current `sigmatch.commonType` just accepts any type for the rematch which can cause invalid matches. The rematch can be a conversion match as well which is not trivial to handle in `commonType`. So this is changed to just recurse over `singleArgImpl` same as how `linearMatch` recurses over itself. This does not give the `InvalidRematch` error message though, instead giving a normal match error.

### DIFF
--- a/tests/nimony/nosystem/tfib2.nif
+++ b/tests/nimony/nosystem/tfib2.nif
@@ -104,8 +104,7 @@
     (else 2,1
      (stmts 7
       (asgn ~7 result.0 11
-       (add ~6,~21
-        (i -1) ~6
+       (infix \2B ~6
         (call ~3 fib.0.tfitjdpcx 2
          (infix \2D ~1 a.1 1 +1)) 5
         (call ~3 fib.0.tfitjdpcx 2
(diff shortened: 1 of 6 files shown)


tests/nimony/nosystem/tfib2.nif
```diff
@@ -104,8 +104,7 @@
     (else 2,1
      (stmts 7
       (asgn ~7 result.0 11
-       (add ~6,~21
-        (i -1) ~6
+       (infix \2B ~6
```
Araq, line 107: Looks wrong to me.

metagn: Thought this at first too but this is actually the `+` in `fib(a-1) + fib(a-2)`, which should match to the concept proc which converts to the identifier. What I think happened before is: `-` in `a-1` and `a-2` matched to the concept proc and inferred `Self` to `int` and so the return type instantiated to `int`, then the `+` matched to the `int` overload. Now `Self` infers to `T` and stays that way.

PR #605, opened 2025-02-23
make compat untyped templates/generics opt in via pragma for now

metagn:
Unimplemented stuff etc can break these tests when new templates/generic procs are added to system, we can hold off on maintaining them until the feature is finished.


Araq:
Was about to suggest the same...

PR #608, opened 2025-02-23
fix some line infos, unused variables

metagn:


Araq:
Does this mean `renderer.nim` can activate `when false: #if m0.isValid:`? Would be nice if the compiler doesn't dump NIF anymore.

metagn:
The line infos given to renderer are all correct but the output of filelinecache does not match them

metagn:
For maximum ease of implementation we could maybe output something like:

```nim
tests/nimony/lookups/tambtype.nim(3, 8) Error:
var x: Foo = 123
         ^ ambiguous identifier
```

Araq:
> For maximum ease of implementation we could maybe output something like: ...

Opt-in, maybe but in general I prefer error messages that are on a single line, esp. when more than one error is produced. Also keeps it more managable for nimsuggest/editor integration. But for these maybe we should even go all-in and only produce NIF.

PR #614, opened 2025-02-25
Add newString proc

demotomohiro:
`lib/std/strutils.nim` uses `newString` and `lib/std/parseopt.nim` import it.

### DIFF
--- a/lib/std/system/stringimpl.nim
+++ b/lib/std/system/stringimpl.nim
@@ -231,3 +231,6 @@ proc prepareMutation*(s: var string) =
       oomHandler len
       s.i = EmptyI
     s.a = a # also do this for `a == nil`
+
+proc newString*(len: int): string =
+  result.setLen len
--- a/lib/std/system/stringimpl.nim
+++ b/lib/std/system/stringimpl.nim
@@ -238,4 +238,4 @@ proc newString*(len: int): string =
     result = string(a: a, i: len shl LenShift)
   else:
     oomHandler len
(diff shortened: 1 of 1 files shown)


lib/std/system/stringimpl.nim
```diff
@@ -231,3 +231,6 @@ proc prepareMutation*(s: var string) =
       oomHandler len
       s.i = EmptyI
     s.a = a # also do this for `a == nil`
+
+proc newString*(len: int): string =
+  result.setLen len
```
Araq, line 236:
```suggestion
  let a = cast[StrData](alloc(len))
  if a != nil:
    result = string(a: a, i: len shl LenShift)
  else:
    oomHandler len
    result = string(a: nil, i, EmptyI)
```

Araq: Would be the correct implementation, as far as I can tell.

lib/std/system/stringimpl.nim
```diff
@@ -231,3 +231,11 @@ proc prepareMutation*(s: var string) =
...
       s.i = EmptyI
     s.a = a # also do this for `a == nil`
+
+proc newString*(len: int): string =
+  let a = cast[StrData](alloc(len))
+  if a != nil:
+    result = string(a: a, i: len shl LenShift)
+  else:
+    oomHandler len
+    result = string(a: nil, i, EmptyI)
```
Araq, line 241:
```suggestion
    result = string(a: nil, i: EmptyI)
```

PR #622, opened 2025-02-26
enable some tests, fix sigmatch for proc types

metagn:
The test in `tlocalcall` is enabled but with an explicit `proc ()` annotation as otherwise its type gets inferred to `ParamsT` without the return type, so in the AST it becomes `(let foo . . (params ...) bar)`, so the return type of `foo` is skipped from the `params` to `bar` and so the type of `foo(123)` becomes literally `bar`. In any case the `proc ()` annotation has worked in other places for a while now.

A small fix is needed in sigmatch for proc types with empty pragmas to work: `skipToEnd` is called before `inc` which is supposed to skip the `ParRi`, but `skipToEnd` already skips the `ParRi`, so a new `skipUntilEnd` is added which doesn't skip it. Also `nkEmpty` params and pragmas for proc types in nifler are now allowed to generate `.`.

The disabled `when defined` tests in `twhen` are enabled and moved to a different test that imports system, just to not have to redefine `and`/`or` but it could be moved back to ensure it works without system.

The string index test in `tindex` is enabled as strings are now implemented. Although this is in a `main` proc that does not run as it does not allocate the string, it only tests that it compiles.

### DIFF
--- a/src/lib/nifcursors.nim
+++ b/src/lib/nifcursors.nim
@@ -80,11 +80,24 @@ proc skip*(c: var Cursor) =
   inc c
 
 proc skipToEnd*(c: var Cursor) =
+  ## skips `c` until an unmatched ParRi is found, then skips the ParRi and returns
   var nested = 0
   while true:
     if c.kind == ParRi:
       if nested == 0:
         inc c
+        break
+      dec nested
+    elif c.kind == ParLe:
+      inc nested
+    inc c
+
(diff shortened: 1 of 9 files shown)


Araq:
> Also nkEmpty params and pragmas for proc types in nifler are now allowed to generate ..

Wait, isn't this even worse? Empty pragmas should just be `(pragmas)` IMO, these dots are super annoying to work with.

PR #632, opened 2025-02-27
transform tuple unpacking in sem, implement underscores

### LINKED ISSUE #628: implement tuple unpacking
Rewrite it after semantic checking so that all the other systems don't have deal with it.

metagn:
closes #628

Also supports nesting out of the gate, both in nifler and sem, mostly to make sure we don't need a redesign to implement it when we need it.


Araq:
While you're at it ensure underscores are valid and "redefinable", `var _ = 4; var _ = "abcv"` is valid

metagn:
Done, underscores are checked directly in `symtabs` so any calls to `addOverloadable`/`addNonOverloadable` consider them but we could add `add(Non)OverloadableConsiderUnderscore` overloads that do this as well

Issue #641, opened 2025-02-27
rename nifcache to nimcache


Araq:
- Also make it configurable via a command line option. (It might be necessary to pass this around between the different subsystems: nimony, nimsem, hexer.)

Rationale: git templates already understand that `nimcache` is to be ignored.

PR #643, opened 2025-02-27
implements #641; add nimcache commmand line option and change default…

demotomohiro:
… cache directory

### DIFF
--- a/src/nifler/configcmd.nim
+++ b/src/nifler/configcmd.nim
@@ -203,8 +203,14 @@ proc produceConfig*(infile, outfile: string) =
     b.addHeader()
     b.withTree "config":
       b.buildConfig conf
-    conf.outDir = AbsoluteDir getCurrentDir()
-    conf.outFile = RelativeFile(outfile)
+    if outfile.isAbsolute:
+      let p = outfile.splitPath
+      conf.outDir = AbsoluteDir(p.head)
+      conf.outFile = RelativeFile(p.tail)
+    else:
+      conf.outDir = AbsoluteDir getCurrentDir()
+      conf.outFile = RelativeFile(outfile)
+
     conf.setNote hintSuccessX
     genSuccessX(conf)
   finally:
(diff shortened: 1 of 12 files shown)


src/nifler/configcmd.nim
```diff
@@ -203,8 +203,14 @@ proc produceConfig*(infile, outfile: string) =
...
-    conf.outDir = AbsoluteDir getCurrentDir()
-    conf.outFile = RelativeFile(outfile)
+    if outfile.isAbsolute:
+      let p = outfile.splitPath
+      conf.outDir = AbsoluteDir(p.head)
+      conf.outFile = RelativeFile(p.tail)
+    else:
+      conf.outDir = AbsoluteDir getCurrentDir()
+      conf.outFile = RelativeFile(outfile)
+
```
Araq, line 213: Why was this change required?

mrgaturus:
bonus: to allow nimony compile outside compiler folder you need add a `config.paths.add compilerDir()` in `nimsem.nim` and `nimony.nim` to ensure the hardcoded system module mangle is still correct

src/nifler/configcmd.nim
demotomohiro: `outfile` can be an absolute path.
From `genSuccessX(conf)` proc, `absOutFile` proc in `https://github.com/nim-lang/Nim/blob/devel/compiler/options.nim#L723` is called.
`absOutFile` proc do `result = conf.outDir / conf.outFile`.
So `conf.outFile` must be relative.

demotomohiro:
In this PR, most of paths in makefile become absolute path.
Should I remove `toAbsolutePath` when assigning to `config.nifcachePath` in `src/nimony/nimony.nim` so that paths in makefile are relative?

Araq:
> Should I remove toAbsolutePath when assigning to config.nifcachePath in src/nimony/nimony.nim so that paths in makefile are relative?

Yes, exactly.

PR #651, opened 2025-02-28
WIP: openArray implementation

Araq:


lib/std/system/openarrays.nim
```diff
@@ -0,0 +1,48 @@
+
+type
+  openArray*[T] {.view.} = object
+    a: ptr UncheckedArray[T]
+    len: int
+
+proc `[]`*[T](x: openArray[T]; idx: int): var T {.inline, requires: idx >= 0 and idx < x.len.} = x.a[idx]
```
blackmius, line 7: what `requires` pragma do? is it checked in runtime?

PR #673, opened 2025-03-03
Test ubuntu 24.04 + gcc 14

ringabout:
Tests failed locally on my Ubuntu 24.04


ringabout:
merged into https://github.com/nim-lang/nimony/pull/673

PR #677, opened 2025-03-03
make linearMatch aware of special cases

### LINKED ISSUE #646: `linearMatch` does not handle invocations matching instantiated symbols
```nim
type Foo[T] = object

proc foo[T](x: ptr Foo[T]) = echo "b"

var x = Foo[int]()
foo(addr x)
```

```
(9, 4) Error: Type mismatch at [position]
[1] expected: (ptr
 (at Foo.0.tsp9tki5d1 T.1.tsp9tki5d1)) but got: (ptr Foo.1.tsp9tki5d1)
```

Leaving unassigned in case someone has a better idea but I'm guessing it would be enough to just reuse the same code in `singleArg` for `InvokeT`

metagn:
closes #646

This makes `linearMatch` more a way of matching types invariantly to each other rather than just a check for tree equality that also considers typevars. Maybe `linearMatch` is not the best implementation for this but it's the expected behavior from `linearMatch` in the places where it's used.

For integral types, bits are now normalized, so `ptr int` now matches `ptr int64`. Edit: This is now disabled for `typedesc`, i.e. `typedesc[int]` and `typedesc[int64]` are considered different. In original Nim they match each other, they just match the equal versions better, but I remember this not being intended.

For procs, `procTypeMatch` is called directly since it currently basically checks for equality, but in the future if stuff like convertibility from `nimcall` to `closure` is implemented then this will have to be disallowed when it's called from `linearMatch`.

The behavior of `InvokeT` matching a symbol type is now moved to `linearMatch` (`InvokeT` matching `InvokeT` already calls linearMatch) but `singleArg` will probably need its own implementation in the future too so that generic inheritance can be implemented.

A problem with mixing other procs and `linearMatch` is that it stops iterating on an error, when procs expect it to fully skip the types in some cases; and when `linearMatch` calls other procs or recurses over itself, the only way it knows to stop iterating is by checking for an error, which may have been produced previously and not necessarily in the call to the proc. An option for this is to make `linearMatch` indicate if it stopped iterating due to an error via a flag, and add another version like `linearMatchAlwaysSkip` that skips `f` and `a` fully by checking for this flag (i.e. `if error: f = fOrig; skip f`). Could do this here or in a followup, or maybe there is a better idea - Edit: Always fully skips now

### DIFF
--- a/lib/std/system.nim
+++ b/lib/std/system.nim
@@ -409,8 +409,8 @@ template `>`*(x, y: untyped): untyped =
 
 template default*(x: typedesc[bool]): bool = false
 template default*(x: typedesc[char]): char = '\0'
-template default*(x: typedesc[int]): int = 0
-template default*(x: typedesc[uint]): uint = 0'u
+#template default*(x: typedesc[int]): int = 0
+#template default*(x: typedesc[uint]): uint = 0'u
 template default*(x: typedesc[int8]): int8 = 0'i8
 template default*(x: typedesc[uint8]): uint8 = 0'u8
 template default*(x: typedesc[int16]): int16 = 0'i16
--- a/lib/std/system.nim
+++ b/lib/std/system.nim
@@ -409,8 +409,8 @@ template `>`*(x, y: untyped): untyped =
 
 template default*(x: typedesc[bool]): bool = false
 template default*(x: typedesc[char]): char = '\0'
-#template default*(x: typedesc[int]): int = 0
-#template default*(x: typedesc[uint]): uint = 0'u
+template default*(x: typedesc[int]): int = 0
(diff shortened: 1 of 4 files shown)


Araq:
> Could do this here or in a followup, or maybe there is a better idea

why not make it always process the entire token stream.

metagn:
Since we are iterating over `f` and `a` at the same time their token streams could be completely different after erroring. I guess `nested` is supposed to be equal as long as they don't error so we could do something like `while nested != 0: skipToEnd f; skipToEnd a; dec nested`. `f = fOrig; skip f` etc will always work but might not be as efficient, but maybe the repeated calls to `skipToEnd` is slower too.

PR #678, opened 2025-03-03
handle string/cstring case in sem, test error messages

metagn:
refs #668

Doesn't work in hexer yet, `getType` gives `auto` for the type of `a` in tstringcase.

For the overlap checks, the ID of the string literal is used.

`cstring("abc")` is now supported in expreval.

`evalConstIntExpr` and the new `evalConstStrExpr` now use the output AST from expreval to properly propagate evaluation errors, previously they would just return NaN without producing an error in `c.dest`. We could replace it with a simple `if nan: error "expected integer"` but using the output from expreval gives more insight to what couldn't be evaluated.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -690,14 +690,14 @@ proc semConstIntExpr(c: var SemContext; n: var Cursor) =
     c.dest.shrink start
     c.dest.add valueBuf
 
-proc semConstExpr(c: var SemContext; it: var Item) =
+proc semConstExpr(c: var SemContext; it: var Item): TokenBuf =
   let start = c.dest.len
   semExpr c, it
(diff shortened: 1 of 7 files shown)


src/nimony/sem.nim
```diff
@@ -3822,12 +3849,19 @@ proc semWhen(c: var SemContext; it: var Item) =
     producesVoid c, info, it.typ
 
 proc semCaseOfValue(c: var SemContext; it: var Item; selectorType: TypeCursor;
-                    seen: var seq[(xint, xint)]) =
+                    seen: var seq[(xint, xint)]; isString: bool) =
```
Araq, line 3852: Instead duplicate this proc so that it takes `seen: var HashSet[string/StrId]` as it needs to check for duplicate string literals too.

PR #697, opened 2025-03-05
nimony: fix missing tools resolve when compiling outside compiler folder

mrgaturus:
a quick follow up of #643 to allow nimony compile outside compiler folder

### DIFF
--- a/src/nimony/nimony.nim
+++ b/src/nimony/nimony.nim
@@ -156,8 +156,8 @@ proc handleCmdLine() =
     quit "command missing"
   of SingleModule:
     if not isChild:
-      createDir(binDir())
       createDir(config.nifcachePath)
+      createDir(binDir())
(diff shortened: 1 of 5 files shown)


src/nimony/nimony.nim
```diff
@@ -157,18 +158,25 @@ proc handleCmdLine() =
     if not isChild:
       createDir(config.nifcachePath)
       createDir(binDir())
+      # configure required tools
+      setCurrentDir(compilerDir())
```
Araq, line 162: I don't like this solution at all. Isn't there a better way?

mrgaturus: i think i can tweak the `requiresTool` to find the src folder properly like it does with binaries, and add something in semos to execute `exec "git submodule update --init"` always in compiler path

PR #710, opened 2025-03-06
implement empty seq

### LINKED ISSUE #470: make empty seq literals work

metagn:
closes #470

Converters from `[]`/`@[]` to a concrete type still do not work for openarray but empty openarrays might not be that common, and we could special case openarray directly instead of making converters in general work


src/nimony/sigmatch.nim
```diff
@@ -662,6 +662,27 @@ proc isStringType*(a: Cursor): bool {.inline.} =
 proc isSomeStringType*(a: Cursor): bool {.inline.} =
   result = a.typeKind == CstringT or isStringType(a)
 
+proc isSomeSeqType*(a: Cursor, elemType: var Cursor): bool =
+  # check that `a` is either an instantiation of seq or an invocation to it
+  result = false
+  var a = a
+  if a.kind == Symbol:
```
Araq, line 669: I think you need a while loop here plus a counter due to the possibility of: 

```nim
type
  MyAlias = OtherAlias
  OtherAlias = StupidAlias
  StupidAlias = seq
```

src/nimony/sigmatch.nim
```diff
@@ -807,6 +828,30 @@ proc isEmptyLiteral*(n: Cursor): bool =
     skip n # type
     result = n.kind == ParRi
 
+proc isEmptyCall*(n: Cursor): bool =
+  if n.exprKind notin CallKinds:
+    return false
+  var n = n
+  inc n
+  # overload of `@` with empty array param:
+  result = n.kind == Symbol and pool.syms[n.symId] == "@.1." & SystemModuleSuffix
```
Araq, line 837: What if the `@` was overloaded? Still correct here then as it's done after overload resolution?

src/nimony/sigmatch.nim
metagn: Aliases are supposed to be inlined, as in `MyAlias` would immediately resolve to `seq` when semchecked, including with generics, but could still add it for safety

src/nimony/sigmatch.nim
metagn: Yes, this requires the argument to be semchecked already in a context that allows empty containers (can clarify this in a comment), which should disambiguate to the overload that takes an empty array. It would only break if `@` is overriden with an overload that matches an empty array better than the system one. If we remove the empty overload we could also check for the first overload of `@`, we could also just check for identifier equality to `@`.

Araq: No, it's fine this way then.

PR #713, opened 2025-03-06
adds a test for valgrind leaking verification

ringabout:


mrgaturus:
the test could fail if valgrind version is different to the hardcoded in the test output (`Valgrind-3.22.0`)

ringabout:
first three lines are skipped now in https://github.com/nim-lang/nimony/pull/881

PR #725, opened 2025-03-09
special symbol scheme for type instantiations

### LINKED ISSUE #715: Calling the generic procedure in imported module that construct object causes type mismatch compile error
Calling the generic procedure in imported module that construct object causes type mismatch compile error

mytests/test2.nim:
```nim
type
  Foo[T] = object
  Bar* = object
    y: Foo[int]

proc foo[T](): Foo[T] = discard

proc initBar*[T]() =
  discard Bar(y: foo[int]())
```

mytests/test.nim:
```nim
import test2

initBar[int]()
```

```console
bin/nimsem  m nimcache/tes7tvw8r1.1.nif nimcache/tes7tvw8r1.2.nif nimcache/tes7tvw8r1.2.idx.nif
bin/nimsem  m nimcache/tes9hvujc.1.nif nimcache/tes9hvujc.2.nif nimcache/tes9hvujc.2.idx.nif
src/nimony/nimsem.nim(119) nimsem
src/nimony/nimsem.nim(116) handleCmdLine
src/nimony/nimsem.nim(51) singleModule
src/nimony/sem.nim(6066) semcheck
src/nimony/sem.nim(575) instantiateGenerics
src/nimony/sem.nim(567) instantiateGenericProc
src/nimony/nimony_model.nim(3483) semProc
src/nimony/sem.nim(727) semProcBody
src/nimony/sem.nim(710) semStmtsExprImpl
src/nimony/sem.nim(5798) semExpr
src/nimony/sem.nim(3756) semDiscard
src/nimony/sem.nim(5905) semExpr
src/nimony/sem.nim(4743) semObjConstr
src/nimony/sem.nim(5872) semExpr
src/nimony/sem.nim(1715) semCall
src/lib/nifcursors.nim(1515) resolveOverloads
src/nimony/sem.nim(1018) typeofCallIs
src/nimony/sem.nim(196) commonType
src/nimony/sem.nim(28) typeMismatch
src/nimony/sembasics.nim(177) buildErr
mytests/test2.nim(9, 26) Error: type mismatch: got: Foo.0.tes9hvujc but wanted: Foo.1.tes7tvw8r1
type mismatch: got: Foo.0.tes9hvujc but wanted: Foo.1.tes7tvw8r1
```

Calling `initBar[int]()` in mytests/test2.nim and compiling it doesn't generate errors.

I rewrote `proc initTable` in `lib/std/tables.nim` to use `newSeq` proc as `@[]` didn't work.
I imported it and called `initTable` proc but got compile error from `initTable` proc.
I think above code are minial code that reproduce the same bug.

metagn:
fixes #715, supersedes #724

The scheme is:

```
abc.123.origmod.Iabcdefgh.instmod
```

where `abcdefgh` is the encoded hash of the invocation NIF string (`typeToCanon` depends on runtime literal IDs). The first 3 parts are the same as the original symbol. But the original symbol being preserved isn't necessary, it can be included in the hash, so we could have something as bare as:

```
abc.Iabcdefgh.instmod
```

or any other arrangement in between if the current one is too ugly.

The final `.instmod` part is ignored both in `sigmatch` for comparisons, and removed in the NIFC output by hexer, so that the type is generated under the same name. The scheme also applies to any symbol declared inside instantiations so that fields are the same across instantiations too.

I also tried including proc instantiations but it broke for inline procs and isn't really necessary to make procs work.

### DIFF
--- a/src/gear2/modnames.nim
+++ b/src/gear2/modnames.nim
@@ -41,6 +41,14 @@ const
 const
   Base36 = "0123456789abcdefghijklmnopqrstuvwxyz"
 
+proc uhashBase36*(s: string): string =
+  var id = uhash(s)
+  result = newStringOfCap(8)
+  # Convert decimal number to base 36, reversed since it does not matter:
+  while id > 0'u32:
+    result.add Base36[int(id mod 36'u32)]
+    id = id div 36'u32
+
 proc moduleSuffix*(path: string; searchPaths: openArray[string]): string =
   var f = relativePath(path, getCurrentDir(), '/')
   # Select the path that is shortest relative to the searchPath:
--- a/src/hexer/nifcgen.nim
+++ b/src/hexer/nifcgen.nim
@@ -1520,6 +1520,7 @@ proc writeOutput(c: var EContext, rootInfo: PackedLineInfo) =
       else:
         let s = pool.syms[n.symId]
         if isInstantiation(s):
+          # ensure instantiations have the same name across modules:
           b.addSymbol(removeModule(s))
         else:
           b.addSymbol(s)
@@ -1530,6 +1531,7 @@ proc writeOutput(c: var EContext, rootInfo: PackedLineInfo) =
       else:
         let s = pool.syms[n.symId]
         if isInstantiation(s):
+          # ensure instantiations have the same name across modules:
           b.addSymbolDef(removeModule(s))
         else:
           b.addSymbolDef(s)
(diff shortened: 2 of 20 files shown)


Araq:
> so we could have something as bare as: `abc.Iabcdefgh.instmod`

That's the same scheme that I came up with (didn't tell you about it though) so yeah, let's do it this way.

metagn:
`extractBasename` checks for the `.123` part after the basename and there is the possibility that there are multiple symbols with the same basename in the instantiation, so I made it `abc.123.Iabcdefgh.instmod` for now. We can still remove it but I didn't consider it would need more changes.

PR #733, opened 2025-03-10
implement mutual generic disambiguation

### LINKED ISSUE #647: generic disambiguation not implemented
```nim
import std/syncio

type Foo[T] = object

proc foo[T](x: T) = echo "a"
proc foo[T](x: Foo[T]) = echo "b"

proc bar[T](x: Foo[T]) =
  foo(x)
```

(Matching generic `Foo[T]` is for independence from #646)

```
(9, 6) Error: ambiguous call: 'foo'
```

Not sure how much to copy old nim here, either implementing both of `checkGeneric` (complex) and `complexDisambiguation` (simple) in order as per the spec, implementing just one for the time being, or coming up with a new method

metagn:
closes #647

The params of the callees are matched against each other, to check if they match either as a generic match and the other one doesn't. I don't exactly know why only generic matches are counted in the original compiler and not also equal matches as in the case of `var T` matching `T` (though the vice versa also matches for now so this wouldn't disambiguate it yet). To get `var T` disambiguation we also need the specificity counter method but that can be done in a followup.

### DIFF
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -1107,6 +1107,36 @@ type
     FirstWins,
     SecondWins
 
+proc mutualGenericMatch(a, b: Match): DisambiguationResult =
+  # same goal as `checkGeneric` in old compiler
+  result = NobodyWins
+  let c = a.context
+  var aParams = a.fn.typ
+  var bParams = b.fn.typ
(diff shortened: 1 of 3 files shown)


Araq:
There should be a prize for PRs like these... :D

PR #743, opened 2025-03-11
experiments with a NIF compression algorithm

Araq:


Araq:
Better done with a filesystem abstraction that lets us compress across NIF files.

PR #752, opened 2025-03-12
added system.cmp

Araq:


mrgaturus:
a branchless version could be:
```nim
proc cmp*[T: Comparable](x, y: T): int =
  int(x > y) - int(x < y)
```

Araq:
CPUs are superb with branch prediction these days so your way is probably slower. That said, your impl doesn't require a `==` operator, very interesting.

PR #758, opened 2025-03-12
handle re-exports by turning them into imports

### LINKED ISSUE #747: implement re-exports

metagn:
closes #747, follows up #791 and #794 (were split from this PR)

Importing a module with re-exports builds a list of imports of the forwarded modules, which imports the forwarded symbols when applied. For filtered imports, the export/import filters are combined, which is equivalent to computing set intersections and so is commutative and doesn't lose information, so we do not have to keep track of every filter and can just merge them as we encounter them.

The filters are also changed to use `HashSet` because `PackedSet` is 320 bytes and not really appropriate for this use case, where we sometimes need to copy them etc.


metagn:
Using this PR for the final split

src/nimony/programs.nim
```diff
@@ -85,6 +108,19 @@ proc loadInterface*(suffix: string; iface: var Iface;
...
+      case kind
+      of ExportIdx: ImportAll
+      of FromexportIdx: FromImport
+      of ExportexceptIdx: ImportExcept
+      else: ImportAll
+    var exportFilter = ImportFilter(kind: filterKind)
+    for s in names:
+      exportFilter.list.incl(s)
+    mergeFilter(exportFilter, filter)
+    exports.add (path, move exportFilter)
```
Araq, line 123:
```suggestion
    exports.add (path, ensureMove exportFilter)
```

PR #771, opened 2025-03-13
hexer: fixes const strings

ringabout:
- `destroy` should not be applied to consts
- string literals are not simple: we need to add a `conv` for string literals of `string` types, and we need to restrict the inlining for string literals of `cstring` types


Araq:
> string literals are not simple: we need to add a conv for string literals of string types, and we need to restrict the inlining for string literals of cstring types

I don't agree, we need to revisit this decision.

ringabout:
Yeah, I should fix calls with string literals instead

PR #775, opened 2025-03-13
partially fixes 716; changing command line options changes module suffix

demotomohiro:
When command line options that can affects generated nif or c files are changed, new generated files have different module suffix and existing cached files are ignored.
But module suffix of the system module are not changed because it must be constant equals to `nimony/builtintypes.SystemModuleSuffix`.

This PR doesn't support changing `--passC` and `--passL` options.
I will create separate PR to support these options.


Araq:
No, convoluted and hard to debug. Instead generate a file with the options inside (only overwrite it when it changed) and tell the makefile that the sem'checked results depend on it.

PR #776, opened 2025-03-14
LLVM backend: WIP

Araq:


mrgaturus:
it's possible use some of libclang to allow llvm backend use c header files to make {.header.} pragma working with llvm backend, Zig uses it for it's `@cImport`, `@cDefine` and `@cInclude`

Araq:
No, I'm not interested in this "we need a C compiler for good" business.

Araq:
Instead you can decide for every module whether to compile the module to C or to LLVM. You then get not good debug info for your POSIX wrappers but then who needs that. This design keeps the overall complexity low as no new language pragmas needs to be introduced.

PR #782, opened 2025-03-14
nifcgen: fixes string suffix and adds StringLit back

ringabout:
follow up https://github.com/nim-lang/nimony/pull/771


ringabout:
It seems to need https://github.com/nim-lang/nimony/pull/781

PR #783, opened 2025-03-14
fix consecutive imports to the same module

metagn:
Previously the first import of a module would be the only one considered. Now they are additive. The idea is that a module being imported and another module exporting symbols from it do not clash with each other but I'm not sure if this fix is needed for that. In any case silently doing nothing on a consecutive import is not good.

### DIFF
--- a/src/nimony/programs.nim
+++ b/src/nimony/programs.nim
@@ -4,7 +4,7 @@
 # See the file "license.txt", included in this
 # distribution, for details about the copyright.
 
-import std / [syncio, os, tables, times, packedsets]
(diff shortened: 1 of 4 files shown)


Araq:
I don't understand this. What else should it do besides ignoring the second import?

metagn:
As in the test, the second import might import symbols that the first import didn't (as in a `from import` first then a full `import`). So we fill the import table from the module symbols again this time with the new filter, but we don't have to load the module again. Since nothing is removed from the import table, this just additively combines all the import statements. Also applies for consecutive `from import`s, i.e. `from m import a; from m import b`. It now acts like `from m import a, b` rather than just `from m import a`.

PR #791, opened 2025-03-15
implement sem for re-exports

metagn:
split from #758, 1/3

Exports in a module build a table from modules to the filter of symbols exported from the modules; either the whole module, symbols with specific names, or symbols without specific names. The symbols themselves aren't saved, only their names, importing only considers the names so storing the symbols would be redundant.

The symbols inside `export` statements are semchecked before adding them to the filter based on their module (or exporting the whole module if the symbol is a module); so this also works for qualified symbols, it will only export the symbols with that name from the module it was qualified from.

The built exported symbols are not added to the index in this PR in case there are necessary changes here, then it can be split from #758.


Araq:
Fine but sem.nim is getting unwidely. Use an `include semimport` and move some logic to semimport.

metagn:
Moved all the procs under "include/import handling" to semimport

PR #794, opened 2025-03-16
add re-exports to index

metagn:
split from #758, 2/3

The exports gathered in #791 are now output in the index. Each module that symbols are exported from gets its own export tag, with the format: `(export "modpath")` if the entire module was exported, `(exportexcept "modpath" a b c)` if all symbols except some were exported, and `(fromexport "modpath" a b c)` (new tag unique to indexes) if only some symbols from a module were exported, as in `export foo.a, foo.b` etc. These correpond with the import required to access the symbols from the forwarded module, but another format is also fine. The module path used is an absolute path,


src/nimony/sem.nim
```diff
@@ -6021,6 +6021,32 @@ proc semExpr(c: var SemContext; it: var Item; flags: set[SemFlag] = {}) =
 proc reportErrors(c: var SemContext): int =
   result = reporters.reportErrors(c.dest)
 
+proc buildIndexExports(c: var SemContext): TokenBuf =
+  if c.exports.len == 0:
+    return default(TokenBuf)
+  result = createTokenBuf(32)
+  for m, ex in c.exports:
+    let path = toAbsolutePath(c.importedModules[m].path)
```
Araq, line 6029: Absolute paths are always annoying.

Araq: -but we can change that later.

PR #799, opened 2025-03-17
C files are recompiled when the passC command line option was changed

demotomohiro:
Follow up to https://github.com/nim-lang/nimony/pull/780

### DIFF
--- a/src/nimony/deps.nim
+++ b/src/nimony/deps.nim
@@ -266,6 +266,9 @@ proc toBuildList(c: DepContext): seq[CFile] =
       let customArgs = i[2]
       result.add (path, obj, customArgs)
 
+proc cachedPassCFile(config: NifConfig): string =
+  config.nifcachePath / "cachedpasscfile.txt"
+
 proc generateFinalMakefile(c: DepContext; commandLineArgsNifc: string; passC, passL: string): string =
   var s = makefileHeader
   let dest =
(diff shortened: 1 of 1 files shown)


src/nimony/deps.nim
```diff
@@ -359,15 +363,25 @@ proc generateFrontendMakefile(c: DepContext; commandLineArgs: string): string =
...
-  let path = c.config.cachedConfigFile()
-  let configStr = c.config.getOptionsAsOneString() & " " & c.rootNode.files[0].nimFile
-  let needUpdate = if semos.fileExists(path) and not c.forceRebuild:
-                     configStr != readFile path
-                   else:
-                     true
-  if needUpdate:
-    writeFile path, configStr
+proc generateCachedConfigFile(c: DepContext; passC: string) =
+  block:
```
Araq, line 367: Oh god no... Why does this need yet another tiny file?

demotomohiro: When only a passC command line option was changed, only the backend C compiler and linker need to be executed.
If passC was written to `cachedconfigfile.txt` like other command line options, nimsem, Hexer and NIFC are executed even if only the passC was changed and don't need to execute them.

Araq: PassC changes are so rare that it doesn't matter, instead keep it simple.

demotomohiro: Created simpler version of this PR: https://github.com/nim-lang/nimony/pull/807

PR #800, opened 2025-03-17
hexer: implements case string

ringabout:


src/hexer/stringcases.nim
```diff
@@ -99,12 +99,15 @@ proc transformStringCase*(c: var EContext; n: var Cursor) =
...
+      inc nb
+      while nb.kind != ParRi:
+        inc nb
+        inc i
+      inc nb # skip ParRi
       traverseStmt c, nb
       c.dest.copyIntoUnchecked "jmp", info:
         c.dest.add symToken(afterwards, info)
       skipParRi nb
-      inc i
```
Araq, line 107: Removing this line cannot be correct.

PR #808, opened 2025-03-18
implement annotating constant types, fix cstring literals

metagn:
`semConstExpr` now fits the evaluated constant value to an expression matching the type of the constant. Though `semConstExpr` is only used for the values of `const` declarations which is still disabled and can be enabled in another PR after this one (it requires some test changes).

A new suffix `"C"` is added for cstring literals, previously `"R"` was used for cstring literals in expreval which is wrong, it is used for raw string literals.

### DIFF
--- a/src/hexer/nifcgen.nim
+++ b/src/hexer/nifcgen.nim
@@ -1078,10 +1078,11 @@ proc traverseExpr(c: var EContext; n: var Cursor) =
       let arg = suf
       skip suf
       assert suf.kind == StringLit
-      if arg.kind == StringLit and pool.strings[suf.litId] in ["R", "T"]:
+      if arg.kind == StringLit and pool.strings[suf.litId] == "C":
         # cstring conversion
         inc n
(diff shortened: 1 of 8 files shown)


Araq:
I think the code can be simplified if the evaluator produces typed values to begin with, probably opt-in via a mode.

metagn:
I figured the evaluator would want untyped values so they are easier to operate on, for example when evaluating:

```nim
type Foo = enum a = 1, b = 2, c = 3
Foo(int32(a) + int32(b))
```

`int32(a)` would produce `1'i32`, then this would be stripped to perform the addition, then addition would have to produce a typed node again like `3'i32`, then this would be converted to a node with type `Foo` which also has to produce a typed representation that could be operated on (so either a `Dconv` which gets skipped or the field symbol `c`). In any case we would need most of this code anyway, either to implement conversions or to produce typed nodes from the result of operations like addition as mentioned. I thought it would be easiest for the evaluator for it to be done at the end but I don't know if it's worse for correctness.

PR #819, opened 2025-03-19
Change pairs iterator return type and add mpairs iterator

demotomohiro:


lib/std/tables.nim
```diff
@@ -111,10 +111,12 @@ proc mgetOrPut*[K: Keyable, V](t: var Table[K, V]; k: sink K; v: sink V): var V
 
 proc len*[K, V](t: Table[K, V]): int {.inline.} = t.data.len
 
-iterator pairs*[K, V](t: Table[K, V]): (K, V) =
+iterator pairs*[K, V](t: Table[K, V]): (lent K, lent V) =
+  for i in 0 ..< t.data.len:
+    yield (t.data[i][0], t.data[i][1])
+
+iterator mpairs*[K, V](t: var Table[K, V]): (lent K, var V) =
```
Araq, line 118: The variant without `var Table` (instead just `Table`) should also compile and communicates clearly that `mpairs` itself does not mutate, it only allows the client to mutate things.

demotomohiro: When I add `iterator mpairs*[K, V](t: Table[K, V]): (lent K, lent V) =` to tables.nim, I got `Error: ambiguous call: 'mpairs'` in the test code that calls `mpairs`.
Is that compiler bug or not yet implemented?

demotomohiro: When I changed `mpairs` parameter type from `var Table[K, V]` to `Table[K, V]`, it compiled without error but I cannot mutate it.
```nim
import std/[hashes, syncio, tables]

var t = initTable[int, int]()
t[123] = 321

for k, v in t.mpairs:
  v = 777

echo t[123]   # output is "321"
```
When it is `var Table[K, V]`, I can mutate the table.

Araq: Well your example program must work so that's a bug.

PR #830, opened 2025-03-20
hexer: fixes local type decls

ringabout:
follow up https://github.com/nim-lang/nimony/pull/812

Lifts local type decls into the top level

### DIFF
--- a/src/hexer/duplifier.nim
+++ b/src/hexer/duplifier.nim
@@ -792,9 +792,9 @@ proc trTypeDecl(c: var Context; n: var Cursor) =
   let symId = iter.symId
   var dest = createTokenBuf()
   takeTree(dest, n)
+  c.dest.add dest
   if isLocalDecl(symId):
     publish(symId, dest)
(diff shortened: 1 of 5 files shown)


src/hexer/duplifier.nim
```diff
@@ -786,9 +786,22 @@ proc trDeref(c: var Context; n: var Cursor) =
...
+    var dest = createTokenBuf()
+    takeTree(dest, declStart)
+    publish(symId, dest)
+
 proc tr(c: var Context; n: var Cursor; e: Expects) =
   if n.kind == Symbol:
     trLocation c, n, e
+  elif n.stmtKind == TypeS:
+    trTypeDecl c, n
+    takeTree c.dest, n
```
Araq, line 804: I don't understand. So you copy the type declaration twice?!

ringabout: `trTypeDecl` is used to publish the local type decl in the `mem` so that `getType` will work. I improved the code

PR #831, opened 2025-03-20
hexer: transform while loop (eliminates `continue`)

ringabout:
todo: refactor related functions


Araq:
Quite complex. Maybe NIFC should support `continue`?

ringabout:
It is what has been done for `for loop`. Probably, we can unify the treatment of for loops and while loops to make it consistent one way or the other

PR #848, opened 2025-03-24
make intsets work

### LINKED ISSUE #675: uint shl uint causes type mismatch compile error
mytests/test.nim:
```nim
discard 1'u shl 1'u
```

```console
bin/nimsem  m nifcache/tes9hvujc.1.nif nifcache/tes9hvujc.2.nif nifcache/tes9hvujc.2.idx.nif
src/nimony/nimsem.nim(116) nimsem
src/nimony/nimsem.nim(113) handleCmdLine
src/nimony/nimsem.nim(50) singleModule
src/nimony/sem.nim(6002) semcheck
src/nimony/sem.nim(758) semStmt
src/nimony/sem.nim(5710) semExpr
src/nimony/sem.nim(3702) semDiscard
src/nimony/sem.nim(5784) semExpr
src/nimony/sem.nim(1712) semCall
/home/player/worktmpfs/nim-2.3.1/lib/system/iterators_1.nim(1544) resolveOverloads
mytests/test.nim(1, 13) Error: Type mismatch at [position]
[2] expected: (i -1) but got: (u -1)
[1] expected: +16 but got: -1
[1] expected: (i +16) but got: (u -1)
[1] expected: +32 but got: -1
[1] expected: (i +32) but got: (u -1)
[1] expected: (i +8) but got: (u -1)
[1] expected: (i +64) but got: (u -1)
[1] expected: +8 but got: -1
Type mismatch at [position]
[2] expected: (i -1) but got: (u -1)
[1] expected: +16 but got: -1
[1] expected: (i +16) but got: (u -1)
[1] expected: +32 but got: -1
[1] expected: (i +32) but got: (u -1)
[1] expected: (i +8) but got: (u -1)
[1] expected: (i +64) but got: (u -1)
[1] expected: +8 but got: -1
```

lib/std/hashes.nim uses uint and `shl`.

### LINKED ISSUE #847: array indexing does not work
Extracted from intsets.nim:

```nim
import std / [syncio, tables, hashes]

const
  UIntSize = 64'u

const
  TrunkSize = 8'u
  BitsPerTrunk = TrunkSize.uint * UIntSize

type
  Trunk = object
    a: array[TrunkSize, uint]

  IntSet* = object
    t: Table[uint, Trunk]

proc initIntSet*(): IntSet = IntSet(t: initTable[uint, Trunk]())

proc split(x: uint): (uint, uint, uint) {.inline.} =
  (x div BitsPerTrunk, (x mod BitsPerTrunk) div UIntSize, x mod UIntSize)

proc incl*(s: var IntSet; x: int) =
  let (a, b, c) = split cast[uint](x)
  let tr = addr(s.t.mgetOrPut(a, default(Trunk)))
  discard tr.a[b]
  #= tr.a[b] or (1'u shl c)
```

metagn:
closes #847, closes #675

1. The array index overload did not allow uint indexes when the index type had an int base type, the index is now left underspecified as its own generic parameter. No constraints since `array[I, T]` usually has no constraints on `I` so it would be really annoying for typechecked generics
2. `SomeInteger` workaround in `shr`/`shl` is expanded to add overloads for `uint`. Still a workaround though, `int | uint` does not work either since concrete constraints are not implemented
3. `shr`/`shl` semcheck is corrected to allow different type for the RHS
4. ~~Mangled types now normalize bits~~ moved to #850

Not necessary here but noticed when trying to fix:

5. or/and types did not always skip `f` on failure of `matchesConstraint`, just added separate variable for it
6. `T: Ordinal` now matches another `Ordinal` typevar


metagn:
I thought array indexes in original Nim had a special type check but it does seem like they just call sigmatch which only makes this work because of a [specific edge case for range types](https://github.com/nim-lang/Nim/blob/0b9ed84d32c45f036c44d6235cb6c3bf3c20d202/compiler/sigmatch.nim#L570-L571), even index types like `range[0'i64..2'i64]` break it. Sigmatch currently does not match int to uint and vice versa unless a literal is given, so they are not convertible between each other. Maybe range types can have a new matching behavior like "any type that contains the range can be converted to the range base type" since then we could do the range check in the original integer type and this would make sure the integer type conversion is safe too, maybe via a generalized `(chckrange (rangetype basetype a b) n)`

Araq:
My attempt is https://github.com/nim-lang/nimony/pull/849 which works except that the codegen needs to canonicalize `u -1` to `u 64` as you have noticed.

metagn:
Sorry, I understand now, #849 should be fine and original Nim behaves similarly by just copying the type of the literal. But we will probably have this problem again of trying to index int indexed arrays with uints.

Araq:
Well please put the codegen bugfix into its own PR.

Araq:
The sigmatch fix needs to be its own PR, the rest was done better by my PR, IMO.

Issue #851, opened 2025-03-25
untyped templates do not work as expected


Araq:
As far as I understand the untyped templates, this code should compile:

```nim
import std / [syncio]

type
  Inner = object
    s: string
  Function = object
    inner: Inner

template emit(s: string) {.untyped.} = f().inner.s.add s

proc useGlobal =
  template f(): untyped {.untyped.} = myObj
  var myObj = Function(inner: Inner(s: "abc"))
  emit " xyz"

  echo myObj.inner.s

useGlobal()
```

But it doesn't.

metagn:
Untyped pragma is ignored without compat mode, also the template here does not work here in regular Nim either because of the `.s`

PR #855, opened 2025-03-25
enable untyped pragma without compat mode

metagn:
refs #851

### DIFF
--- a/tests/nimony/untyped/tuntypedtemplnocompat.nim
+++ b/tests/nimony/untyped/tuntypedtemplnocompat.nim
@@ -1,0 +1,18 @@
+import std / [syncio]
+
+type
+  Inner = object
+    s: string
+  Function = object
+    inner: Inner
+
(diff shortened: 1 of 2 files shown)


tests/nimony/untyped/tuntypedtemplnocompat.nim
```diff
@@ -0,0 +1,18 @@
+import std / [syncio]
+
+type
+  Inner = object
+    s: string
+  Function = object
+    inner: Inner
+
+template emit(s2: string) {.untyped.} = f().inner.s.add s2
```
Araq, line 9: Nim compiles with with `s: string` btw, I think.

metagn: https://play.nim-lang.org/#pasty=QrXzWNid, also tested devel, doesn't work, some code probably depends on this too

PR #857, opened 2025-03-25
make `..` generic

metagn:


metagn:
To stop having to update so many tests all the time I also want to move a fair amount of the code in system to include files but in a PR after this one so that the diff is easier to understand.

PR #858, opened 2025-03-25
move most of system to include files

metagn:
refs https://github.com/nim-lang/nimony/pull/857#issuecomment-2752267545


Araq:
We need to understand why now the silly (expr (expr ...)) got removed.

metagn:
Sorry, forgot to mention I also removed the unary `succ`/`pred` template overloads in favor of default params (previously I tried `y: V = V(1)` which doesn't infer `V` so I didn't think it would work but `T(1)` works). `inc`/`dec` also would have worked with default params but I didn't merge them since they generate full procs for now.

PR #887, opened 2025-03-29
try fix bare typedesc

metagn:
refs #880


metagn:
`typedesc` expands to `(typekind (typedesc))` which fails due to no inferred generics. The error message is bad here because this fails the match and causes a template recursion, which I have a fix for

Issue #826, opened 2025-03-20
operations to implement in expreval


metagn:
Hopefully expreval will not need to be a full fledged VM but there are still operations we probably need to support for basic const evaluation.

Implemented as magics:

* [ ] arithmetic, comparison
* [ ] numeric type conversions, casts
* [ ] `sizeof` etc
* [ ] set ops

Implemented as procs in system but will probably be specialized in expreval:

* [ ] string ops, maybe just basic ones like `&`, `len`, comparisons, as strings are represented by string literal tokens in expreval
* seq ops and openarrays are probably not used too commonly
* maybe set ops can be implemented as procs and overriden in expreval as well

These could be implemented via pragmas like `{.vmop: "string.len".}` or by hard coding the symbol names in the compiler which might get annoying. But the pragmas would probably build a table from symbols to their respective operations anyway, just at compiler runtime.

PR #891, opened 2025-03-29
builtin ops for expreval

metagn:
refs #826

Procs have a `semantics` pragma that is searched when a proc is called in expreval, then special cased as if they were magics. Currently string `&`, `len`, `==` are implemented this way. `concat` is a template unfortunately so it will expand before we can treat it as a builtin.


Araq:
I think this should `builtin` should be called `semantics` so that it makes it more unlikely to clash with a custom pragma named `builtin`.

Araq:
Actually Nim has an undocumented `exportnims` pragma which does almost the same thing. We could use it here but then we didn't care about name clashes for the new `build` pragma either.

Your choice.

PR #897, opened 2025-03-30
newSeq: Set len to zero on OOM.

planetis-m:


Araq:
You need to update the tests via `src/hastur --overwrite nimony`

PR #898, opened 2025-03-31
implement fields/fieldPairs

### LINKED ISSUE #893: implement system.fieldPairs etc.
These are required for general purpose `toNif`/`fromNif` operations which we could use for a general "eval at compile-time" mechanism.

metagn:
closes #893

Performs direct identifier substitution for simplicity for now because we would need a prepass to define/substitute symbols. This means we can't shadow the loopvars but this wasn't fixed in Nim either until recently (maybe still isn't?)

The fields are also not rigorously typed and just built as `(dot obj fieldname)` since we are going to pass everything to typechecking anyway.

When the type of the given object/tuple is a bare typevar, the entire loop is treated as an untyped expression.


src/nimony/sem.nim
```diff
@@ -616,13 +616,17 @@ proc semConstExpr(c: var SemContext; it: var Item) =
 
 proc semStmtsExpr(c: var SemContext; it: var Item; isNewScope: bool) =
   let before = c.dest.len
+  if isNewScope:
+    openScope c
```
Araq, line 620: Should not be necessary as `sem` generates `ScopeS` nodes, but does not depend on them for scope creation. Instead all the control flow constructs already know when they need to introduce new scopes.

metagn: Here I am building `(scope)` for each unroll of the loop before passing it to typechecking, I can replace it with `if true` but I figured this might be less complex. `block` would interfere with `break`.

PR #908, opened 2025-04-02
adapt original int64 etc matching rules

metagn:
Integer type matching is brought closer to original sigmatch:

* `int`/`uint` now match their respective bit type (i.e. `int64`/`uint64`) with `isIntConv`, vice versa matches with `isConvertible`
* Integer types with bigger bits than the given type are supposed to match with `isConvertible` and only match with `isIntConv` if the formal type is `int`/`uint`

This means that `int`/`uint` are not considered the same as their corresponding bit types anymore and generate conversions (although `linearMatch` is not updated for this, in original Nim `array[2, int64]` does not match `array[2, int]` but here it does). So overloads for `int`/`uint` are added for each of the numeric operation primitives, same as original Nim. Generic versions of `min`/`max` are also added.

There is a pre-existing behavior in original Nim that an int literal like `1` is ambiguous between `int64` and `uint64`, `syncio.write` is given overloads for `int`/`uint` as well to account for this.

### DIFF
--- a/lib/std/hashes.nim
+++ b/lib/std/hashes.nim
@@ -24,8 +24,7 @@ when not defined(nimony):
     result = !$result
 
 proc hash*(u: uint): Hash {.inline.} = u
-when not defined(nimony):
-  proc hash*(x: int): Hash {.inline.} = cast[Hash](x)
+proc hash*(x: int): Hash {.inline.} = cast[Hash](x)
 
 proc hash*(x: int64): Hash {.inline.} = cast[Hash](x)
 proc hash*(x: int32): Hash {.inline.} = cast[Hash](int x)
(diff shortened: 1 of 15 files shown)


Araq:
If you change min/max anyway, try to make them generic please.

metagn:
Without this int/uint just behave like aliases to int64/uint64.

```nim
let a: int = 1
let b = a + a
```

If `+` picks the int64 overload here, the type of `b` becomes int64 and not int. I figured since this somewhat loses information it wouldn't be the desired behavior, otherwise instead of normalizing bits everywhere (`typeToCanon` for generics still does not for example) we could eliminate `(i -1)` wherever we can.

Araq:
> we could eliminate (i -1) wherever we can.

Yup, that's the way to go.

metagn:
Since the new overloads are needed for this to work, adding it to compat mode would require a conditional definition of those overloads which might be annoying to maintain so we can postpone this until it's needed

Issue #909, opened 2025-04-02
generic converters


Araq:
From looking at the codebase, it seems like generic converters where the generic is then instantiated in a different module cannot possibly work.

PR #910, opened 2025-04-02
test generic converters across modules

metagn:
refs #909


metagn:
Seems to work, unless this is not the issue described in #909?

tests/nimony/converter/deps/mgenericconverter.nim
```diff
@@ -0,0 +1,9 @@
+type Foo*[T] = object
+  val: T
+
+converter toFoo*[T](x: T): Foo[T] = Foo[T](val: x)
+
+proc bar[T](x: Foo[T], y: T) = discard
+
+bar(123, 456)
+let foo: Foo[int] = 123
```
Araq, line 9:
```suggestion
```

Test works even better when no instantiation is done at all.

Issue #903, opened 2025-04-01
seqs and strings need to use overflow checks for safety


Araq:
- It's ok to map overflows to `high(int)` which should then produce OOMs.

Araq:
In case it isn't obvious, this should use the new `{.keepOverflowFlag.}` and `overflowFlag` mechanisms.

PR #913, opened 2025-04-02
adds overflow checks to strings

demotomohiro:
Implements https://github.com/nim-lang/nimony/issues/903 for strings

I think overflow checks are required only when adding two strings.
And if system memory size is smaller than or equals to 2^62 bytes, you will get out of memory error before getting overflow.

In `max(newLen, len + (len shr 1))`, if `len + (len shr 1)` overflowed, its value is smaller than `len` and `newLen`.
So `max(newLen, len + (len shr 1)) == newLen`.

When adding single character to a string, `len+1` overflows only if `len == high(int)`. But if `len == high(int)`, it fails to allocate memory.

### DIFF
--- a/lib/std/system/stringimpl.nim
+++ b/lib/std/system/stringimpl.nim
@@ -150,9 +150,16 @@ proc makeAllocated(s: var string; newLen: int) =
     oomHandler newCap
     s.i = EmptyI
 
+proc sumLen(a, b: int): int =
+  {.keepOverflowFlag.}:
+    result = a + b
+    if overflowFlag():
+      # When required length is overflowed, cause out of memory.
+      result = high(int)
+
 proc add*(s: var string; part: string) =
   let len = s.len
-  let newLen = len + part.len
+  var newLen = sumLen(len, part.len)
   if not isAllocated(s):
     makeAllocated s, newLen
   else:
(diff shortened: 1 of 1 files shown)


Araq:
> And if system memory size is smaller than or equals to 2^62 bytes, you will get out of memory error before getting overflow.

We also support 32bit machines... And we try to support 16 bit machines.

demotomohiro:
If we support 32bit machines with 4gb system memory or 16bit machines with 64kb and one string can use more than 50% of system memory, string length must be an unsigned int type or int64/int32 type.
If a string length is always smaller than high(int), I think overflow checks in this PR are also fine on 32bit or 16bit machines.

Araq:
So what about this line:

`let newCap = len+1`

demotomohiro:
`len+1` overflows only when `len == high(int)`, but `len` can be `high(int)`?
Don't we assume allocating high(int) bytes always fail?
Even if allocating high(int) bytes always fail, are there cases `len` become `high(int)`?
`len` can be `high(int)` when the string is const?
`len` can be `high(int)` even if the string is not pointing to the high(int) bytes memory?

Araq:
Good points.

PR #914, opened 2025-04-03
misc fixes extracted from #908

metagn:
1. generic min/max
2. minor bug in sigmatch
3. consider derefs of `lent` types in `getType`

### DIFF
--- a/lib/std/system/comparisons.nim
+++ b/lib/std/system/comparisons.nim
@@ -106,6 +106,9 @@ template `>`*(x, y: untyped): untyped =
   ## "is greater" operator. This is the same as `y < x`.
   y < x
 
+type Orderable = concept
+  proc `<=`(x, y: Self): bool
+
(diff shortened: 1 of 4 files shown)


lib/std/system/comparisons.nim
```diff
@@ -106,6 +106,9 @@ template `>`*(x, y: untyped): untyped =
   ## "is greater" operator. This is the same as `y < x`.
   y < x
 
+type Orderable = concept
+  proc `<=`(x, y: Self): bool
+
 proc min*(x, y: int8): int8 {.noSideEffect, inline.} =
```
Araq, line 112: Why do we still need these?

metagn: Original system is the same having both these specified overloads and the generic one, also the implementation for float is different due to NaN. Shouldn't be needed otherwise

PR #916, opened 2025-04-03
string: add more cstring conversion utils

mrgaturus:
- add more terminating zero utils
- add the missing `fromCString` mentioned
- fix a possible redundant string copy in syncio `open`
- add cstring tests

### DIFF
--- a/lib/std/syncio.nim
+++ b/lib/std/syncio.nim
@@ -89,8 +89,10 @@ proc open*(f: out File; filename: string;
     of fmReadWriteExisting: cstring"r+b"
     of fmAppend: cstring"ab"
 
+  # XXX: avoid a possible double copy when the string is allocated
   var tmpFilename = filename.terminatingZero()
-  f = fopen(tmpFilename.toCString, m)
+  let rawFilename = cast[cstring](tmpFilename.rawData)
+  f = fopen(rawFilename, m)
   if f != nil:
     result = true
     if bufSize >= 0:
--- a/lib/std/syncio.nim
+++ b/lib/std/syncio.nim
@@ -89,7 +89,7 @@ proc open*(f: out File; filename: string;
     of fmReadWriteExisting: cstring"r+b"
     of fmAppend: cstring"ab"
 
-  var tmpFilename = filename
+  var tmpFilename = filename.terminatingZero()
   f = fopen(tmpFilename.toCString, m)
   if f != nil:
     result = true
--- a/lib/std/system/stringimpl.nim
+++ b/lib/std/system/stringimpl.nim
@@ -10,6 +10,15 @@ const
   LenShift = 1
   IsAllocatedBit = 1
 
+proc sumLen(a, b: int): int =
+  {.keepOverflowFlag.}:
+    result = a + b
+    if overflowFlag():
+      # When required length is overflowed, cause out of memory.
+      result = high(int)
+
+# --- string attributes ---
+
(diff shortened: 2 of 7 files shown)


lib/std/system/stringimpl.nim
```diff
@@ -70,30 +84,35 @@ proc `=dup`*(s: string): string {.exportc: "nimStrDup", inline, nodestroy.} =
...
-func strlen(a: cstring): csize_t {.importc: "strlen", header: "<string.h>".}
-
 proc borrowCStringUnsafe*(s: cstring): string =
   ## Creates a Nim string from a `cstring` by borrowing the
   ## underlying storage. You have to ensure the `cstring` lives
   ## long enough for this to be safe! If in doubt,
   ## use `fromCString` instead.
   string(a: cast[StrData](s), i: (int(strlen(s)) shl LenShift))
 
+proc isTerminatingZero*(s: string): bool {.inline.} =
```
Araq, line 105: 1. Name should be `hasTerminatingZero`.
2. This proc sounds like a bad idea as it introduces non-determinism for allocations that don't initialize which we generally prefer these days.

mrgaturus: removing it makes sense despite i can implement it somewhere in my code thanks to `rawData`

lib/std/system/stringimpl.nim
```diff
@@ -117,12 +133,38 @@ proc ensureTerminatingZero*(s: var string) =
...
+    # ensure zero termination anyway:
+    result = string(i: EmptyI)
+    result.a = cast[StrData](cstring"")
+
+proc fromCString*(s: cstring): string =
+  ## Creates a Nim string from a `cstring`
+  ## by copying the underlying storage.
+  ## The created string has a null terminator
+  let aux = borrowCStringUnsafe(s)
+  result = aux.terminatingZero()
```
Araq, line 158: But why? You cannot access this zero.

lib/std/system/stringimpl.nim
```diff
@@ -117,12 +133,38 @@ proc ensureTerminatingZero*(s: var string) =
       s.a = cast[StrData](cstring"")
       s.i = EmptyI
 
+proc terminatingZero*(s: string): string =
```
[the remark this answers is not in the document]
mrgaturus: using `s & "\0"` doesn't ensure an empty but terminated zero string at out of memory

lib/std/syncio.nim
```diff
@@ -89,7 +89,7 @@ proc open*(f: out File; filename: string;
     of fmReadWriteExisting: cstring"r+b"
     of fmAppend: cstring"ab"
 
-  var tmpFilename = filename
+  var tmpFilename = filename.terminatingZero()
   f = fopen(tmpFilename.toCString, m)
```
Araq, line 93: you need a new `toCString`-like proc here that is allowed to assume the terminating zero.

mrgaturus: might be a toCString returning a `tuple[tmp: string, s: cstring]` would do the trick

mrgaturus: Other way could be a `distinct string` type that has a converter to cstring

Araq: Yes. Or: You simply do what I tell you.

mrgaturus: the quirky way to change the toCString here is use a `cast[cstring](tmpFilename.rawData)`

Araq: Good enough for now.

PR #924, opened 2025-04-05
match concrete types in constraints

### LINKED ISSUE #675: uint shl uint causes type mismatch compile error

metagn:
closes #675

When matching against a constraint, if the constraint type is not found to be a typeclass, it is matched like a regular type. `linearMatch` is used for this, which for now means conversions from the argument to the constraint type are not allowed (i.e. `"abc"` matching `T: cstring` or `1'i8` matching `T: int`) which is a difference from original Nim (it would also need to infer the typevar to the *converted* type and not the argument type, not to mention considering converters etc).

This is done so that `SomeInteger` can now be implemented as intended and used for `shl`/`shr` etc. However implicit generics are not implemented so the signatures use an explicit generic parameter for it.

To make the call to `linearMatch` here less hacky, `m.err` now just tracks if the match failed so it is harmless to temporarily set it to `false`, and a new field `hasError` is added to track if the error message should be updated.

Generic base symbols i.e. `Foo` in `Foo[T]` are also now considered typeclasses and match their instances/themselves.

### DIFF
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -58,6 +58,7 @@ type
     args*, typeArgs*: TokenBuf
     err*, flipped*: bool
     concreteMatch: bool
+    hasError: bool # mark that error message was set
     skippedMod: TypeKind
     argInfo: PackedLineInfo
     pos, opened: int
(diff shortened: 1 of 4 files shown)


src/nimony/sigmatch.nim
```diff
@@ -271,14 +325,8 @@ proc matchesConstraint(m: var Match; f: var Cursor; a: Cursor): bool =
...
-    let res = tryLoadSym(f.symId)
-    assert res.status == LacksNothing
-    var typeImpl = asTypeDecl(res.decl)
-    if typeImpl.kind == TypeY:
-      result = matchesConstraint(m, typeImpl.body, a)
-    inc f
-  else:
-    result = matchesConstraintAux(m, f, a)
+    return matchSymbolConstraint(m, f, a)
+  result = matchesConstraintAux(m, f, a)
```
Araq, line 329:
```suggestion
    result = matchSymbolConstraint(m, f, a)
  else:
    result = matchesConstraintAux(m, f, a)
```

Early returns break symmetry and suck.

PR #925, opened 2025-04-05
disable winget install dependency, use windows-2025

metagn:
The dependency is broken, apparently windows-2025 and above do not need it anyway since winget can be installed without any dependencies


metagn:
Will merge when CI passes since CI is broken otherwise.

Also looked into why the windows CI lasts 9 minutes when the macos/linux CIs last 1-2. I assumed maybe a single test was hanging it up, but it's how long tests take in general, running a test on average takes 400 milliseconds on macos/linux but takes more than 3 seconds on windows.

PR #927, opened 2025-04-06
setops: add xor funcs using `XorSet` magic

mrgaturus:
there is an unused `XorSet` magic, it makes sense create simple funcs in `system/setops` using that magic. XOR operations in sets are very useful for gui programming (checkboxes, toggle buttons, etc) and detect flags changes in sets (reactive state changes).

in docs we need specify that it's equivalent to a XOR operation instead only call it "symmetric difference"

### DIFF
--- a/lib/std/system/setops.nim
+++ b/lib/std/system/setops.nim
@@ -1,10 +1,10 @@
 func incl*[T](x: var set[T], y: T) {.magic: "Incl".}
 
+func excl*[T](x: var set[T], y: T) {.magic: "Excl".}
+
 template incl*[T](x: var set[T], y: set[T]) =
   ## Includes the set `y` in the set `x`.
   x = x + y
-
-func excl*[T](x: var set[T], y: T) {.magic: "Excl".}
 
 template excl*[T](x: var set[T], y: set[T]) =
   ## Excludes the set `y` from the set `x`.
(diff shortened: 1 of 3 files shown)


metagn:
In original Nim this is exposed as [`setutils.symmetricDifference`](https://github.com/nim-lang/Nim/blob/10c9ebad9303d9c4be393da913f4e12650783539/lib/std/setutils.nim#L79-L85), full discussion here: https://github.com/nim-lang/Nim/pull/24286

mrgaturus:
> ref [nim-lang/Nim#24286](https://github.com/nim-lang/Nim/pull/24286), is exposed in setutils in original Nim

I know (also I asked for it 🤣), but I wanted it promoted to system because it's a magic and it's more useful than `card(set[T])`, if the posture of keeping it quite hidden stays I think i will close this PR

Araq:
It's nice but doesn't belong into system.nim. Add std/setutils to Nimony?

mrgaturus:
> It's nice but doesn't belong into system.nim. Add std/setutils to Nimony?

Most stuff of std/setutils in nim 2.x uses macro things so it looks not worth it create a module with just 3 procs (8 lines). I think it's still too early think about the other misc parts of stdlib so let's revisit it more later

PR #928, opened 2025-04-06
std: introduce setutils + add an items iterator to set[T]

mrgaturus:
- introduce `std/setutils` like nim 2.x
- add an items iterator to `set[T]` in `system/setops`
- add tests of `std/setutils` and `set[T]` items iterator

### DIFF
--- a/lib/std/setutils.nim
+++ b/lib/std/setutils.nim
@@ -2,9 +2,11 @@ func symmetricDifference*[T](x, y: set[T]): set[T] {.magic: "XorSet".}
 
 func `-+-`*[T](x, y: set[T]): set[T] {.magic: "XorSet".}
  
-template toggle*[T: Ordinal](x: var set[T], y: T) =
+proc toggle*[T: Ordinal](x: var set[T], y: T) {.inline.} =
   ## Toggles the element `y` in the set `x`.
-  x = x.symmetricDifference({y})
+  if x.contains(y):
+    x.excl(y)
+  else: x.incl(y)
 
 template toggle*[T: Ordinal](x: var set[T], y: set[T]) =
   ## Toggles the set `y` in the set `x`.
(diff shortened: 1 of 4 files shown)


lib/std/setutils.nim
```diff
@@ -0,0 +1,45 @@
...
+
+func complement*[T: Ordinal](s: set[T]): set[T] =
+  ## Returns a `set[T]` with the elements that are not inside
+  fullSet(T) - s
+
+func isEmpty*[T: Ordinal](s: set[T]): bool {.inline.} =
+  ## Check if the `set[T]` is empty, same as `s == {}`
+  s == {}
+
+func isFull*[T: Ordinal](s: set[T]): bool {.inline.} =
```
Araq, line 31: are these all in Nim 2 too?

mrgaturus: those are new ones but it could be backported to Nim 2, the new ones are `emptySet`, `isEmpty`, `isFull`, `[]` and the `toggle` with the T type. some looks pointless but it's better make it quite symmetric (full implies the existence of empty)

mrgaturus: original nim 2.x `setutils` only has `symmetricDifference`, `-+-`, `toggle`, `fullSet`, `complement` and `[]`

Araq: Please only add what Nim 2 offers.

mrgaturus: what if i backport first the new ones to nim 2?

Araq: That's not how it works. Most API additions require some arguing first or even an RFC.

mrgaturus: ok so I will remove the new ones to match the setutils of nim 2.x

lib/std/system/setops.nim
```diff
@@ -39,3 +39,12 @@ proc cardSetImpl(s: ptr UncheckedArray[uint8], len: int): int {.inline.} =
...
   result = cardSetImpl(s, len)
+
+iterator items*[T: Ordinal](x: set[T]): T =
+  ## Iterates over all elements inside the `set[T]`.
+  ## This iterator walks from `low(T)` to `high(T)`
+  ## yielding each element found
+  var i = low(T)
+  while i <= high(T):
+    if x.contains(i): yield i
+    inc(i)
```
Araq, line 50:
```suggestion
  var i = low(T)
  if i <= high(T):
    while true:
      if x.contains(i): yield i
      if i >= high(T): break
      inc(i)
```

This way it works for T=byte with values 0..255 without wraparound/overflows.

PR #931, opened 2025-04-07
implement typeclasses matching typeclasses

metagn:
I am not 100% sure this makes sense and I have not tested every case that might be relevant, but hopefully this is a minimum working state that should not be too hard to change later on.

Sem fixes: `and` and `not` types are implemented as direct syntax checks for now same as `|` (which now also includes `or`), this could maybe be changed to a magic in the future. Also standalone `array`, `OrdinalEnum` etc symbols now compile to the typeclass everywhere except invocation heads instead of only in generic constraints, so that `type Foo = array` etc works properly. Could split these from the PR.

When matching the constraints of a typevar, if the given typevar is a typevar itself (i.e. in generic typechecking), the constraints of that typevar are used for the match. This means that typeclasses may need to match other typeclasses, i.e. `A and not B` may need to match `A or B`. In the common case of recursion, a typeclass needs to match itself as well. However these types by themselves will not be recognized by most typeclasses, they need to be "broken down" so that the typeclasses can understand them (this depends on the typeclass but most of them need it), but without losing information. This is done in the following order:

1. if the given argument is an `or` type, each branch of the argument `or` type is matched to the constraint, such that every possible branch matches
2. the `and`/`or`/`not` type constraints are matched
3. now that the constraint is "atomic", if the given argument is an `and` type, each branch of the `and` type is matched to the constraint, such that at least one branch matches

There are some notable missing steps here to keep it simpler for the time being:

**1.** This order is only correct if the argument type is in "sum of products" form, for example it works for `(A and B) or (C and D)` but not `(A or B) and (C or D)`. To fix this another step is needed to reorder the argument type to fit this format before doing anything or at least process it as if it was reordered, i.e. `(A or B) and (C or D)` becomes `(A and C) or (A and D) or (B and C) or (B and D)`. `not (A or B)` would probably also have to fold to `not A and not B`. Without doing this, information is lost along the way, for example when matching `(A or B) and (C or D)` to `(A or B)`:

1. splitting argument `or` does nothing
2. `A` and `B` are individually matched to `(A or B) and (C or D)`, consider `A` (we already fail here by matching it to `(A or B)`)
3. `A` matches `(A or B)` and `(C or D)` individually, assume it recurses from step 1 and matches `(A or B)`
5. `A` matches `A` and `B` individually and expects both to match, `A` matching `B` fails

It might be possible to make this work for "product of sums" form as well by shifting the order, but this order seems to be the nicest.

**2.** Nothing is done with argument `not` types yet; since a type `A` generally does not match `not A`, `not A` happens to match itself. In original Nim, if both the argument and the constraint are `not` types, the `not` part is removed and the match is flipped, i.e. `not A` matches `not B` if `B` matches `A`. To do this here, we would probably track the "negation" state of the constraint as a boolean (so that `not not` resets to normal), then keep track of the negation state of the argument when removing the `not` from it and possibly flip the match if both are negated.

**3.** Finally, there is a special case of the "atomic" typeclasses that is not handled properly: typevars themselves as constraints. If a typevar `T2` is given then matching the constraint `T1` should not receive `T2`'s broken down constraints but `T2` itself. The solution might be to keep track of the full typevar we are matching against then use this for inference rather than specific constraint types, and walk this back later when the match fails. But this is too complex and too niche to do here for now.


metagn:
Drafting again to split for #966 and fix #952 in the meantime.

PR #933, opened 2025-04-07
Replace Biggest types in nifbuilder

demotomohiro:


metagn:
We should probably have these in system anyway, in original Nim they are in [ctypes](https://github.com/nim-lang/Nim/blob/devel/lib/system/ctypes.nim) which we should be able to add as an include file to system except for `cstringArray` (and without the deprecated symbols of course)

Araq:
Yes but this is an improvement regardless as the BiggestTypes have outlived their usefulness (cannot change the impl without breaking too much code).

PR #936, opened 2025-04-07
adds open proc that returns File to syncio

demotomohiro:
`open` proc returns `File` is used in nifbuilder.

### DIFF
--- a/lib/std/syncio.nim
+++ b/lib/std/syncio.nim
@@ -100,6 +100,18 @@ proc open*(f: out File; filename: string;
   else:
     result = false
 
+proc open*(filename: string,
+            mode: FileMode = fmRead, bufSize: int = -1): File =
+  ## Opens a file named `filename` with given `mode`.
+  ##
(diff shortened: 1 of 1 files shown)


lib/std/syncio.nim
```diff
@@ -100,6 +100,18 @@ proc open*(f: out File; filename: string;
...
+            mode: FileMode = fmRead, bufSize: int = -1): File =
+  ## Opens a file named `filename` with given `mode`.
+  ##
+  ## Default mode is readonly. Raises an `IOError` if the file
+  ## could not be opened.
+  result = default(File)
+  if not open(result, filename, mode, bufSize):
+    # TODO: raise exception when it is supported.
+    #raise newException(IOError, "cannot open: " & filename)
+    discard
```
Araq, line 113: Use `quit` here.

demotomohiro: Fixed it.

PR #939, opened 2025-04-08
fixes destructors don't work for `return`

ringabout:
```nim
import std/syncio

proc cho(x: string): string =
  result = x

proc bar(): string =
  var x = "1212334"
  x.add "123"
  return cho(x)
  # return if true: x else: "12343"

echo bar()
```
It doesn't work because the parameter `x` is destroyed before returning `cho(x)`. We may make a cursor into the return expression so that destroying happens after the assigning `tmp = cho(x); destroy(x)`. I will check `duplifier` later

### DIFF
--- a/src/hexer/xelim.nim
+++ b/src/hexer/xelim.nim
@@ -72,6 +72,19 @@ proc declareTemp(c: var Context; dest: var TokenBuf; n: Cursor): SymId =
     copyTree dest, typ # type
     dest.addDotToken() # value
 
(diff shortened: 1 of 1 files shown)


Araq:
duplifier must transform this into:

```nim
proc bar(): string =
  var x = "1212334"
  x.add "123"
  result = cho(x)
  return result
```

ringabout:
succeeded by https://github.com/nim-lang/nimony/pull/944

PR #942, opened 2025-04-08
adds $ operator for ints

demotomohiro:
nifbuilder uses `$` for int and uint.


Araq:
```nim
const
  ten = [
    "0", "1", "2", "3", "4",
    "5", "6", "7", "8", "9"]

proc `$`*(x: uint64): string =
  result = ""
  var y = x
  while true:
    result.add ten[int(y mod 10'u)]
    y = y div 10'u
    if y == 0'u: break
  let L = result.len-1
  var i = 0
  let b = result.len div 2
  while i < b:
    let ch = result[i]
    result[i] = result[L-i]
    result[L-i] = ch
    inc i

proc `$`*(x: int64): string =
  if x < 0:
    if x == -1:
      # -1 is so common that it deserves a string literal which avoids the allocations
      result = "-1"
    if x == -9223372036854775808:
      result = "-" & $cast[uint64](x)
    else:
      result = "-" & $(0-x)
  elif x < 10:
    result = ten[x]
  else:
    result = $cast[uint64](x)

proc addInt*(s: var string; x: int64) {.inline.} =
  s.add $x

proc addInt*(s: var string; x: uint64) {.inline.} =
  s.add $x
```

Araq:
^ The above is a more interesting implementation (and likely faster for the common cases).

demotomohiro:
Thank you!
I used your code but changed to not using `ten` look up table.

PR #944, opened 2025-04-09
fixes ret returning complex expressions destruction

ringabout:


src/hexer/duplifier.nim
```diff
@@ -161,12 +161,59 @@ proc isResultUsage(c: Context; n: Cursor): bool {.inline.} =
...
+    else:
+      result = false
+      skip n
+  of ParRi, SymbolDef, UnknownToken, EofToken:
+    result = false
+    inc n
+
 proc trReturn(c: var Context; n: var Cursor) =
-  copyInto c.dest, n:
+  var ret = createTokenBuf()
```
Araq, line 199: Instead of the `ret` temporary and the swap dance, use this:

```nim
  let retVal = n.firstSon
  var exp = retVal
  if isResultUsage(c, retVal):
    copyInto(c.dest, n): takeTree c, n
  elif isSimpleExpression(exp) or c.resultSym == NoSymId:
    copyInto(c.dest, n): tr(c, n, WantOwner)
 else:
   ...

```

PR #948, opened 2025-04-09
fixes return if expr: lowerExprs before injectDups

ringabout:
`injectDups` won't work properly without transforming if expressions (i.e. lowerExprs). Otherwise, we need to treat `if` differently in `injectDups` instead of handling it like statements, which causes double free


Araq:
Yeah, control flow inside expressions simply sucks and the rest of the compiler becomes better once eliminated.

PR #968, opened 2025-04-12
include holey enums in `enum` typeclass, add `sumGeneric` disambiguation

### LINKED ISSUE #952: `enum` typeclass does not include holey enums
It checks for the `EnumT` type kind which holey enums do not have. Fixing this would mean the `Ordinal` typeclass cannot match `enum` anymore though, but we should be able to make it so that `enum and Ordinal` matches `enum and Ordinal`.

metagn:
closes #952

The `enum` typeclass now matches holey enums, and so the `Ordinal` typeclass does not match the `enum` typeclass anymore. Because of this the mutual generic match disambiguation fails now between the `Ordinal` and `enum` overloads of `==` etc so the `sumGeneric` method of disambiguation in the original compiler is added which considers `enum` more specific than `Ordinal`.

Another `items` overload for sets of `enum` is added since the original one only matches `Ordinal`. This one converts the enum to `int` to iterate over as in the original system implementation. The `items` overload needs a constraint so that it can match the `low`/`high` overloads, `Ordinal | enum` does not work since constraints cannot match other `or` types right now (#931). Maybe we can remove the constraint on the magic overloads of `low`/`high` though.

### DIFF
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -288,6 +288,9 @@ proc matchesConstraintAux(m: var Match; f: var Cursor; a: Cursor): bool =
     inc f
     assert f.kind == ParLe
     result = aTag.kind == ParLe and f.tagId == aTag.tagId
+    if not result and f.tagId == TagId(EnumTagId):
+      # enum typeclass also accepts holey enums
+      result = a.typeKind == HoleyEnumT
     inc f
     assert f.kind == ParRi
     inc f
@@ -300,7 +303,9 @@ proc matchesConstraintAux(m: var Match; f: var Cursor; a: Cursor): bool =
     of TypeKindT:
       var aTag = a
       inc aTag
-      result = isOrdinalTypeKind(aTag.typeKind)
+      result = isOrdinalTypeKind(aTag.typeKind) and
+        # enum typeclass can also include holey enums which are not ordinal
+        aTag.typeKind != EnumT
     else:
       result = isOrdinalType(a)
     skip f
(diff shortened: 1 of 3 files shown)


src/nimony/sigmatch.nim
```diff
@@ -1220,6 +1333,8 @@ proc cmpMatches*(a, b: Match): DisambiguationResult =
     else:
       if a.fn.typ.typeKind == ParamsT and b.fn.typ.typeKind == ParamsT:
         result = mutualGenericMatch(a, b)
+        if result == NobodyWins:
+          result = complexDisambiguation(a, b)
```
Araq, line 1337: Why is this required? It was more elegant without this logic.

metagn: Without this overloads for `Ordinal` and `enum` are ambiguous. Since `Ordinal` does not match `enum` and `enum` does not match `Ordinal` with this PR, `mutualGenericMatch` doesn't work. `sumGeneric` calculates the value of `Ordinal` as 0 and `enum` as 1 as in the original compiler.  `==`, `<` etc have overloads for `Ordinal` and `enum` each, and this PR also makes `items` for sets have overloads for `Ordinal` and `enum`.

Araq: > Since Ordinal does not match enum

How come? Do you mean enum with holes?

metagn: `T: enum` could be either a holey enum or an ordinal enum, so it does not always satisfy `T: Ordinal`, so it doesn't match. We could make it match but it would fail during instantiation because the overloads are chosen at generic time.

Araq: What if we introduce a dedicated typeclass `enumgap` to system.nim.

metagn: The problem of #952 would still exist, it would probably be a big incompatibility that holey enums which are declared as `enum` do not match `enum`

Araq: An enum without holes should match `enum` and an enum with holes should match the new `enumgap`. Then if both are supported `enum|enumgap` can be used. Ordinal includes `enum` but not `enumgap`. Would this work?

metagn: > An enum without holes should match  enum`

The problem is that this is an incompatibility. And it doesn't read well, holey enums are declared as `type Foo = enum` same as regular enums. It would work otherwise.

Maybe we could have all 3, `enum` matches any enum, `OrdinalEnum` matches only ordinal enums etc. But still it would be good to be able to disambiguate between `T: enum` and `T: Ordinal` because original Nim already does.

Araq: The `sumGeneric` code is fine to have in --compat mode but too ugly to go forward with the language in the long run. Esp since the "enum but it's not ordinal" is a problem in Nim 2 too that we never resolved properly. So I would like to have a real solution here.

metagn: Going to see about splitting `enum` and overloads defined on it then but it might need #931 to work properly (`nifbuilder.addFlags` would break without it but it's unused)

PR #979, opened 2025-04-15
Add memfiles

demotomohiro:


metagn:
Nightlies seem to have failed the builds since https://github.com/nim-lang/nightlies/pull/104, the run for yesterday failed due to atlas having a problem with the `sat` dependency which might have been fixed since by https://github.com/nim-lang/Nim/pull/24884

PR #995, opened 2025-04-17
implement sum of products form for typeclass matching

metagn:
follows up #931

Algorithm seems to work but I did not try every test case (cannot find much on this, [typescript](https://github.com/microsoft/TypeScript/blob/027829fbcd0a84172ab4a36d60a0ac915ca7ea09/src/compiler/checker.ts#L8467) seems to do the distributive law but not de morgan since typescript doesn't have `not` types). Since reordering can get quite slow (reordering `(and (or <M terms>) (or <N terms>)` is `O(MN)`) the typeclass is checked if it is already ordered first.

A problem with performance is the complexity of the produced terms also reflects on the complexity of the type match, `(A or B) and (C or D) and (E or F)` expands to:

```
(or
 (and A C E)
 (and A C F)
 (and A D E)
 (and A D F)
 (and B C E)
 (and B C F)
 (and B D E)
 (and B D F))
```

which performs 24 individual matches for only 6 types in the worst case. `A and SomeInteger` for example matches `A` equally as many times as the number of types in `SomeInteger`. An idea would be to cache the match results but getting a key for the type and looking it up might be slower than just matching.

This PR isn't really a priority, I just wanted to get the implementation out of the way

### DIFF
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -375,6 +375,19 @@ proc matchConstraintSplitOr(m: var Match; f: var Cursor; a: Cursor): bool =
   else:
     result = matchBooleanConstraint(m, f, a)
 
+proc matchesConstraintAux(m: var Match; f: var Cursor; a: Cursor): bool =
+  if a.typeKind in {OrT, AndT, NotT}:
+    # typeclass matching typeclass, might need to be reordered to match properly:
+    if isSumOfProducts(a):
+      result = matchConstraintSplitOr(m, f, a)
+    else:
+      var reorderBuf = createTokenBuf(32)
+      reorderSumOfProducts(reorderBuf, a)
+      var reordered = beginRead(reorderBuf)
+      result = matchConstraintSplitOr(m, f, reordered)
+  else:
+    result = matchBooleanConstraint(m, f, a)
+
 proc matchesConstraint*(m: var Match; f: var Cursor; a: Cursor): bool =
   result = false
   if f.kind == DotToken:
@@ -386,7 +399,7 @@ proc matchesConstraint*(m: var Match; f: var Cursor; a: Cursor): bool =
     if res.decl.symKind == TypevarY:
       var typevar = asTypevar(res.decl)
       return matchesConstraint(m, f, typevar.typ)
-  result = matchConstraintSplitOr(m, f, a)
+  result = matchesConstraintAux(m, f, a)
 
 proc matchesConstraint(m: var Match; f: SymId; a: Cursor): bool =
   let res = tryLoadSym(f)
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -382,6 +382,7 @@ proc matchesConstraintAux(m: var Match; f: var Cursor; a: Cursor): bool =
       result = matchConstraintSplitOr(m, f, a)
     else:
       var reorderBuf = createTokenBuf(32)
+      var a = a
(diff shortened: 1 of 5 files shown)


Araq:
Why is this transformation necessary at all?

metagn:
It has to do with the direct mapping to set algebra:

1. `(A ∪ B) ⊂ (X ∪ Y)` is equivalent to `A ⊂ (X ∪ Y) and B ⊂ (X ∪ Y)`, not `(A ∪ B) ⊂ X or (A ∪ B) ⊂ Y`
2. `(A ∩ B) ⊂ (X ∩ Y)` is equivalent to `(A ∩ B) ⊂ X and (A ∩ B) ⊂ Y`, not `A ⊂ (X ∩ Y) or B ⊂ (X ∩ Y)`

So the order in which we separate the terms of the LHS and RHS of the match are different, specifically we separate the unions on the RHS first, then all unions or intersections on the LHS, then the intersections on the RHS. For this to work, the RHS needs to be in "unions of intersections" form.

Long explanation:

Using the notation `T <: U` for "U matches T": In general `T <: (A and B)` iff `T <: A or T <: B` and `T <: (A or B)` iff `T <: A and T <: B`. But this is not the case when `T` is a boolean typeclass. Assuming `X` and `Y` are unit typeclasses (for example concepts), and assuming `A` and `B` are supposed to be some permutation of them:

`(X or Y) <: (A and B)` <=> `(X or Y) <: A or (X or Y) <: B` - works
`(X and Y) <: (A and B)` <=> `(X and Y) <: A or (X and Y) <: B` - wrong, information was lost. This implies we need to split the `and` part in the match pattern before we split it from the matchee:
`(X and Y) <: (A and B)` <=> `X <: (A and B) and Y <: (A and B)` <=> `(X <: A or X <: B) and (Y <: A or Y <: B)` - correct. Can we do the same if the match pattern is `or`?
`(X or Y) <: (A and B)` <=> `X <: (A and B) or Y <: (A and B)` <=> `(X <: A or X <: B) or (Y <: A or Y <: B)` - still works. So it is safe to split any boolean typeclass on the match pattern first before the `and` type in the matchee.

The conditions change if the matchee is an `or` type:

`(X or Y) <: (A or B)` <=> `(X or Y) <: A and (X or Y) <: B` - works
`(X and Y) <: (A or B)` <=> `(X and Y) <: A and (X and Y) <: B` - does what we expect

Now if we try to apply the same rule used for `and` and try to split the match pattern before the matchee:

`(X and Y) <: (A or B)` <=> `X <: (A or B) and Y <: (A or B)` - equivalent to before
`(X or Y) <: (A or B)` <=> `X <: (A or B) or Y <: (A or B)` <=> `(X <: A and X <: B) or (Y <: A and Y <: B)` - never works

So we need to split the matchee `or` type, then the matcher `and`/`or` types, then the matchee `and` type in that order, as done in #931. For this to work, the matchee typeclass needs to have a top layer of `or` types, then a layer of `and` types, without intertwining. This is the "sum of products" form.

Araq:
I understand what you say but it's not the compiler's job to solve set theory puzzles. This needs to be documented in a new "type theory manual" and we need at least some protection against memory usage blowups (aka "sensible implementation limits") or ideally some other way to restrict the complexity.

metagn:
Typescript limits the number of produced `and` types to 100000 (source file is 3 MB, ctrl+f "checkCrossProductUnion" [here](https://raw.githubusercontent.com/microsoft/TypeScript/refs/heads/main/src/compiler/checker.ts)). Maybe we can limit to 256.

PR #1001, opened 2025-04-19
fixes #1000

### LINKED ISSUE #1000: Floating point number literal `1e17` causes compile error
mytests/test.nim
```nim
discard 1e17
```

```console
bin/nimsem  --isMain m nimcache/tes9hvujc.1.nif nimcache/tes9hvujc.2.nif nimcache/tes9hvujc.2.idx.nif
src/nimony/sem.nim(6802) semcheck
src/nimony/sem.nim(6457) semExpr
src/nimony/sem.nim(4025) semDiscard
src/nimony/sembasics.nim(392) takeParRi
src/nimony/programs.nim(136) error
[Error] mytests/test.nim(1, 1) expected ')', but got: e+17
```

The same error happens with the exponential number larger than 17.
When the exponential number is smaller than 17, compiles without errors.

demotomohiro:


demotomohiro:
Nif spec requires upper case 'E' for floating point numbers.
But if we change the spec to use lower case 'e' instead of 'E', `proc addFloatLit*(b: var Builder; f: float)` can be simplified as it doesn't need to convert lower case 'e' to 'E'.
We can also simplify it without changing Nif spec by changing `formatfloat.addFloat` to output upper case 'E', but it is a proc in Nim 2 stdlib and it is breaking change.

Issue #1003, opened 2025-04-20
Implement table constructors


Araq:
Easy: In the frontend transform `{:}` to `[]` and `{key: val, ...}` to `[(key, val), ...]` and then semcheck the result.

demotomohiro:
If Nifler converts table constructors to `[]` or `[(key, val), ...]`, Nifler never output `tabconstr` node kind.
So `tabconstr` is no longer used?

PR #1005, opened 2025-04-20
implements #1003; transform table constructors to array constructors

demotomohiro:


src/nifler/bridge.nim
```diff
@@ -614,6 +613,16 @@ proc toNif*(n, parent: PNode; c: var TranslationContext; allowEmpty = false) =
       toNif(n[i], n, c)
     c.b.endTree()
     c.depsEnabled = oldDepsEnabled
+  of nkTableConstr:
```
Araq, line 616: Shall be done in sem.nim, not in Nifler.

PR #1014, opened 2025-04-22
implements `getSize` for case objects

### LINKED ISSUE #1010: sizeof calculation for case objects
would go [here](https://github.com/nim-lang/nimony/blob/bc2f91804737f68043c984d5eab057012abb5cd4/src/nimony/sizeof.nim#L109)

ringabout:
fixes #1010

There are some pre-existing issues with `getSize`, which may be fixed in the following PRs. i.e.

```nim
type Foo = object
  x: bool
  z: string
  t: bool
```

With/Without this PR, `getSize` returns `24` for this structure, while `sizeof(Foo)` returns `32` instead

### DIFF
--- a/src/nimony/decls.nim
+++ b/src/nimony/decls.nim
@@ -192,7 +192,7 @@ type ObjFieldIter* = object
 proc initObjFieldIter*(): ObjFieldIter =
   result = ObjFieldIter(nested: 1)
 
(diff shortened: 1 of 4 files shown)


Araq:
Ok, but the most important next step is that nifcgen produces a union for these so that sizeof has any chance of matching.

Issue #1023, opened 2025-04-25
`tests/nimony/stdlib/tparseutils.nim` randomly fails on macos


demotomohiro:
It seems `tests/nimony/stdlib/tparseutils.nim` passes on Windows and Linux but randomly fails on Macos.

PR #1032, opened 2025-04-26
manually insert \0 for parseutils array passed to `strtod`

metagn:
refs #1023

The array is not zeroed out in nimony AFAIK, so a \0 is manually added here. Another option might be to call `zeroMem`. Passed 5 CI runs without failing.


metagn:
CI passed for 5 runs in a row but could still fail

PR #1033, opened 2025-04-27
nifgram: implement STACK\POP operations

ASVIEST:
Now you can write this grammar:
```lisp
(RULE :DataValue TYP (OR STRINGLITERAL INTLIT UINTLIT FLOATLIT))
(RULE :DataKey (POP :TYP) (OR DataValue (times ".rept " INTLIT (DO "nl") DataValue (DO "nl") ".endr")) (DO "nl"))
(RULE :DataDecl SYMBOLDEF ":" (DO "nl") (STACK (ONE_OR_MANY (OR
  (string ".string " DataKey)
  (byte ".byte " DataKey)
  (word ".short " DataKey)
  (long ".long " DataKey)
  (quad ".quad " DataKey)

))) (DO "nl"))
```
And this means that need to pop the DataKey from the stack, write it to the variable TYP, and then add it to stack.

### DIFF
--- a/src/nifgram/nifgram.nim
+++ b/src/nifgram/nifgram.nim
@@ -844,63 +844,6 @@ proc findSccs(c: ScanContext): Table[string, int] =
       s.visit(rule)
   s.rindex
 
-proc sortedRules(c: var ScanContext): seq[string] =
-  # to get available pop vars we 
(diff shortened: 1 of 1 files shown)


Araq:
If you like Nifgram, consider rewriting it to use the newer cursor based APIs.

Issue #896, opened 2025-03-30
implement inheritance, methods and the `of` operator


Araq:
Also requirement for closures.

Plan:

- New node kind: the only-fans (`of`) operator
- New node kind: the procCall override
- `base` pragma that we can just ignore, I think.
- The buildup of vtables should be done in the hexer.

mrgaturus:
i think to make vtable the unique way to method dispatch we need get rid of multi-methods for now

Araq:
Yes, sure.

Araq:
Still to do:

- [ ] Implement it for `ref object`.
- [ ] Add basic declarations to system.nim
- [ ] Implement generic methods.

ringabout:
> inherited fields in object constructors do not work, should be handled by NIFC? (open issue)

I'm thinking of transforming/desugaring `oconstr` or `newobj` into default construction + `dot`/`ddot` statements, since `dot`/`ddot` are already handled properly. 
e.g. `x = Ref(a: 1, b: 2)` into `x = (let tmp = Ref(); tmp.a = 1; tmp.b = 2)`. Though it could be a problem for `oconstr` because of copying

PR #1040, opened 2025-04-29
basic generic/ref inheritance

metagn:
refs #896

Inheritance is implemented in sigmatch for `ref`/`ptr` and generic versions of object types. This is not perfectly implemented for edge cases like "generic objects matching generic objects" but should handle the most common cases fine.

Changes to vtables_backend:

* Upcasts for ref/ptr types are just `cast`s of their pointers, this works since the first field is always the parent object and makes the refcount field for `ref`s work.
* Access to the vtable field needs an upcast to the hierarchy root
* Misc fixes: Temps are fixed (leaves `(expr` open instead of immediately closing, generates `ptr` type for `addr` case), generic types do not generate unused vtables, empty methods array generates `{ NIL }` since it is a flexible array.

Leaving for followup, added to issue todo and as XXX comments:

* Generic methods need to be instantiated for every single instantiation of the attached generic type like hooks are at the end of sem, in the test this is worked around by the methods matching statically for each instance.
* Nil checks are needed for method dispatch and for `of` with pointer type operands. Method dispatch is straightforward but I don't fully understand what `of` does in the old compiler, the codegen implies it should always return `false` when the operand is nil but it behaves like a static match instead. In any case this would probably pollute the diff.

### DIFF
--- a/src/hexer/vtables_backend.nim
+++ b/src/hexer/vtables_backend.nim
@@ -429,39 +429,20 @@ proc trBaseobj(c: var Context; dest: var TokenBuf; nn: var Cursor) =
             tr c, dest, bufn
     skipParRi n
   else:
-    n = nn
     let isPtr = typ.typeKind in {RefT, PtrT}
     if isPtr:
-      if typ.typeKind == RefT:
-        dest.addParLe(CastX, info)
-        dest.addSubtree typ
-      dest.addParLe(AddrX, info)
-    copyInto dest, n:
-      if isPtr:
-        assert n.typeKind == typ.typeKind
-        inc n
-        tr c, dest, n
-        skipParRi n
-      else:
-        tr c, dest, n
+      # to preserve the refcount field position, cast entire pointer
+      # works since the supertype is always the first field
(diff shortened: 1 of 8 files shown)


Araq:
> // how to get b from a? currently generates:
> b = a.Q;

b = &a->Q;

Not sure I understand the question.

metagn:
`&a->Q` is not compatible with `ref_RootObj`

Araq:
True, but `cast` should work as the offsets for the reference counter are identical. It bites with C's strict aliasing rules so I guess we have no choice but to the change the `ref` implementation to have the refcount at a negative offset. This also makes it more compatible with Nim.

But for now try something like `cast[](&a->Q)`

Araq:
btw object conversion for `ref` also need nil checks

metagn:
Left nil checks for followup as well since they will need a disproportionately large diff just to add simple guards to the codegen

Araq:
Fine but `x of FooRef` should be true when `x` is `nil` as `nil` is a valid citizen of FooRef. But probably not worth breaking compat with Nim for this edge case.

PR #1041, opened 2025-04-29
fixes `DconvX` conv, keeps the type as it is

ringabout:


src/nimony/derefs.nim
```diff
@@ -653,7 +653,10 @@ proc tr(c: var Context; n: var Cursor; e: Expects) =
     of ExprX:
       trStmtListExpr c, n, e
     of DconvX:
-      trSons c, n, e
+      takeToken c, n
+      takeTree c.dest, n # takes the type as it is
```
Araq, line 657: As opposed to what? derefs does nothing with types. Or does it?

PR #1042, opened 2025-04-29
instantiate methods for generic types

metagn:
Needs to work across modules


Araq:
Now that `.raises` is a thing we need to check that overriden methods agree on the `.raises` aspect. But should be done in a follow-up PR.

PR #1045, opened 2025-04-29
nifcgen: implements case objects

ringabout:


Araq:
Told you this gets simpler when we extend NIFC to support anon structs and unions within structs.

metagn:
See [here](https://github.com/nim-lang/Nim/pull/24459/files#diff-1208ab8dfbcc04a982aeb83a28655324f8880fa9b756b83db873163dfe1fec6cR693) for my method of generating named unions for the codegen of the old compiler, from what I remember enabling named unions for the normal C generation made CI fail, probably due to a byte layout difference.

PR #1058, opened 2025-05-02
keep order of provided object constructor fields in sem

### LINKED ISSUE #1056: object constructor generation field reordering in sem breaks order of evaluation
i.e.

```nim
type Foo = object
  a, b: int

let x = Foo(
  b: foo()
  a: bar())
```

gets rewritten to:

```nim
Foo(
  a: bar()
  b: foo())
```

in an attempt to set every field. It should probably use the order of the given fields, then set the defaults of the remaining fields (or do this first).

metagn:
closes #1056 (for the sem issue, evaluation order is still not fixed)

The compiler reorders the fields of an object constructor depending on the field order of the object type, this PR makes it so that provided fields are kept in the order that they were provided and only the default fields keep the object type order. This still encounters UB in the generated C so it doesn't fix the evaluation order, but it might still be useful, and things like:

```nim
Foo(
  b: (let x = 123; x),
  a: x)
```

won't break on resem.

### DIFF
--- a/src/hexer/duplifier.nim
+++ b/src/hexer/duplifier.nim
@@ -652,6 +652,9 @@ proc trObjConstr(c: var Context; n: var Cursor; e: Expects) =
       copyInto c.dest, n:
         takeTree c.dest, n
         tr c, n, WantOwner
+        if n.kind != ParRi:
+          # optional inheritance
+          takeTree c.dest, n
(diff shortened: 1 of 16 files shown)


metagn:
Ended up not fully fixing the issue, could close and postpone this for compat mode.

PR #1072, opened 2025-05-07
wip: vtable front checking for effects

ringabout:


Araq:
all you need to do is to make calling conventions and .raises part of the signature description

PR #1073, opened 2025-05-07
fixes #985

### LINKED ISSUE #985: Modules are compiled even if the import statements are under `when false:`
mytests/modfoo.nim:
```nim
invalid code
```

mytests/testwhenimp.nim
```nim
when false:
  import modfoo
```

```console
$ ./bin/nimony c -r mytests/testwhenimp.nim 
bin/nimsem  m nimcache/modv5te5u.1.nif nimcache/modv5te5u.2.nif nimcache/modv5te5u.2.idx.nif
src/nimony/sem.nim(6686) semcheck
src/nimony/sem.nim(6420) semExpr
src/nimony/sem.nim(1759) semCall
src/nimony/sem.nim(6232) semExpr
src/nimony/sem.nim(3741) semExprSym
src/nimony/sembasics.nim(169) buildErr
mytests/modfoo.nim(1, 9) Error: undeclared identifier
undeclared identifier
```

When importing modules contains OS specific code like:
```nim
when defined(windows):
  import windowsMod
elif defined(linux):
  import linuxMod
```
if `windowsMod` is compiled on Linux, it might fail to compile as it can contains windows specific header pragma, compile options or other code compiles only on Windows.

Udate:

There are similar bugs caused with `when false:`.

Following code produces the executable prints "modfoo".
modfoo.nim:
```nim
import std/syncio

echo "modfoo"
```

main.nim:
```nim
when false:
  import modfoo
```

Following code causes the compile error if `foo.c` doesn't exists.
buildmod.nim:
```nim
{.build("C", "foo.c").}
```

main.nim:
```nim
when false:
  import modfoo
```

Nimony runs nimsem and Hexer with modules imported under `when false:`, because Nifler generates `.deps.nif` files from import statements in input `.nim` files but Nifler doesn't evaluate the expressions in when statements, so `.deps.nif` contains modules actually not imported.
So even if modules imported under `when false:` compile without errors, they can cause side effects with Nimony compiler plugins or `staticExec` as they can run arbitrary code.
I talked about it with Araq and we worry about the side effects from plugins and `staticExec` later.
These side effects can always be used to prevent incremental or parallel builds.
Well written plugins affect the module they are compiled on and nothing else.

demotomohiro:
Currently, when a module that has the compile error imported by N modules, that compile error is printed N times.

This fix has minor drawbacks:
- semchecks modules imported under removed when statements branches
- nimsem compiles all modules even if there is compile errors


demotomohiro:
> Currently, when a module that has the compile error imported by N modules, that compile error is printed N times.

It is fixed.
When a module is failed to compile and other modules tried to imported it, the suffix of the imported module is written to `.2.idx.nif` so that main module print the error from the module only once.
And when nimsem imported a module failed to compile, it quits after writing empty `.2.nif` and `.2.idx.nif` file.

src/nimony/sem.nim
```diff
@@ -7096,6 +7096,61 @@ proc writeOutput(c: var SemContext; outfile: string) =
...
+    if c.importedModules.len != 0:
+      let systemSymId = if {SkipSystem, IsSystem} * c.moduleFlags == {}:
+                          c.currentScope.tab[pool.strings.getOrIncl("system")][0].name
+                        else:
+                          SymId(0)
+      deps.buildTree ImportS, NoLineInfo:
+        for symId, i in c.importedModules:
+          if symId != systemSymId:
+            deps.addStrLit i.path.toAbsolutePath
+  let depsFile = changeFileExt(infile, ".deps.nif")
```
Araq, line 7118: No. Do not update the .deps.nif file. Instead generate an error file.

demotomohiro: Errors are written to ".idx.nif" in `quitWithError` proc.
When imported module has errors in `.idx.nif`, the errors are written to `.idx.nif` in `importSingleFile` proc in `src/nimony/semimport.nim`.
Updated `.deps.nif` are used to generate final Makefile to avoid Hexer and later phase reading modules imported under `when false:`.
It should write to a separate file (`.2.deps.nif`) instead of updating existing `.1.deps.nif` files?

Araq: Oh yeah .2.deps.nif makes sense. I think.

demotomohiro: I changed code to write `.2.deps.nif`.

src/nimony/sem.nim
```diff
@@ -7121,6 +7121,55 @@ proc writeOutput(c: var SemContext; outfile: string) =
...
+    errs.buildTree ErrT, NoLineInfo:
+      var nested = 0
+      while true:
+        case n.kind
+        of ParLe:
+          if n.tagId == ErrT:
+            errs.takeTree n
+            if nested == 0:
+              break
+            continue
```
Araq, line 7157: 1. I don't accept break+continue dances. Remove the continue and duplicate the `inc n` step instead where needed.
2. Breaking in the ParLe case when nested == 0 as opposed to the ParRi case is never done elsewhere in the entire codebase so I don't understand it. I also don't want to understand it, rewrite it so that it is obvious what it does.

src/nimony/semimport.nim
```diff
@@ -67,7 +68,36 @@ proc importSingleFile(c: var SemContext; f1: ImportedFilename; origin: string;
   else:
     result = c.processedModules[suffix]
   let module = addr c.importedModules.mgetOrPut(result, ImportedModule(path: f2))
-  loadInterface suffix, module.iface, result, c.importTab, c.converters, c.methods, exports, mode
+  let beforeErrors = c.dest.len
+  loadInterface suffix, module.iface, result, c.importTab, c.converters, c.methods, exports, c.dest, mode
+  if c.dest.len > beforeErrors:
```
Araq, line 73: Use a helper proc for this logic and ideally make `loadInterface` return a success indicator. Move the parsing logic into `loadInterface`.

src/nimony/sem.nim
```diff
@@ -7121,6 +7121,57 @@ proc writeOutput(c: var SemContext; outfile: string) =
       toBuild: move c.toBuild,
       exportBuf: buildIndexExports(c))
 
+  # Update .2.deps.nif file that doesn't contain modules imported under `when false:`
+  # so that Hexer and following phases doesn't read such modules.
+  var deps = createTokenBuf(16)
+  deps.buildTree StmtsS, NoLineInfo:
+    if c.importedModules.len != 0:
+      let systemSymId = if {SkipSystem, IsSystem} * c.moduleFlags == {}:
+                          c.currentScope.tab[pool.strings.getOrIncl("system")][0].name
```
Araq, line 7130: Gross hack. And why does system to be special cased?

src/nimony/programs.nim
```diff
@@ -90,8 +90,34 @@ proc loadInterface*(suffix: string; iface: var Iface;
                     module: SymId; importTab: var OrderedTable[StrId, seq[SymId]];
                     converters, methods: var Table[SymId, seq[SymId]];
                     exports: var seq[(string, ImportFilter)];
-                    filter: ImportFilter) =
+                    errors: var TokenBuf;
+                    processedModules: var Table[string, SymId];
```
Araq, line 94: Why does this need `processedModules`?

Araq:
This is incomprehensible. Please create a new PR that only does this:

- semchecker produces ".2.deps.nif" files.
- Makefile generator uses these.

src/nimony/sem.nim
demotomohiro: In the first commit in this PR, sem.nim didn't exclude system module from `.1.deps.nif` and that caused compile errors as nimsem compiled system module without `--isSystem`.
`src/nimony/deps.nim` supposes system module is not in `.deps.nif` and system module is handled differently from normal modules.

Current nimsem writes `.2.deps.nif` and they are used to generate final Makefile that calls Hexer and later phases that doesn't handle system module differently.
So sem.nim can include system module in `.2.deps.nif` now.

Excluding system module from `.deps.nif` makes file size smaller.

demotomohiro:
> This is incomprehensible. Please create a new PR that only does this:
> 
> * semchecker produces ".2.deps.nif" files.
> * Makefile generator uses these.

Done: https://github.com/nim-lang/nimony/pull/1087

PR #1075, opened 2025-05-08
fixes #1047; makes `callconvs` and `raises` parts of method signature

### LINKED ISSUE #1047: `.raises` and calling convention must be part of the method signature
Now that .raises is a thing we need to check that overriden methods agree on the .raises aspect. And since we are looking at the pragmas we might as well enforce .nimcall for every method, or calling convention consistency.

ringabout:
fixes #1047

```nim
type
  Base = object of RootObj
  Class = object of Base
    id: int

method foo(x: Base) {.raises, nimcall.} =
  discard

method foo(x: Class) {.cdecl.} =
  discard
```

```
foo:SE:nimcall:1
foo:SE:cdecl:0
```

In the `index` file

```lisp
(method
 (kv Base.0.teszpno6r1
  (stmts
   (kv foo.0.teszpno6r1 "foo\3ASE\3Animcall\3A1")))
 (kv Class.0.teszpno6r1
  (stmts
   (kv foo.1.teszpno6r1 "foo\3ASE\3Acdecl\3A0"))))
```

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -3840,8 +3840,8 @@ proc attachMethod(c: var SemContext; symId: SymId;
       var methodName = pool.syms[symId]
       extractBasename methodName
       c.dest.endRead()
-      let pragams = cursorAt(c.dest, beforePragmas)
-      signature = pool.strings.getOrIncl(methodKey(methodName, rest, pragams))
+      let pragmas = cursorAt(c.dest, beforePragmas)
+      signature = pool.strings.getOrIncl(methodKey(methodName, rest, pragmas))
   if root == SymId(0):
     let typ = typeToString(params)
     c.dest.endRead()
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -3821,7 +3821,7 @@ proc attachConverter(c: var SemContext; symId: SymId;
         c.converterIndexMap.add((root, symId))
 
 proc attachMethod(c: var SemContext; symId: SymId;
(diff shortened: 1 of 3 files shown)


src/nimony/sem.nim
```diff
@@ -3839,7 +3839,9 @@ proc attachMethod(c: var SemContext; symId: SymId;
       skipParRi rest
       var methodName = pool.syms[symId]
       extractBasename methodName
-      signature = pool.strings.getOrIncl(methodKey(methodName, rest))
+      c.dest.endRead()
+      let pragams = cursorAt(c.dest, beforePragmas)
+      signature = pool.strings.getOrIncl(methodKey(methodName, rest, pragams))
```
Araq, line 3844:
```suggestion
      let pragmas = cursorAt(c.dest, beforePragmas)
      signature = pool.strings.getOrIncl(methodKey(methodName, rest, pragmas))
```

Araq:
Still much more complex than necessary. We know that after the parameters and the return type the pragmas follow...

ringabout:
fixed

src/nimony/vtables_frontend.nim
```diff
@@ -50,4 +72,6 @@ proc methodKey*(name: string; a: Cursor): string =
   inc a
   # also add return type:
   mangle b, a
-  result = name & ":" & b.extract()
+  inc a
```
Araq, line 75:
```suggestion
  skip a
```

PR #1093, opened 2025-05-14
fixes #985; runs frontend Makefile twice

### LINKED ISSUE #985: Modules are compiled even if the import statements are under `when false:`

demotomohiro:
1st frontend Makefile calls nimsem to produce `.2.deps.nif` files that doesn't contains modules imported under `when false:` and are used by 2nd frontend Makefile.
1st frontend Makefile uses `.1.deps.nif` generated by Nifler contains modules imported under `when false:`. So it can call nimsem for such modules.
2nd frontend Makefile uses `.2.deps.nif` and doesn't process such modules.

1st frontend Makefile ignores errors as errors can come from modules actually not imported and to process all modules used by the project with nimsem.
And nimsem writes dummy empty output files before starting semcheck and they are overwritten only when nimsem completed successfully.

You still get errors from modules imported under `when false:` if it causes error in Nifler.

I think this PR is simpler than https://github.com/nim-lang/nimony/pull/1073 excepts Makefile generation part.


Araq:
Let me try to fix this bug and see what I can come up with. In the meantime, work on the "polish" task.

PR #1105, opened 2025-05-18
prints inner error first

demotomohiro:
When I compiled following code, it printed only type mismatch error, didn't print undeclared identifier.
```nim
proc foo(x: int): int =
  discard

discard foo(foo(undeclaredVar))
```

This PR scan err tree recursively and print the most inner error first.
So when you compile above code, nimsem prints undeclared identifier error first.
I think this PR prints better errors that helps to find the location need to be fixed.

### DIFF
--- a/src/nimony/reporters.nim
+++ b/src/nimony/reporters.nim
@@ -101,6 +101,50 @@ proc infoToStr*(info: PackedLineInfo): string =
     result = pool.files[rawInfo.file].shortenDir()
     result.add "(" & $rawInfo.line & ", " & $(rawInfo.col+1) & ")"
 
+proc reportErrorsImpl(n: var Cursor; r: var Reporter; errTag: TagId): int =
+  assert n.kind == ParLe and n.tagId == errTag
+  result = 1
+  let info = n.info
+  inc n
+  # original expression, optional:
+  if n.kind == DotToken:
+    inc n
+  else:
+    if n.kind == ParLe:
+      if n.tagId == errTag:
+        inc result, reportErrorsImpl(n, r, errTag)
+      else:
+        var nested = 0
+        inc n
+        while true:
+          if n.kind == ParRi:
+            inc n
+            if nested == 0: break
+            dec nested
+          elif n.kind == ParLe:
+            if n.tagId == errTag:
+              inc result, reportErrorsImpl(n, r, errTag)
+            else:
+              inc nested
+              inc n
+          else:
+            inc n
+    else:
+      inc n
+  let doReport = not r.reportedErrSources.containsOrIncl(info)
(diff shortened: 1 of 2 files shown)


Araq:
This will fix one of the biggest annoyances. Thanks!

Araq:
Instead we shouldn't produce nested errors in `sem.nim`. `buildErr` should scan for errors and if there is an error inside the nodes, use that error instead of building one on top of that.

demotomohiro:
> Instead we shouldn't produce nested errors in `sem.nim`. `buildErr` should scan for errors and if there is an error inside the nodes, use that error instead of building one on top of that.

I tried to implement it but it affects `defined` proc.

Following code prints `true` when compiled with `-d:abcdef` before implementing it.
```nim
import std/syncio

let z = defined(abc.def)
echo z
```
But after implementing it, prints `false`.

`semDefined` proc in `src/nimony/sem.nim` calls `getDottedIdent` proc that builds an identifier by recursively scanning `DotX` nodes.
So `getDottedIdent` proc seems need nested errors.

metagn:
`getDottedIdent` skips errors for the original AST in them, maybe AST is being deleted? i.e. it takes in the input:

```
(err
  (dot
    (err abc "undeclared identifier")
    def)
  "undeclared field for type auto")
```

Maybe this can become:

```
(err
  (dot abc def)
  "undeclared identifier")
```

or we can stop trying to reconstruct ast from errors and implement proper untyped arguments for `defined`

demotomohiro:
> `getDottedIdent` skips errors for the original AST in them, maybe AST is being deleted? i.e. it takes in the input:
> 
> ```
> (err
>   (dot
>     (err abc "undeclared identifier")
>     def)
>   "undeclared field for type auto")
> ```
> 
> Maybe this can become:
> 
> ```
> (err
>   (dot abc def)
>   "undeclared identifier")
> ```

It works for `getDottedIdent` proc but `buildErr` proc is called with other errors.

Transforming
```nim
(err
  (call foo.0.tesru68kz
   (err abc "undeclared identifier")) "Type mismatch at [position] ...")
```
to
```nim
(err
  (call foo.0.tesru68kz abc) "undeclared identifier")
```
would makes the error message less readable.

Araq:
As I said, solve the problem at its root cause where the nested error is created. It's not the reporter's job to guess the best error message, it should not be ambiguous.

demotomohiro:
> Instead we shouldn't produce nested errors in `sem.nim`. `buildErr` should scan for errors and if there is an error inside the nodes, use that error instead of building one on top of that.

Implemented it in https://github.com/nim-lang/nimony/pull/1120 .

PR #1106, opened 2025-05-18
accept stdin input in debug mode to resume compilation on errors

metagn:
Threw this together if it's useful, when debug mode encounters an error, instead of quitting, it asks for stdin input such that one can either quit or resume compilation. It asks to type "q" to quit and anything else (including just enter) to resume but this can be changed, also ctrl c works.

If this is sufficient for `defined` etc then we can remove the `c.debugAllowErrors` workaround, there is also another workaround in `semDconv` that adds an error message instead of using `buildErr`.


Araq:
Lol, I cannot imagine getting used to stdin questioning.

Issue #1107, opened 2025-05-18
we need argument dependent lookups


Araq:
This program should work but doesn't:

```nim
import std / syncio

type
  MyS = distinct string

proc write*(f: File; s: MyS) = write f, string(s)

proc main =
  var s90 = MyS"abc"
  echo s90

main()
```

metagn:
I think this specific case is caused by `OchoiceX` not considering introduced symbols but implementing typebound ops is fine as well

Araq:
The typebound ops should generalize the technique we use for concepts which already enrich the scope. If I recall correctly.

PR #1110, opened 2025-05-18
expand `OchoiceX` on resem

metagn:
refs #1107 (makes issue test case work but isn't a substitute for argument dependent lookup in general)

`OchoiceX` when semchecked again (as is done for callee symbols in calls) now looks up the current context again for newly introduced symbols and adds them to the symchoice. This is not done in original Nim: instead of rewriting the symchoice node, open symchoices give new symbols during overload iteration. Doing it this way might unintentionally "accumulate" symbols after consecutive resems but I'm not sure that this would necessarily be a bad thing.

To do it how original Nim does it, we can either only expand the symchoice if a callee semcheck flag is passed, or maybe `resolveOverloads` when iterating over `OchoiceX` can also look up newly introduced symbols. But maybe these are not general enough.

Worth noting that this is the first difference in behavior between `OchoiceX` and `CchoiceX` in the codebase.


Araq:
> makes issue test case work but isn't a substitute for argument dependent lookup in general

We want it in general though. At least as an experiment. Maybe like Nim's typebound ops.

src/nimony/sembasics.nim
```diff
@@ -136,6 +136,22 @@ proc buildSymChoice*(c: var SemContext; dest: var TokenBuf; identifier: StrId; i
     dest.shrink oldLen
     dest.add identToken(identifier, info)
 
+proc addSymChoiceSyms*(c: var SemContext; identifier: StrId; marker: var HashSet[SymId]; info: PackedLineInfo) =
```
Araq:
```suggestion
proc addSymChoiceSyms*(c: var SemContext; dest: var TokenBuf; identifier: StrId; marker: var HashSet[SymId]; info: PackedLineInfo) =
```

src/nimony/sembasics.nim
```diff
@@ -136,6 +136,22 @@ proc buildSymChoice*(c: var SemContext; dest: var TokenBuf; identifier: StrId; i
...
+  # like rawBuildSymChoice but adds to an existing symchoice, ignoring duplicates
+  var it = c.currentScope
+  while it != nil:
+    for sym in it.tab.getOrDefault(identifier):
+      if not marker.containsOrIncl(sym.name):
+        c.dest.addSymUse sym, info
```
Araq:
```suggestion
        dest.addSymUse sym, info
```

src/nimony/sembasics.nim
```diff
@@ -136,6 +136,22 @@ proc buildSymChoice*(c: var SemContext; dest: var TokenBuf; identifier: StrId; i
...
+    it = it.up
+  # mirror considerImportedSymbols:
+  for moduleId in c.importTab.getOrDefault(identifier):
+    # prevent copies
+    let candidates = addr c.importedModules[moduleId].iface[identifier]
+    for defId in candidates[]:
+      if not marker.containsOrIncl(defId):
+        c.dest.add symToken(defId, info)
```
Araq:
```suggestion
        dest.add symToken(defId, info)
```

PR #1120, opened 2025-05-20
don't construct nested errors in buildErr proc

demotomohiro:
Unlike https://github.com/nim-lang/nimony/pull/1105, when `orig` param contains error node in `buildErr` proc, use that error to avoid nest errors.
`define(abc.def)` doesn't work correctly with this PR.

### DIFF
--- a/src/nimony/sembasics.nim
+++ b/src/nimony/sembasics.nim
@@ -180,11 +180,32 @@ proc buildErr*(c: var SemContext; info: PackedLineInfo; msg: string; orig: Curso
       if orig.kind != DotToken:
         echo "Source: ", toString(orig, false)
       quit 1
-  c.dest.buildTree ErrT, info:
-    c.dest.addSubtree orig
-    for instFrom in items(c.instantiatedFrom):
-      c.dest.add dotToken(instFrom)
-    c.dest.add strToken(pool.strings.getOrIncl(msg), info)
+  var n = orig
+  var hasErr = false
+  if n.kind == ParLe:
+    if n.tagId == ErrT:
+      hasErr = true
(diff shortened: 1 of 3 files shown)


src/nimony/sembasics.nim
```diff
@@ -180,11 +180,32 @@ proc buildErr*(c: var SemContext; info: PackedLineInfo; msg: string; orig: Curso
...
+          else:
+            inc nested
+  if hasErr:
+    c.dest.addSubtree n
+  else:
+    c.dest.buildTree ErrT, info:
+      c.dest.addSubtree orig
+      for instFrom in items(c.instantiatedFrom):
+        c.dest.add dotToken(instFrom)
+      c.dest.add strToken(pool.strings.getOrIncl(msg), info)
```
Araq, line 208:
```suggestion
  c.dest.buildTree ErrT, info:
    c.dest.addSubtree orig
    for instFrom in items(c.instantiatedFrom):
      c.dest.add dotToken(instFrom)
    if hasErr:
      inc n
      while n.kind != ParRi: c.dest.takeTree n
    else:
      c.dest.add strToken(pool.strings.getOrIncl(msg), info)
```

This way the instantiation info does not get lost.

demotomohiro: When `hasErr` is true, it adds the content of `n` after current instantiation info. That doesn't match the existing err tag node format.
Is this means when err tag was found in `orig`, `buildErr` proc uses `orig` and the error message in `n` but uses instantiation info in the current `SemContext`?

Araq: That's the idea, yes.

PR #1131, opened 2025-05-23
fixes #1102; assertion fails when `Cursor` went out of valid range

### LINKED ISSUE #1102: Cursor can become invalid without failing assertion
```nim
import ../src/lib/[lineinfos, nifcursors, nifstreams]

var buf = createTokenBuf(1)
buf.add dotToken(NoLineInfo)
var n = beginRead buf
inc n   # n goes out of valid range but assertion doesn't fail
discard n.info  # can dereference invalid n without assertion failed
inc n   # assertion failed
```

`inc` proc in `src/lib/nifcursors.nim` has `assert c.rem > 0` but it doesn't detect Cursor become out of bounds.
So bugs like the one fixed in https://github.com/nim-lang/nimony/pull/1100 are hard to find.

If subtract 1 from `Cursor.rem` when creating `Cursor` from `TokenBuf`, the assertion in `inc` proc can always detect `Cursor` went out of valid range. But it probably requires fixing many lines of existing code to don't fail the assertion.
Another solution is adding `assert c.rem > 0` to `load*(c: Cursor): PackedToken` proc and allow `Cursor` goes out of valid range but don't allow dereference invalid `Cursor`.

demotomohiro:


Araq:
This is getting nowhere. Instead make `load` check for a valid token and keep the incs as they are.

Issue #1066, opened 2025-05-06
polish, polish, polish


Araq:
- [x] Improve error messages. Figure out where line information is off.
- [x] Make the error reporter filter duplicate error messages (these can result from generic instantions).
- [x] Port renderer.nim to Nimony for better error messages.
- [x] Calls to undeclared identifiers can produce confusing error messages.
- [x] Errors listing overloaded routines should list their signatures.

PR #1141, opened 2025-05-27
implements #1066; Errors listing overloaded routines list their signatures

demotomohiro:
It will prints routines signatures properly when renderer with renderNoBody flag is implemented.


demotomohiro:
Updated code prints error message like this:
```
mytests/testmismatch.nim(6, 4) Error: Type mismatch at [position]
(call
 (ochoice foo.0.tesv3m3nb1 foo.1.tesv3m3nb1 foo.2.tesv3m3nb1 foo.3.tesv3m3nb1) +1.0)
[1] too many arguments (declared in mytests/testmismatch.nim(1, 1))
[1] expected: (i -1) but got: (f +64) (declared in mytests/testmismatch.nim(2, 1))
[1] expected: string.0.sysvq0asl but got: (f +64) (declared in mytests/testmismatch.nim(3, 1))
[2] too few arguments (declared in mytests/testmismatch.nim(4, 1))
mytests/testmismatch.nim(10, 3) Error: Type mismatch at [position]
(call bp.0.tesv3m3nb1 +1.0)
[1] expected: (i -1) but got: (f +64) (declared in mytests/testmismatch.nim(9, 5))
```
from following code:
```nim
proc foo() = discard
proc foo(x: int) = discard
template foo(x: string): untyped = discard
proc foo[T](x, y: T) = discard

foo(1.0)

proc bar(x: int) = discard
let bp = bar
bp(1.0)
```

Issue #1129, opened 2025-05-23
good error message for the missing closure support


Araq:
As long as closures are not supported, at least produce a decent error message for code like:

```nim
import std/syncio

proc outer =
  var x = 1
  proc inner =
    echo x
  inner()

outer()
```

PR #1150, opened 2025-05-31
implements #1129; detects local variables used in inner procs

demotomohiro:
This PR doesn't works with templates as it requires more data structure changes to know which proc declares the sym.


Araq:
Just add the check to derefs.nim. Add some scope kind information to typenav.TypeScope to detect cross proc usages easily in getLocalInfo

Araq:
Did it myself.

PR #1156, opened 2025-06-02
fixes #1155

### LINKED ISSUE #1155: prints "closures are not supported" error even if closures are not required
Following code doesn't requires closures but print the error.
```nim
var x = 0
proc foo =
  proc bar =
    discard x
```

```
$ ./bin/nimony c mytests/test4.nim 
mytests/test4.nim(4, 13) Error: cannot access local variable `x` from another routine; closures are not supported
```

demotomohiro:
`res.crossedProc` is always false when `res.kind` is neither `VarY` nor `LetY` so that nest procs using only global variables doesn't cause the error.

### DIFF
--- a/src/nimony/typenav.nim
+++ b/src/nimony/typenav.nim
@@ -98,7 +98,7 @@ proc getLocalInfo*(c: var TypeCache; s: SymId): LocalInfo =
     if it.kind == compareTo:
       crossedProc = true
     if res.kind != NoSym:
-      res.crossedProc = crossedProc
+      res.crossedProc = crossedProc and res.kind in {VarY, LetY}
       return res
     it = it.parent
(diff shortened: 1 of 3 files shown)


src/nimony/typenav.nim
```diff
@@ -98,7 +98,7 @@ proc getLocalInfo*(c: var TypeCache; s: SymId): LocalInfo =
     if it.kind == compareTo:
       crossedProc = true
     if res.kind != NoSym:
-      res.crossedProc = crossedProc
+      res.crossedProc = crossedProc and res.kind in {VarY, LetY}
```
Araq:
```suggestion
      res.crossedProc = crossedProc and res.kind in {VarY, LetY, ParamY, ResultY}
```

PR #1159, opened 2025-06-03
adds `[]` proc with `BackwardsIndex` and `pop` proc to TokenBuf

demotomohiro:
These procs makes handling a last element of `TokenBuf` easier.
If you like them, I will change existing code to use them.


Araq:
No, I don't like them as there is nothing about `buf[^1]` that makes sense in general. A token buffer is not a stack.

Issue #1162, opened 2025-06-04
Make index extraction from a NIF file automatic


Araq:
Nimony needs to be able to import a NIF file as it's generated by Nim 2. Nim 2 does not generate an index file, Nimony should compute one. This means we need to extract the logic for the indexing of hooks, methods and converters.

PR #1169, opened 2025-06-06
implements #1162

demotomohiro:


demotomohiro:
v2 should uses `getClass` proc and `nominalRoot` proc in `nimony/typeprops.nim` so that converters and methods using types in other modules works.
But they calls `program.tryLoadSym` proc that doesn't works with Nif files generated by `nim nif` command as they have different file extension.
This PR doesn't uses these procs and it works only with converters and methods doesn't have types in other modules.

Probably need to allow changing file extension used in `suffixToNif` proc in `nimony/programs.nim`.

Araq:
> Probably need to allow changing file extension used in suffixToNif proc in nimony/programs.nim.

Yes, I'm changing this logic as we speak. The idea is that nimcache/v2 contains the new notion of a "bundle". A bundle is a directory of NIF files that can reference each other.

demotomohiro:
> Logic moved into src/nimony/indexgen.nim so please update.

Done.

Parse `compile` pragmas in Nim2 NIF files and add them to the index file with `BuildIdx` tag?

Araq:
> Parse compile pragmas in Nim2 NIF files and add them to the index file with BuildIdx tag?

Yes, but also if you can manage, move the build instructions from the index file to the .deps file. Reading all the index files just for the rare build/compile pragmas is a stupid idea.

src/nimony/indexgen.nim
```diff
@@ -40,35 +41,105 @@ proc indexFromNif*(infile: string) =
...
-          assert param.substructureKind == ParamsU
-          inc param
-          assert param.substructureKind == ParamU
-          let typ = takeLocal(param, SkipExclBody).typ.skipModifier
-          # this assertion fails when got generics proc as generics parameters are not supported yet.
-          if typ.kind == Symbol:
-            let obj = typ.symId
-            let isGeneric = routine.typevars.substructureKind == TypevarsU
-            hookIndexLog[op].add HookIndexEntry(typ: obj, hook: symId, isGeneric: isGeneric)
+          if param.substructureKind == ParamsU:
```
Araq: Use a helper proc for the `MethodS` case.

src/nimony/indexgen.nim
```diff
@@ -40,35 +41,105 @@ proc indexFromNif*(infile: string) =
...
+          inc n
+        inc n
       else:
         skip n
     else:
       skip n
 
   endRead buf
 
+  var exportBuf = default(TokenBuf)
```
Araq: Use a helper proc for this logic.

src/nimony/indexgen.nim
demotomohiro: Done.

src/nimony/indexgen.nim
demotomohiro: Done.

PR #1181, opened 2025-06-11
fixes is operators for tuples

ringabout:
```nim
proc `is`*[T, S](x: T, y: S): bool {.magic: "Is", noSideEffect.}
```

### DIFF
--- a/src/nimony/decls.nim
+++ b/src/nimony/decls.nim
@@ -267,6 +267,9 @@ proc getTupleFieldType*(c: Cursor): Cursor =
   else:
     result = c
 
+  if result.typeKind == TypedescT:
+    inc result
+
 type
   ForStmt* = object
     kind*: StmtKind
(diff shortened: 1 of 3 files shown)


src/nimony/decls.nim
```diff
@@ -267,6 +267,9 @@ proc getTupleFieldType*(c: Cursor): Cursor =
   else:
     result = c
 
+  if result.typeKind == TypedescT:
+    inc result
```
Araq, line 271: It's not getTupleFieldType's job to skip over typedescs. Make the one callsite skip it where it can really come up.

src/nimony/semtypes.nim
```diff
@@ -623,6 +626,8 @@ proc semLocalTypeImpl(c: var SemContext; n: var Cursor; context: TypeDeclContext
         skipParRi n
       elif xkind == TupX:
         semTupleType c, n
+      elif xkind == TupConstrX:
+        semTupleType(c, n, isTupConstr = true)
```
Araq, line 630: I think you should instead give the `is` operator more special logic rather than infesting `semTupleType` with is-operator related logic.

ringabout:
Succeeded by https://github.com/nim-lang/nimony/pull/1181

PR #1183, opened 2025-06-11
fixes #1176; when remove build artifacts, remove cached timestamp

### LINKED ISSUE #1176: Nimony doesn't recompile after changing the source code
Compile code like this:
```nim
import std/syncio

echo "foo"
```
Compile it with nimony and run.
Then edit the source code to output a different message.
Compile it again but compiled code still prints old message, doesn't print new one.

Compiling with `-f` option works correctly.

demotomohiro:
Timestamps of output files are cached before they are updated.
Even after output files are updated, nodes that uses them as input uses cached old timestamp in `needsRebuild`.


Araq:
Awesome work!

PR #1188, opened 2025-06-13
parses compile pragmas in Nim 2

demotomohiro:
Parses compile pragmas in Nim 2 and write the build instructions to `.deps` file.
Compile pragmas that takes output file name like `{.compile: ("file.c", "$1.o").}` is not supported as `build` tag doesn't have an output file name field.


Araq:
Did it differently.

PR #1194, opened 2025-06-15
implements const set union

demotomohiro:
This is required to compile `src/lib/nifreader.nim` with Nimony.


src/nimony/expreval.nim
```diff
@@ -217,6 +217,28 @@ template evalFloatBinOp(c: var EvalContext; n: var Cursor; opr: untyped) {.dirty
   else:
     cannotEval orig
 
+proc evalPlusSet(c: var EvalContext; n: var Cursor): Cursor =
```
Araq: Produces a wrong representation for `{'a'} + {'a'}`. This will matter once we support eval for `card`.

PR #1210, opened 2025-06-19
fixes #1203; if expressions contains return causes error in Hexer

### LINKED ISSUE #1203: if expressions contains return causes error in Hexer
Following code compiles without error with Nim 2, but fails to compile with Nimony:
```nim
proc foo =
  var x = if true:
      1
    else:
      return
```

```
$ ./bin/nimony c mytests/test2.nim 
[Error] type expected but got: 6,5,mytests/test2.nim(ret .)
src/hexer/hexer.nim(99) hexer
src/hexer/hexer.nim(96) handleCmdLine
src/hexer/nifcgen.nim(1897) expand
src/hexer/nifcgen.nim(1703) trStmt
src/hexer/nifcgen.nim(861) trProc
src/hexer/nifcgen.nim(736) trProcBody
src/hexer/nifcgen.nim(1693) trStmt
src/hexer/nifcgen.nim(1485) trIf
src/hexer/nifcgen.nim(1631) trStmt
src/hexer/nifcgen.nim(1631) trStmt
src/hexer/nifcgen.nim(1631) trStmt
src/hexer/nifcgen.nim(1678) trStmt
src/hexer/nifcgen.nim(1342) trExpr
src/hexer/nifcgen.nim(446) trType
src/hexer/hexer_context.nim(80) error
```

ringabout:
fixes #1203

If it is an `ExprX` ending with a `noreturn` statement, we should leave it alone


src/hexer/xelim.nim
```diff
@@ -92,9 +92,14 @@ proc trExprInto(c: var Context; dest: var TokenBuf; n: var Cursor; v: SymId) =
   trExpr c, dest, n, tar
 
   let info = n.info
-  copyIntoKind dest, AsgnS, info:
-    dest.addSymUse v, info
+
+  var rest = beginRead(tar.t)
+  if isNoReturn(rest):
```
Araq, line 97: Makes no sense to me. It's xelim's job to eradicate `ExprX` nodes. `isNoReturn` doesn't matter for this.

ringabout:
```nim
proc foo =
  var x = if true:
      1
    else:
      return
```

xelim insert an assignment for `return` here

```
var tmp: int
if true:
  (asgn tmp 1)
else:
  (asgn tmp (ret))
```

Shouldn't we leave it alone for the `noreturn` branch? i.e.




```nim
var tmp: int
if true:
  (asgn tmp 1)
else:
  (ret)
```

Araq: correct but shouldn't the criterion be "expression has type void"?

ringabout: That would do too. Since it only needs to check one expression, a `noreturn` test should suffice. I will try` getType` instead.

PR #1212, opened 2025-06-19
fixes #1209

### LINKED ISSUE #1209: Returning empty set causes compile error when nimsem compiled with `-d:debug`
nimsem compiles following code without errors when `-d:debug` is not defined but fails when nimsem is compiled with `-d:debug`.
```nim
proc foo(): set[char] = {}
```

```
$ ./bin/nimony c -r mytests/test3.nim 
src/nimony/nimsem.nim(133) nimsem
src/nimony/nimsem.nim(126) handleCmdLine
src/nimony/nimsem.nim(55) singleModule
src/nimony/sem.nim(4862) semcheck
src/nimony/sem.nim(4810) semcheckCore
src/nimony/sem.nim(641) semStmt
src/nimony/sem.nim(4321) semExpr
src/nimony/semdecls.nim(585) semProc
src/nimony/semdecls.nim(15) semProcBody
src/nimony/sem.nim(4501) semExpr
src/nimony/sem.nim(2633) semCurly
src/nimony/sembasics.nim(218) buildErr
src/nimony/sembasics.nim(175) buildErr
mytests/test3.nim(1, 25) Error: empty set needs a specified type
```

`default` template added to system module in https://github.com/nim-lang/nimony/pull/1205 causes compile error when `-d:debug` is defined.

demotomohiro:
`semProcBody` proc in `nimony/semdecls.nim` calls `semExpr` twice for the last expression of the proc body.
If the last expression was `[]` or `{}` and `semExpr` was called without `AllowEmpty`, `semBracket` or `semCurly` calls `buildErr` proc.
If nimsem was built with `-d:debug`, `buildErr` prints given error and quit.
If it was build without `-d:debug`, `buildErr` outputs `ErrT` node and doesnt quit. Then `semProcBody` removes that `ErrT` node and no compiler error is reported.
That is why the compile error was generated only when `-d:debug` is defined.

This PR calls `semExpr` with `AllowEmpty` in `semProcBody` so that `semBracket` or `semCurly` doesn't calls `buildErr` when they got `[]` or `{}`.

### DIFF
--- a/src/nimony/semdecls.nim
+++ b/src/nimony/semdecls.nim
@@ -12,7 +12,7 @@ proc semProcBody(c: var SemContext; itB: var Item) =
       beforeLastSon = c.dest.len
       lastSonInfo = it.n.info
       beforeLastSonCursor = it.n
-      semExpr c, it
+      semExpr c, it, {AllowEmpty}
   if c.routine.kind == TemplateY:
     case c.routine.returnType.typeKind
     of UntypedT:


ringabout:
```nim
proc foo(): int = {}

discard foo()
```
A related case is fixed in https://github.com/nim-lang/nimony/pull/1214

PR #1224, opened 2025-06-24
implements `wasMoved` for trivial types

ringabout:
Implements assigning defaults to most trivial types. I thought of using the `default` call, but it seems to be generics+overloads, I'm not sure whether I can call it in the hexer. Without `zeromem`, `move` and `=wasmoved` calls don't work properly for trivial types


Araq:
The design actually leaves the wasMoved state for trivial types undefined in order to gain efficiency. I don't like this change.

Issue #1231, opened 2025-06-26
Unable to pass anonymous procedure to function


mostik:
```nim
# main.nim
# import std/syncio, std/system

proc sort[T](s: var seq[T], f: proc( a: T, b: T ): bool ) =
  for j in 0..high(s):
    for i in 0..high(s)-1:
      if f( s[i], s[i+1]):
        swap s[i], s[i+1]

var arr = @[ "a", "f", "d", "c", "b", "e" ]  

# works in both nim and nimony
proc f( a: string, b: string ): bool =
  return a < b

sort( arr, f )

# works only in nim
let f2 = proc( a: string, b: string ): bool = a < b 
sort( arr, f2 )

# works only in nim 
sort( arr, proc( a: string, b: string ): bool = a < b )

# for item in arr:
#   echo item
```
```bash
#nimony c -r main.nim
/nimony/src/nifler/nifler.nim(89) nifler
/nimony/src/nifler/nifler.nim(74) handleCmdLine
/nimony/src/nifler/bridge.nim(728) parseFile
/nimony/src/nifler/bridge.nim(695) moduleToIr
/nimony/src/nifler/bridge.nim(675) toNif
/nimony/src/nifler/bridge.nim(302) toNif
/nimony/src/nifler/bridge.nim(349) toNif
/nimony/src/nifler/bridge.nim(675) toNif
/nimony/src/nifler/bridge.nim(206) toNif
/.choosenim/toolchains/nim-#devel/lib/std/assertions.nim(41) failedAssertImpl
/.choosenim/toolchains/nim-#devel/lib/std/assertions.nim(34) raiseAssert
Error: unhandled exception: /nimony/src/nifler/bridge.nim(206, 5) `allowEmpty` unexpected nkEmpty [AssertionDefect]
FAILURE: /nimony/bin/nifler --portablePaths --deps parse main.nim nimcache/mai3r4ykk.1.nif
```

Araq:
@ringabout the fact that we skip to the params for the proc's type is a design bug. We should instead treat the entire ProcS etc as the type. This way we also can distinguish between iterator and proc types.

ringabout:
It seems to include statements as a part of proc types?

```nim
 (glet :x.0.tesp9owq11 . . ~4,~3
  (proc 5 :main.0.tesp9owq11 . . . 9
   (params 1
    (param :a.0 . . 3
     (i -1) .)) . . . 20
   (stmts 2,116,lib/std/syncio.nim
[...]

PR #1235, opened 2025-06-27
turns anonymous procs into normal procs

ringabout:
ref #1231

```nim
let s = proc (name: string): string = "Hello "
```

It becomes

```
 (glet :s.0.teskvlx9c . . 9
  (params 1
   (param name . . 6 string .)) 9 `anonproc.0.teskvlx9c) 4,9

 (proc :`anonproc.0.teskvlx9c . . .
  (params 1
   (param :name.0 . . 6 string.0.sysvq0asl .)) 16 string.0.sysvq0asl . . 25
  (stmts
   (result :result.4 . . ~18 string.0.sysvq0asl .)
   (result :result.0 . . ~18 string.0.sysvq0asl .)
   (asgn result.0 "Hello ") ~25
   (ret result.0) ~25
   (ret result.4))) 14,9
```

### DIFF
--- a/src/nimony/semdecls.nim
+++ b/src/nimony/semdecls.nim
@@ -645,9 +645,9 @@ proc semProcImpl(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind;
 
 proc semProc(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind) =
   var n = it.n
-  var typ = asRoutine(n).params
   inc n
   if n.kind == DotToken:
+    var typ = asRoutine(it.n).params
     let name = identToSym(c, "`anonproc", ProcY)
 
     swap c.dest, c.pending
--- a/src/nimony/semdecls.nim
+++ b/src/nimony/semdecls.nim
@@ -487,12 +487,21 @@ proc hookThatShouldBeMethod(c: var SemContext; hk: HookKind; beforeParams: int):
   else:
     result = false
 
-proc semProc(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind) =
+proc semProcImpl(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind; newName = NoSymId) =
   let info = it.n.info
   let declStart = c.dest.len
   takeToken c, it.n
   let beforeName = c.dest.len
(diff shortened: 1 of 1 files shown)


src/nimony/semdecls.nim
```diff
@@ -634,6 +643,24 @@ proc semProc(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind) =
   producesVoid c, info, it.typ
   publish c, symId, declStart
 
+proc semProc(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind) =
+  var n = it.n
+  var typ = asRoutine(n).params
```
Araq: With a bit of creativity you can move this statement into the uncommon `n.kind == DotToken` case.

src/nimony/semdecls.nim
```diff
@@ -634,6 +643,24 @@ proc semProc(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind) =
...
+  inc n
+  if n.kind == DotToken:
+    let name = identToSym(c, "`anonproc", ProcY)
+
+    swap c.dest, c.pending
```
Araq: Seems dangerous to use `c.pending` for this, better use a new buffer here.

src/nimony/semdata.nim
```diff
@@ -113,6 +113,7 @@ type
     unoverloadableMagics*: HashSet[StrId]
     debugAllowErrors*: bool
     pending*: TokenBuf
+    anons*: TokenBuf
```
Araq: Not the point. Use a local variable instead!

src/nimony/semdecls.nim
```diff
@@ -634,6 +643,25 @@ proc semProc(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind) =
   producesVoid c, info, it.typ
   publish c, symId, declStart
 
+proc semProc(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind) =
+  var n = it.n
+  inc n
+  if n.kind == DotToken:
```
Araq:
```suggestion
  if it.n.firstSon.kind == DotToken:
    # anon routine
```

PR #1242, opened 2025-06-30
evaluate `sizeof` at compile time if possible

demotomohiro:
I think `sizeof` need to be evaluated in nimsem to test `getSize` proc in `nimony/sizeof.nim`.


Araq:
desugar.nim comes to mind. But you shouldn't work on this but on `.union`!

demotomohiro:
I'm working on `.union` pragma: https://github.com/nim-lang/nimony/pull/1244
I recently added code to nimony/sizeof.nim and I thought evaluating `sizeof` using `getSize` proc is the easiest way to test it.

PR #1245, opened 2025-07-01
fixes #1168; generates `haddr` for `var T` destroy hooks

### LINKED ISSUE #1168: Defining `=destroy` and `=dup` generates invalid C code
I got compile error from backend C compiler when compiled following code:
```nim
type
  Foo = object
proc `=destroy`(x: var Foo) = discard
proc `=dup`(x: Foo): Foo = x
```

ringabout:
fixes #1168

### DIFF
--- a/src/hexer/duplifier.nim
+++ b/src/hexer/duplifier.nim
@@ -270,7 +270,7 @@ proc evalLeftHandSide(c: var Context; le: var Cursor): TokenBuf =
       copyIntoSymUse result, tmp, info
     c.typeCache.registerLocalPtrOf(tmp, VarY, typ)
 
-proc callDestroy(c: var Context; destroyProc: SymId; arg: var TokenBuf) =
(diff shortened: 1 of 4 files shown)


src/hexer/duplifier.nim
```diff
@@ -270,11 +270,15 @@ proc evalLeftHandSide(c: var Context; le: var Cursor): TokenBuf =
       copyIntoSymUse result, tmp, info
     c.typeCache.registerLocalPtrOf(tmp, VarY, typ)
 
-proc callDestroy(c: var Context; destroyProc: SymId; arg: TokenBuf) =
+proc callDestroy(c: var Context; destroyProc: SymId; arg: var TokenBuf) =
```
Araq: The `var` here is still unnecessary as far as I can tell.

Araq:
```suggestion
proc callDestroy(c: var Context; destroyProc: SymId; arg: TokenBuf) =
```

PR #1261, opened 2025-07-04
fixes #1255

### LINKED ISSUE #1255: bad error message for != when there is no matching ==
Like in:

```nim

import std/[assertions, syncio]

type
  Obj = object
    a: int

var x: Obj

proc foo =
  while x.a != nil:
    discard

foo()

```

demotomohiro:
Don't remove tokens generated by `semExpr` even if there is an error.


src/nimony/sem.nim
```diff
@@ -529,12 +529,10 @@ proc fetchSym(c: var SemContext; s: SymId): Sym =
     result = Sym(kind: NoSym, name: s, pos: InvalidPos)
 
 proc semBoolExpr(c: var SemContext; n: var Cursor) =
-  let start = c.dest.len
   var it = Item(n: n, typ: c.types.autoType)
   semExpr c, it
   let t = skipModifier(it.typ)
   if classifyType(c, t) != BoolT:
-    c.dest.shrink start
```
Araq, line 538: But this breaks the AST's structure then and will cause bugs everywhere else. For instance the code assumes a `WhileS` has 2 children.

demotomohiro: New commit puts the node generated by `semExpr` and the ErrT node generated by `buildErr` in one Stmts.
So `semBoolExpr` generates only one node and no ErrT nodes are removed.

Araq:
I think a more general `combineErrors` mechanism will serve us better than this special case hack.

Issue #1263, opened 2025-07-04
implement the `using` statement


Araq:
- The implementation strategy from Nim 2.0 can be used.
- This is purely a semantic checker extension, not much complexity should be introduced.

PR #1271, opened 2025-07-06
implements #1263; implements the using statement

demotomohiro:


Araq:
Add a test case that shows procs using parameters that use the `using` statement can be called without problems from a different module.

demotomohiro:
> Add a test case that shows procs using parameters that use the `using` statement can be called without problems from a different module.

Done

PR #1281, opened 2025-07-08
prints errors when object type only pragmas are used incorrectly

demotomohiro:
prints errors when pragmas that are allowed only on object types used with types other than object types.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1378,7 +1378,7 @@ proc semPragma(c: var SemContext; n: var Cursor; crucial: var CrucialPragma; kin
     if kind != TypeY:
       buildErr c, n.info, $pk & " pragma is only allowed on types"
       hasErr = true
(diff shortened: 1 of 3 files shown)


src/nimony/sem.nim
```diff
@@ -1374,13 +1374,22 @@ proc semPragma(c: var SemContext; n: var Cursor; crucial: var CrucialPragma; kin
     c.dest.addParRi()
     inc n
   of ViewP, InheritableP, PureP, FinalP, PackedP, UnionP:
-    if kind == TypeY:
+    var hasErr = false
+    if kind != TypeY:
+      buildErr c, n.info, $pk & " pragma is only allowed on types"
+      hasErr = true
+    elif pk in {InheritableP, FinalP, PackedP, UnionP}:
```
Araq: `ViewP` is also only valid on object types.

PR #1291, opened 2025-07-12
implements rewrite rule: proc pragmas can also be macro invocations

demotomohiro:
ref https://github.com/nim-lang/nimony/issues/524

### DIFF
--- a/src/nimony/semdecls.nim
+++ b/src/nimony/semdecls.nim
@@ -662,7 +662,7 @@ proc semProcImpl(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind;
 proc findMacroInvocs(c: SemContext; n: Cursor; kind: SymKind): seq[Cursor] =
   # find all macro/template identifiers in pragmas to invoke them with parent proc definition
   result = newSeq[Cursor]()
-  if kind in {ProcY, FuncY}:
+  if kind in RoutineKinds:
     var n = asRoutine(n).pragmas
     if n.substructureKind == PragmasU:
       inc n
--- a/src/nimony/semdecls.nim
+++ b/src/nimony/semdecls.nim
@@ -690,9 +690,9 @@ proc transformMacroInvoc(c: var SemContext; it: var Item; macroInvocsPos: seq[Cu
 
   # adds last one in macroInvocsPos to buf first as it is invoked last.
   let info = it.n.info
-  for i in 0 ..< macroInvocsPos.len:
+  for i in countdown(macroInvocsPos.len - 1, 0):
     inBuf.addParLe CallX, info
-    var n = macroInvocsPos[^(i + 1)]
+    var n = macroInvocsPos[i]
     let isCall = n.exprKind == CallX
     if isCall:
       inc n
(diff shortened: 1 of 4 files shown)


Araq:
Use helper procs, avoid `block` and in general use the coding style the rest of the compiler uses.

src/nimony/semdecls.nim
```diff
@@ -654,8 +654,119 @@ proc semProcImpl(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind;
...
+
+  # adds last one in macroInvocsPos to buf first as it is invoked last.
+  let info = it.n.info
+  var offset = 0
+  if macroInvocsPos.len mod 2 == 1:
+    addCall inBuf, inBufReadPos, 1, macroInvocsPos, info
+    offset = 1
+  for i in 0 ..< (macroInvocsPos.len div 2):
+    addCall outBuf, outBufReadPos, i * 2 + 1 + offset, macroInvocsPos, info
+    addCall  inBuf,  inBufReadPos, i * 2 + 2 + offset, macroInvocsPos, info
```
Araq: What is this? Rewrite this, all you need to do is to create a call expression, `templateHere: proc ...` that you semcheck. What problem does this complex logic solve? Why are there outBuf and inBuf that both get some calls?

demotomohiro: Sorry, I misunderstand about how templates work.
I rewrote it to simpler code.

src/nimony/semdecls.nim
```diff
@@ -654,8 +654,94 @@ proc semProcImpl(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind;
...
+
+proc transformMacroInvoc(c: var SemContext; it: var Item; macroInvocsPos: seq[Cursor]) =
+  # transform `proc foo() {.macrofoo, macrobar.}` to `macrobar: macrofoo: proc foo()`.
+  var inBuf = createTokenBuf()
+
+  # adds last one in macroInvocsPos to buf first as it is invoked last.
+  let info = it.n.info
+  for i in 0 ..< macroInvocsPos.len:
+    inBuf.addParLe CallX, info
+    var n = macroInvocsPos[^(i + 1)]
```
Araq: Use `countdown` instead?

src/nimony/semdecls.nim
```diff
@@ -654,8 +654,94 @@ proc semProcImpl(c: var SemContext; it: var Item; kind: SymKind; pass: PassKind;
     producesVoid c, info, it.typ
   publish c, symId, declStart
 
+proc findMacroInvocs(c: SemContext; n: Cursor; kind: SymKind): seq[Cursor] =
+  # find all macro/template identifiers in pragmas to invoke them with parent proc definition
+  result = newSeq[Cursor]()
+  if kind in {ProcY, FuncY}:
```
Araq: This should be done for all routine kinds, I think. But maybe not for templates, check what Nim 2.0 does.

src/nimony/semdecls.nim
demotomohiro: I rewrote it to use countdown.

src/nimony/semdecls.nim
demotomohiro: I checked Nim 2. All routine kinds can have macro pragmas.

PR #1303, opened 2025-07-16
Porting unicode.nim to Nimony

ringabout:


ringabout:
I have kept most of the code of `unicode.nim`. There is still a problem to be solved i.e. https://github.com/nim-lang/nimony/issues/1307#issuecomment-3089766830

```nim
proc split*(s: openArray[char], seps: openArray[Rune]#[ = unicodeSpaces]#, maxsplit: int = -1):
    seq[string] {.noSideEffect.} =
```

So tests relies on `split` are disabled for now

Araq:
So is this ready to go?

ringabout:
I think so. There are still bugs left, but the API is correct

PR #1317, opened 2025-07-22
makes push/pop pragma works with variables

demotomohiro:
nimsem semchecks procedures in `SemcheckSignatures` and `SemcheckBodies` phases, but semchecks variables only in `SemcheckBodies` phase.
So makes push/pop pragma processed in both `SemcheckSignatures` and `SemcheckBodies` phases so that pushed pragmas are applied to both procedures and variables.
But pushed pragmas are added to procedures twice.


Araq:
> But pushed pragmas are added to procedures twice.

Well that's not good, is it?

PR #1319, opened 2025-07-23
fixes #131; array construct can't infer from openarray types

ringabout:
fixes #131

required by https://github.com/nim-lang/nimony/pull/1303


Araq:
Arbitrary logic, `openArray` is not builtin anymore. Can the converter be used instead for this inference?

PR #1324, opened 2025-07-24
adds procs and consts in math in Nim 2

demotomohiro:


lib/std/math.nim
```diff
@@ -1,18 +1,61 @@
...
+                                             ## for Nim's `float` type.
+  MinFloatNormal* = 2.225073858507201e-308   ## Smallest normal number for Nim's
+                                             ## `float` type (= 2^-1022).
 
+when defined(posix) and not defined(genode) and not defined(macosx):
+  {.passL: "-lm".}
+
+# tgmath.h provides type-generic macros and they reduces a number of C functions to import.
+# But it requires C compiler supports C99.
+const CMathHeader = "<tgmath.h>"
```
Araq, line 41: Remove this type-generic stuff in a follow-up PR, it's hard for a native NIFC code generator to support.

PR #1333, opened 2025-07-26
adds `+=`, `-=` and `*=` for ints and floats

demotomohiro:
They are frequently used in math module.


lib/std/system/arithmetics.nim
```diff
@@ -174,3 +174,12 @@ proc `+`*(x, y: float): float {.magic: "AddF64", noSideEffect.}
...
 proc `/`*(x, y: float): float {.magic: "DivF64", noSideEffect.}
+
+template `+=`*[T: SomeNumber](x: var T; y: T) {.untyped.} =
+  x = x + y
+
+template `-=`*[T: SomeNumber](x: var T; y: T) {.untyped.} =
+  x = x - y
+
+template `*=`*[T: SomeNumber](x: var T; y: T) {.untyped.} =
+  x = x * y
```
Araq, line 185: These need to be inline procs so that `x` is evaluated only once.

PR #1335, opened 2025-07-27
fixes #1334

### LINKED ISSUE #1334: Assigning a value to `openArray` causes compile error
Example code:
```nim
proc f(x: var int; y: int) =
  x = y

proc test(x: openArray[int]) =
  for i in 1 ..< x.len:
    x[i] = 1
    #f(x[i], 1) # assigning a value using a proc with var param compiles.
```

error message:
```
lib/std/system/openarrays.nim(11, 4) Error: cannot mutate expression x[i]
```

demotomohiro:


Araq:
No. Instead `f(x[i], 1)` should not compile either.

PR #1364, opened 2025-08-05
fixes #1363; collapses errors into one tree using `buildErr`

### LINKED ISSUE #1363: Declaring the same name const or types twice creates invalid AST
Example code to reproduce:
```nim
const
  x = 1
  x = 1
```

Error message:
```
src/nimony/nimsem.nim(133) nimsem
src/nimony/nimsem.nim(126) handleCmdLine
src/nimony/nimsem.nim(55) singleModule
src/nimony/sem.nim(5184) semcheck
src/nimony/sem.nim(5132) semcheckCore
src/nimony/sem.nim(645) semStmt
src/nimony/sem.nim(4673) semExpr
src/nimony/semdecls.nim(142) semLocal
src/nimony/sembasics.nim(127) semLocal
src/nimony/sembasics.nim(447) takeParRi
src/nimony/programs.nim(163) bug
[Bug] mytests/test.nim(3, 3) expected ')', but got: (err . "attempt to redeclare\3A x")
```

Similar error is generated when declaring the same name types or for loop variables.

`addSym` proc in `src/nimony/sembasics.nim` calls `buildErr` proc when there is a redeclaration.
And `buildErr` adds `ErrT` node to `c.dest` that makes the parent node with wrong number of children.

ringabout:
fixes #1363

### DIFF
--- a/src/nimony/semdecls.nim
+++ b/src/nimony/semdecls.nim
@@ -128,10 +128,10 @@ proc semLocal(c: var SemContext; n: var Cursor; kind: SymKind) =
     bug "semLocal"
 
   if beforeType != -1:
-    var valueCursor = cursorAt(c.dest, beforeType)
     let hasError = c.addSymForwardError delayed
 
     if hasError:
+      var valueCursor = cursorAt(c.dest, beforeType)
       skip valueCursor # skips types
       let newValuePos = cursorToPosition(c.dest, valueCursor)
       var valueBuf = createTokenBuf()
@@ -141,8 +141,7 @@ proc semLocal(c: var SemContext; n: var Cursor; kind: SymKind) =
 
       let orig = beginRead(valueBuf)
       c.buildErr delayed.info, "attempt to redeclare: " & pool.strings[delayed.lit], orig
-    else:
-      endRead(c.dest)
+
   else:
     c.addSym delayed
 


src/nimony/semdecls.nim
```diff
@@ -123,7 +126,26 @@ proc semLocal(c: var SemContext; n: var Cursor; kind: SymKind) =
         patchType c, it.typ, beforeType
   else:
     bug "semLocal"
-  c.addSym delayed
+
+  if beforeType != -1:
```
Araq, line 130: I don't understand the relevance of a type existing.

src/nimony/semdecls.nim
```diff
@@ -123,7 +126,26 @@ proc semLocal(c: var SemContext; n: var Cursor; kind: SymKind) =
...
+    var valueCursor = cursorAt(c.dest, beforeType)
```
Araq: Can be moved inside the `hasError` branch, I think.

src/nimony/semdecls.nim
ringabout: I think 

```nim
  case kind
  of TypevarY:
```

The `TypevarY` branch doesn't have a value. I want to use this condition to skip that case

Issue #1360, opened 2025-08-04
`default` doesn't work for `importc` objects with non exportc fields


ringabout:
```nim
type
  MachTimebaseInfoData {.pure, final, importc: "mach_timebase_info_data_t",
      header: "<mach/mach_time.h>".} = object
    numer, denom: int32

var machAbsoluteTimeFreq: MachTimebaseInfoData = default(MachTimebaseInfoData)
```

It generates mangled field names for object constructors. In Nim, a `zeroMem` is generated instead for importc objects. Workaround: uses `exportc` for C types, i.e.

```nim
type
  MachTimebaseInfoData {.pure, final, importc: "mach_timebase_info_data_t",
      header: "<mach/mach_time.h>".} = object
    numer {.exportc.}, denom {.exportc.}: int32
```

PR #1370, opened 2025-08-07
fixes `std/locks` on windows

ringabout:
ref  https://github.com/nim-lang/nimony/issues/1360

also a test for dynlib pragmas for windows


Araq:
btw I think locks should not use the .dynlib feature no matter if it exists or not :-)

PR #1372, opened 2025-08-07
don't add pushed pragmas that already exists in the declaration

demotomohiro:
For example,
```nim
{.push header: "<headerX.h>", cdecl.}
proc foo() {.importc: "foo", header: "<headerY.h>".}
{.pop.}
```
Pushed pragma `header: "<headerX.h>"` is not added to `foo` proc as it already have `header: "<headerY.h>"`.
So above code is the same to following code:
```nim
proc foo() {.importc: "foo", header: "<headerY.h>", cdecl.}
```

https://github.com/nim-lang/nimony/pull/1317 has a problem that pushed pragmas are added to procedures twice.
This PR can fix it.

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1460,19 +1460,60 @@ proc semPragma(c: var SemContext; n: var Cursor; crucial: var CrucialPragma; kin
       while n.kind != ParRi: skip n
     skipParRi n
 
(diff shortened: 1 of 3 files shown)


demotomohiro:
Succeeded by https://github.com/nim-lang/nimony/pull/1424

PR #1373, opened 2025-08-08
porting `std/encodings` to Nimony

ringabout:


lib/std/encodings.nim
```diff
@@ -0,0 +1,524 @@
...
+  ## This opens a converter, uses it and closes it again and is thus more
+  ## convenient but also likely less efficient than re-using a converter.
+  ##
+  ## .. warning:: UTF-16BE and UTF-32 conversions are not supported on Windows.
+  var c = open(destEncoding, srcEncoding)
+  # TODO:
+  # try:
+  result = convert(c, s)
+  # finally:
+  #   close(c)
```
Araq: At. least call the `close(c)` here without any finally for the time being.

PR #1379, opened 2025-08-08
adds formatBiggestFloat/formatFloat

demotomohiro:


lib/std/strutils.nim
```diff
@@ -328,6 +328,26 @@ func cmpIgnoreStyle*(a, b: string): int =
     inc i
     inc j
 
+type
+  Natural = int
```
Araq, line 332: Move this alias to system.nim

demotomohiro: It is done in https://github.com/nim-lang/nimony/pull/1415

PR #1384, opened 2025-08-10
fixes module and type plugins

levovix0:


levovix0:
Is there currently no code to support the
```nim
template a_plugin =
  {.plugin: "deps/a.nim".}
```
plugin kind?

Araq:
> Is there currently no code to support the ...

shouldn't this be supported just because of a template's expansion semantics?

PR #1390, opened 2025-08-11
porting `std/envvars` to Nimony

ringabout:
windows pending https://github.com/nim-lang/nimony/pull/1374

parts of `std/os`

### DIFF
--- a/lib/std/envvars.nim
+++ b/lib/std/envvars.nim
@@ -165,10 +165,11 @@ when not defined(nimscript):
         if key.len == 0 or '=' in key:
           # raise newException(OSError, "invalid key, got: " & key)
           quit "invalid key, got: " & key
-        let envToDel = key & "="
-        if c_putenv(cstring envToDel) != 0'i32: bail()
+        var envToDel = key & "="
+        if c_putenv(toCString envToDel) != 0'i32: bail()
       else:
-        if c_unsetenv(cstring key) != 0'i32: bail()
+        var key = key
+        if c_unsetenv(toCString key) != 0'i32: bail()
 
     when defined(windows):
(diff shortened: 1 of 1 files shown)


lib/std/envvars.nim
```diff
@@ -0,0 +1,233 @@
...
+#                                             ## from an environment variable.
+#   WriteEnvEffect* = object of WriteIOEffect ## Effect that denotes a write
+#                                             ## to an environment variable.
+
+#------------------------------------------------------------------
+type
+  # TODO: {.importc: "char**", nodecl.}
+  cstringArray = ptr UncheckedArray[cstring]
+
+proc dollarPair*(key, value: string): string =
```
Araq: Seems to be new and thus should not be exported.

lib/std/private/win_setenv.nim
```diff
@@ -0,0 +1,109 @@
+#[
```
Araq, line 1: What is this file doing? Nim used to use the Windows API directly, now we have this crap.

ringabout: ref https://github.com/nim-lang/Nim/pull/18575  https://github.com/nim-lang/Nim/issues/18533

That PR introduced the changes. 

<del> Probably it intendted to make it consistent between platforms >/del>


```nim
  var
    envComputed {.threadvar.}: bool
    environment {.threadvar.}: seq[string]
````

The issue seems to indicate that the old version is not thread-safe

Araq: Use the version from before that uses the WinAPI.

Araq:
Ping.

PR #1391, opened 2025-08-11
porting oscommons to Nimony

ringabout:


src/hexer/nifcgen.nim
```diff
@@ -475,8 +475,11 @@ proc trType(c: var EContext; n: var Cursor; flags: set[TypeFlag] = {}) =
       return
     let res = tryLoadSym(s)
     if res.status == LacksNothing:
-      var body = asTypeDecl(res.decl).body
-      if body.typeKind == DistinctT: # skips DistinctT
+      var typeDecl = asTypeDecl(res.decl)
+      let prag = parsePragmas(c, typeDecl.pragmas)
```
Araq: Move the parsePragmas into after the `body.typeKind == DistinctT` case.

lib/std/windows/winlean.nim
```diff
@@ -11,11 +11,11 @@ when defined(windows):
 
   # See https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types
   type
-    Handle* {.importc: "HANDLE", header: "<WinDef.h>".} = distinct int
+    Handle* {.importc: "HANDLE", header: "<WinDef.h>", nodecl.} = distinct int
     LONG* {.importc: "LONG", header: "<WinDef.h>".} = int32
     WINBOOL* = distinct int32
       ## `WINBOOL` uses opposite convention as posix, !=0 meaning success.
-    DWORD* = int32
+    DWORD* {.importc: "DWORD", header: "<WinDef.h>", nodecl.} = int32
```
Araq: Now is the time to fix this bug, I'm pretty sure DWORD is actually uint32.

ringabout: Sure, I will see what I can do about it with the knowledge of https://github.com/nim-lang/Nim/issues/12327

PR #1394, opened 2025-08-12
use `toCString` before passed to importc calls

ringabout:
I didn't know Nimony strings are non-null terminated, probably a warning or something would be nicer


Araq:
> I didn't know Nimony strings are non-null terminated, probably a warning or something would be nicer

Yeah, the conversion should only be allowed for string literals.

PR #1407, opened 2025-08-19
fixes #1404

### LINKED ISSUE #1404: Object fields in the imported module hides variables in untyped templates
mod1.nim:
```nim
type
  Foo* = object
    fooField*: int
```

testbug.nim:
```nim
import mod1

template untypedTmpl() {.untyped.} =
  fooField = 'a'

var fooField: char
untypedTmpl()
```

Error message:
```
$ ./bin/nimony c -r mytests/testbug.nim 
mytests/testbug.nim(7, 12) Trace: instantiation from here
mytests/testbug.nim(4, 14) Error: type mismatch: got: char but wanted: int
```

If `Foo` object is declared in the main module like following code, it compiles without errors.
```nim
type
  Foo* = object
    fooField*: int

template untypedTmpl() {.untyped.} =
  fooField = 'a'

var fooField: char
untypedTmpl()
```

It seems the test failed in https://github.com/nim-lang/nimony/pull/1387 due to this issue.

demotomohiro:
untyped template doesn't bind identifiers to symbols at the template definition even if matched symbols are found.
Fixed math.nim as fixed code produced the compile error from it.

### DIFF
--- a/src/nimony/semuntyped.nim
+++ b/src/nimony/semuntyped.nim
@@ -402,6 +402,10 @@ proc semTemplBody*(c: var UntypedCtx; n: var Cursor) =
         # var yz: T
         c.c.dest.shrink start
         c.c.dest.add symToken(firstSym, n.info)
+      elif c.mode == UntypedTemplate:
+        # leave as ident
+        c.c.dest.shrink start
+        c.c.dest.add n
       else:
         semTemplSymbol(c, n, firstSym, count, start)
     inc n
--- a/lib/std/math.nim
+++ b/lib/std/math.nim
@@ -334,7 +334,7 @@ template ceilDivUint[T: SomeUnsignedInt and Arithmetic](x, y: T): T =
   # `x + (y - 1)` can overflow.
   (x + (y - T(1))) div y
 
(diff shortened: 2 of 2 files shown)


src/nimony/semuntyped.nim
```diff
@@ -402,6 +402,10 @@ proc semTemplBody*(c: var UntypedCtx; n: var Cursor) =
         # var yz: T
         c.c.dest.shrink start
         c.c.dest.add symToken(firstSym, n.info)
+      elif c.mode == UntypedTemplate:
+        # leave as ident
+        c.c.dest.shrink start
+        c.c.dest.add n
```
Araq, line 408: This fix makes no sense to me. Why does it bind to an object field to begin with?

demotomohiro: `buildSymChoice` proc in nimony/sembasics.nim picks the symbol of object field that is the same name to the given identifier if it is not defined.
Because if the identifier is not defined in the module nimsem reading, `considerImportedSymbols` proc try to find the same name symbols in imported modules and picks the same name symbol of the object field.

I misunderstand the problem and this PR doesn't fix.

PR #1412, opened 2025-08-20
fixes #1404

### LINKED ISSUE #1404: Object fields in the imported module hides variables in untyped templates

demotomohiro:
`considerImportedSymbols` proc in nimony/sembasics.nim ignores object field symbols in imported modules so that `buildSymChoice` proc never adds object field symbols to `c.dest`.

### DIFF
--- a/src/nimony/sembasics.nim
+++ b/src/nimony/sembasics.nim
@@ -28,9 +28,13 @@ proc considerImportedSymbols(c: var SemContext; name: StrId; info: PackedLineInf
   for moduleId in c.importTab.getOrDefault(name):
     # prevent copies
     let candidates = addr c.importedModules[moduleId].iface[name]
-    inc result, candidates[].len
     for defId in candidates[]:
-      c.dest.add symToken(defId, info)
+      let res = tryLoadSym(defId)
+      if res.status == LacksNothing and res.decl.symKind == FldY:
+        discard "ignores object fields as they are always used with dot or in an object constructor"
+      else:
+        c.dest.add symToken(defId, info)
+        inc result
 
 proc addSymUse*(dest: var TokenBuf; s: Sym; info: PackedLineInfo) =
   dest.add symToken(s.name, info)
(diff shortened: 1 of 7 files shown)


src/nimony/sembasics.nim
```diff
@@ -28,9 +28,13 @@ proc considerImportedSymbols(c: var SemContext; name: StrId; info: PackedLineInf
   for moduleId in c.importTab.getOrDefault(name):
     # prevent copies
     let candidates = addr c.importedModules[moduleId].iface[name]
-    inc result, candidates[].len
     for defId in candidates[]:
-      c.dest.add symToken(defId, info)
+      let res = tryLoadSym(defId)
+      if res.status == LacksNothing and res.decl.symKind == FldY:
+        discard "ignores object fields as they are always used with dot or in an object constructor"
```
Araq: So why are they in `candidates` then? They shouldn't be in the `iface` either!

demotomohiro: When nimsem writes an index file, all exported symbols including object fields are written in the public section in the index file.
When nimsem reads the index file, all symbols in the public section are added to `iface` and `importTab` in `loadInterface` proc in nimony/programs.nim.
If exported object field symbols shouldn't be in the `iface`, they need to be excluded when nimsem writes index files or when reads them.

Araq: > If exported object field symbols shouldn't be in the iface, they need to be excluded when nimsem writes index files or when reads them.

Yes, exactly. When it writes them.

demotomohiro: done

PR #1419, opened 2025-08-22
adds `%`, `format`, `strip` and `find` to strutils

demotomohiro:


lib/std/strutils.nim
```diff
@@ -500,3 +590,160 @@ func formatFloat*(f: float, format: FloatFormatMode = ffDefault,
...
+  while i < inArray.len - 1:
+    if cmpIgnoreStyle(x, inArray[i]) == 0: return i
+    inc(i, 2) # incrementing by 1 would probably lead to a
+              # security hole...
+  return -1
+
+proc invalidFormatString(formatstr: string) {.noinline.} =
+  # TODO: Uncomment when exceptions are implemented.
+  #raise newException(ValueError, "invalid format string: " & formatstr)
+  quit "ValueError: invalid format string: " & formatstr
```
Araq: You can already use 

`raise SyntaxError` here.

PR #1423, opened 2025-08-22
Add mimalloc alloc stats

planetis-m:
Based on: https://github.com/planetis-m/mimalloc_nim/blob/master/src/patchedstd/mimalloc.nim

Note 1: https://github.com/nim-lang/mimalloc/ needs to be updated to a recent version as it lacks the header "mimalloc-stats.h"

Note 2: passC is not implemented so the statements at the top of the linked file were removed but it needs either MI_STAT=1 or MI_DEBUG>0

### DIFF
--- a/lib/std/system/mimalloc.nim
+++ b/lib/std/system/mimalloc.nim
@@ -1,4 +1,4 @@
-{.build("C", "${path}/../../../vendor/mimalloc/src/static.c", "-I${path}/../../../vendor/mimalloc/include").}
+{.build("C", "${path}/../../../vendor/mimalloc/src/static.c", "-DMI_STATS=1 -I${path}/../../../vendor/mimalloc/include").}
 
 type
   MiStatCount {.importc: "mi_stat_count_t", bycopy.} = object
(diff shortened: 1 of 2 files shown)


planetis-m:
@demotomohiro I tried to do that but before [nim-lang/mimalloc/](https://github.com/nim-lang/mimalloc/) is updated it polutes the tests with error messages.

lib/std/system/mimalloc.nim
```diff
@@ -18,3 +51,34 @@ proc dealloc*(p: pointer) =
 
 proc allocatedSize*(p: pointer): int =
   result = int mi_usable_size(p)
+
+when defined(debug):
```
Araq: Huh? These are not debug specific, they are supposed to be always available.

planetis-m: I just tried it in nim v2, and with -DMI_STAT=1 it can still show these basic stats. Not sure how to pass C flags in nimony, passC doesn't work.

Araq: Just edit the `{.build("C", "${path}/../../../vendor/mimalloc/src/static.c", "-I${path}/../../../vendor/mimalloc/include").}` thing.

ringabout:
> @ringabout Please update mimalloc

done

PR #1428, opened 2025-08-25
porting `std/envvars` to Nimony

ringabout:


lib/std/envvars.nim
```diff
@@ -0,0 +1,277 @@
...
+    ## * `putEnv proc <#putEnv,string,string>`_
+    ## * `envPairs iterator <#envPairs.i>`_
+    var indx = findEnvVar(key)
+    if indx < 0: return # Do nothing if the env var is not already set
+
+    when defined(windows) and not defined(nimscript):
+      var key = key
+      var k = newWideCString(key.toCString(), key.len).toWideCString()
+      if setEnvironmentVariableW(k, nil) == 0'i32:
+        raiseOSError(osLastError())
```
Araq, line 255: This should trigger a `.raises` annotation. How come it doesn't?

Issue #1426, opened 2025-08-24
`typeof` does not exist


Araq:
The manual says `typeof` exists like in Nim. But it doesn't, it's never been added to system.nim properly. There are test cases that use it though.

PR #1431, opened 2025-08-25
adds `typeof` to system module

demotomohiro:
Fixes https://github.com/nim-lang/nimony/issues/1426.
tests/nimony/sysbasics/ttypeof.nim needs https://github.com/nim-lang/nimony/pull/1430.


Araq:
In a follow-up PR add the 

type
  TypeOfMode* = enum ## Possible modes of `typeof`.
    typeOfProc,      ## Prefer the interpretation that means `x` is a proc call.
    typeOfIter       ## Prefer the interpretation that means `x` is an iterator call.

logic

PR #1433, opened 2025-08-26
`raiseOSError` now uses `errorcodes`

ringabout:


lib/std/memfiles.nim
```diff
@@ -110,7 +110,8 @@ proc open*(filename: string, mode: FileMode = fmRead,
   if mode == fmAppend:
     # TODO: raise exception when it is supported.
     #raise newEIO("The append mode is not supported.")
-    quit "The append mode is not supported."
+    # quit "The append mode is not supported."
+    raise Failure
```
Araq:
```suggestion
    raise BadOperation
```

ringabout:
> Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.1.0 (MinGW-W64 x86_64-posix-seh, built by Brecht Sanders) 

I use gcc14 locally. Perhaps we need to update mingw

PR #1438, opened 2025-08-27
porting ospaths2 to Nimony

ringabout:


lib/std/private/ospaths2.nim
```diff
@@ -0,0 +1,1036 @@
...
+
+import ../pathnorm
+
+
+
+proc `&`(x: string, y: char): string {.inline.} =
+  result = x & $y
+
+proc `&`(x: char, y: string): string {.inline.} =
+  result = $x & y
```
Araq: Come on, these really belong to stringimpl.nim

lib/std/private/ospaths2.nim
```diff
@@ -0,0 +1,1036 @@
...
+    path.add DirSep
+  elif i > 0:
+    # foo// => foo
+    path.setLen(i)
+  else:
+    # // => / (empty case was already taken care of)
+    path = $DirSep
+
+
+# TODO: Overloading with varness doesn't work in Nimony yet
```
Araq: Overloading based on varness is language design bug and will only be added for backwards compat.

Araq:
Also needs tests.

PR #1439, opened 2025-08-27
rework `from module import nil`

ringabout:


Araq:
"rework"?! It's a one line fix! :D

PR #1449, opened 2025-08-29
fixes #1440; treat Symbol as a non-simple expression while travesing return statement in duplifier

### LINKED ISSUE #1440: nimony defer can mutate local variables, unlike nim et al.
Of this Nim/Nimony code, and also the try/finally rewrite,

```nim
import std/syncio
proc f(n: int): int =
  var n = n
  defer:
    inc n
    echo n
  return n

echo f(5)
```

Nim 2.2.4 outputs
```
6
5
```
and Nimony outputs
```
6
6
```

Similar programs in other languages with defer (zig, odin, nelua, go, d) all have Nim's output.

saffage:
It's unsafe to use a symbol during return as it might be modified indirectly from finally\\defer blocks, which changes `return` semantics.

### DIFF
--- a/src/hexer/duplifier.nim
+++ b/src/hexer/duplifier.nim
@@ -197,7 +197,7 @@ proc isResultUsage(c: Context; n: Cursor): bool {.inline.} =
 proc isSimpleExpression(n: var Cursor): bool =
   ## expressions that can be returned safely
   case n.kind
-  of Symbol, UIntLit, StringLit, IntLit, FloatLit, CharLit, DotToken, Ident:
+  of UIntLit, StringLit, IntLit, FloatLit, CharLit, DotToken, Ident:
     result = true
     inc n
   of ParLe:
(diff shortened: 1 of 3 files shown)


Araq:
Why pessimize the codegen for the one silly edge case? Make the analysis smarter here.

saffage:
Well, I tried to transform the code after the body of the procedure had been checked (immediately after `transformDefer`), since we (probably) need complete information about the variables used. Information about which variables used inside the `finally` block is collected before traversing the `try` statements. Then, if the variable is returned directly, the code is transformed into `result = x; return result`. I think there are other ways to break the semantics again, such as using the variable indirectly, but in that case, it would require a deeper analysis of the returned value.

My knowledge of how the compiler works is not very deep, and I don't know the compiler's codebase very well, so I don't think I've managed to do this optimally or that my solution makes sense at all.

If you think this solution is bad or doesn't make sense, feel free to close the PR.

PR #1450, opened 2025-08-30
adds delete, trimZeros, formatSize and replace to strutils.nim

demotomohiro:


lib/std/strutils.nim
```diff
@@ -143,6 +143,36 @@ proc `$`*(x: char): string =
   result = newString(1)
   result[0] = x
 
+func delete*(s: var string, slice: Slice[int]) {.raises.} =
```
Araq: shouldn't have the `.raises`, instead use `assert` for the check.

demotomohiro: Fixed it.

PR #1521, opened 2025-10-30
implements `incompleteStruct`

ringabout:
```nim
  type
    DIR* {.importc: "DIR", header: "<dirent.h>", incompleteStruct.} = object
    Dirent* {.importc: "struct dirent", header: "<dirent.h>".} = object
      d_type* {.importc: "d_type".}: uint8
      d_name* {.importc: "d_name".}: array[256, char]
```

As in Nim, `DIR` is marked `incompleteStruct`.

Since `DIR` is forward declared, we can not access its size. We need to handle the `sizeof` for it specifically

In Nim, it causes problems for `genTypeInfo`. And some (mis)usage still produce code gen errors

```nim
type
  DIR* {.importc: "DIR", header: "<dirent.h>", incompleteStruct.} = object

let sss = sizeof(DIR)
echo sss
```
or use `DIR` as fields of objects without reference, name using `DIR` by value

In Nimony, it seems that only `sizeof` concerns `incompleteStruct` for now, i.e., `sizeof(DIR)`. Since any type using `DIR` by 
value already produces code gen errors. This PR should make `sizeof(DIR)` use the size of a pointer, while `incompleteStruct` might be useful in the future

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -668,7 +668,7 @@ proc semStmtCallback(c: var SemContext; dest: var TokenBuf; n: Cursor) =
 
 
 proc semGetSize(c: var SemContext; n: Cursor; strict=false): xint =
-  getSize(n, c.g.config.bits div 8, strict, isEval = true)
+  getSize(n, c.g.config.bits div 8, strict)
 
 proc sameIdent(sym: SymId; str: StrId): bool =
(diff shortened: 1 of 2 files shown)


Araq:
Well sizeof must detect this pragma and return "unknown" (I don't know if it can do that already, but we need this).

Issue #1534, opened 2025-11-06
implements `std/osproc`


ringabout:
requirements:

- [ ] std/cpuinfo
- [ ] std/strtabs
- [ ] std/streams
- [ ] std/streamwrappers

Araq:
Meh, I don't think we need all this, just give us a way to run external processes with stdin/out redirection.

ringabout:
Fair, it seems that `streams` etc. can be left out since they exist as functions while `strtabs` is used in relevant function parameters

Araq:
std/osproc.nim exists now. Needs way more testing, of course.

PR #1536, opened 2025-11-06
implements `std/streams`

ringabout:
ref https://github.com/nim-lang/nimony/issues/1534

`std/streams` is required by `std/osproc`


ringabout:
will be succeeded by https://github.com/nim-lang/nimony/pull/1809

PR #1543, opened 2025-11-13
fixes #1542; proc type doesn't match generic constraints proc

### LINKED ISSUE #1542: `proc type` doesn't match generic constraints `proc`
```nim
template foo*[T: proc](x: typedesc[T]): T = T(nil)

var m = foo(proc (s: int))
```

```
(proctype . . . . . . . . .)
(proctype . . . .
 (params
  (param :s.1 . .
   (i -1) .)) . . . .)
```

It is required to implement `default` for proc types and https://github.com/nim-lang/nimony/pull/1536 (because `std/streams` use massive proc pointers)

ringabout:
fixes #1542

Unlike other types classes, `proc` type class is parsed as `(proctype . . . . . . . . .)`. It needs to skip `DotToken` in order to identify whether it is a type class

### DIFF
--- a/src/nifler/bridge.nim
+++ b/src/nifler/bridge.nim
@@ -405,6 +405,11 @@ proc toNif*(n, parent: PNode; c: var TranslationContext; allowEmpty = false) =
     else:
       c.b.addTree(ItertypeL)
 
+    if n.len == 0:
+      # it's a type class
+      c.b.endTree()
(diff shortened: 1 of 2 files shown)


ringabout:
I was not sure which part should handle this. I will do this check in nifler to generate `(proctype)`

PR #1562, opened 2025-11-20
implements basic os functions for `std/osproc`

ringabout:
ref https://github.com/nim-lang/nimony/issues/1534

### DIFF
--- a/lib/std/os.nim
+++ b/lib/std/os.nim
@@ -85,19 +85,19 @@ proc exitStatusLikeShell*(status: cint): cint =
     status
 
 proc execShellCmd*(command: string): int {.tags: [ExecIOEffect].} =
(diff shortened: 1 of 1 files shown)


lib/std/os.nim
```diff
@@ -0,0 +1,103 @@
...
+    ##
+    ## To execute a program without having a shell involved, use `osproc.execProcess proc
+    ## <osproc.html#execProcess,string,string,openArray[string],StringTableRef,set[ProcessOption]>`_.
+    ##
+    ## **Examples:**
+    ##   ```Nim
+    ##   discard execShellCmd("ls -la")
+    ##   ```
+    var command = command
+    result = exitStatusLikeShell(c_system(command.toCString))
```
Araq: Indentation is 2 spaces.

PR #1571, opened 2025-11-23
fixes errors on 32bit CPU

demotomohiro:
This PR fixes compile errors and a part of failed tests on 32bit CPU(i386 and ARM).

### DIFF
--- a/lib/std/syncio.nim
+++ b/lib/std/syncio.nim
@@ -25,12 +25,6 @@ var
   stdout* {.importc: "stdout", header: "<stdio.h>".}: File
   stderr* {.importc: "stderr", header: "<stdio.h>".}: File
 
-let
-  PRId32 {.importc: "\"%\" PRId32", header: "inttypes.h".}: cstring
-  PRId64 {.importc: "\"%\" PRId64", header: "inttypes.h".}: cstring
-  PRIu32 {.importc: "\"%\" PRIu32", header: "inttypes.h".}: cstring
-  PRIu64 {.importc: "\"%\" PRIu64", header: "inttypes.h".}: cstring
-
 proc c_fputc(c: int32; f: File): int32 {.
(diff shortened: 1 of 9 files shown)


lib/std/syncio.nim
```diff
@@ -25,6 +25,12 @@ var
   stdout* {.importc: "stdout", header: "<stdio.h>".}: File
   stderr* {.importc: "stderr", header: "<stdio.h>".}: File
 
+let
+  PRId32 {.importc: "\"%\" PRId32", header: "inttypes.h".}: cstring
+  PRId64 {.importc: "\"%\" PRId64", header: "inttypes.h".}: cstring
+  PRIu32 {.importc: "\"%\" PRIu32", header: "inttypes.h".}: cstring
+  PRIu64 {.importc: "\"%\" PRIu64", header: "inttypes.h".}: cstring
+
```
Araq: Nim's stdlib works on 32bit and doesn't need this weird crap.

demotomohiro: On 32bit CPU, `stdout.write 0` causes compile error: `ambiguous call: 'write'`.
Adding overloaded proc for `int32` like `proc write*(f: File; x: int32) = ...` fixes the compile error.

Araq: Sure but I don't even know what "PRId64" means

demotomohiro: `%d`, `%ld` and `%lld` work for `int`, `long`, `long long` in C, but bit size of these types are not fixed in C spec.
C has fixed width integer types like `int32_t` or `int64_t`.
`PRId32` and `PRId64` macros are used in `fprintf` arguments to print fixed width integer types.
https://en.cppreference.com/w/c/header/inttypes

I thought `PRId32` or `PRId64` is better than "%d" or "%lld" to print fixed bit size int types in Nim with `fprintf`.

PR #1578, opened 2025-11-26
fixes #1574; calling $ on enum value crashes nimony

### LINKED ISSUE #1574: calling $ on enum value crashes nimony
``` nim
import std / syncio
type A = enum
    off, on
  LevelSetting = array[0..3, A]
var level: LevelSetting
for x in level:
  echo $x
```

unhandled exception src/lib/nifstreams.nim(233, 3) `n.kind in {Symbol, SymbolDef}` [AssertionDefect]

ringabout:
fixes #1574

adds `skipModifier` for `semEnumToStr`
adds semchecking back for `semEnumToStr` which is needed for publishing signature and transforming `ret ...` into `ret result` for `controlflow.nim`


Araq:
Fine for now but in the future do the registration and `ret result` yourself instead of letting the semcheck do it.

PR #1600, opened 2025-12-07
implement bitops in expreval

### LINKED ISSUE #1598: "cannot evaluate expression at compile time"
The code below does not compile:
```nim
import std/syncio

const
  a = 1
  b = 2
  c = 4
  ab = a or b
  ac = a and c
  bc = b xor c
  anot = not a
  bShra = b shr a
  bShla = b shl a

echo ab
echo ac
echo bc
echo anot
echo bShra
echo bShla
```
Current output:
```
nimony c test.nim
test.nim(7, 10) Error: cannot evaluate expression at compile time: a bitor b
test.nim(8, 10) Error: cannot evaluate expression at compile time: a bitand c
test.nim(9, 10) Error: cannot evaluate expression at compile time: b bitxor c
test.nim(10, 10) Error: cannot evaluate expression at compile time: a bitnot
test.nim(11, 13) Error: cannot evaluate expression at compile time: ashr(b, a)
test.nim(12, 13) Error: cannot evaluate expression at compile time: b shl a
make: E:\nimony\bin\nimsem.exe m --isMain nimcache\teszpno6r1.p.nif
FAILURE: E:\nimony\bin\nifmake.exe --base:"" -j run nimcache\teszpno6r1.build.nif
```

metagn:
fixes #1598

Both `ShrX` and `AshrX` use `xints.shr` for now, this may be wrong. The new `mask`/`not` code may also be wrong.

### DIFF
--- a/src/nimony/expreval.nim
+++ b/src/nimony/expreval.nim
@@ -321,7 +321,7 @@ template evalShiftOp(c0: var EvalContext; n: var Cursor; opr: untyped) {.dirty.}
     bits = typebits(n.load)
     skipToEnd n
   else:
-    error "expected int or uint type for shl, got: " & typeToString(n), n.info
+    error "expected int or uint type for shift operation, got: " & typeToString(n), n.info
   if bits < 0: bits = c0.c.g.config.bits
   let a = getConstOrdinalValue propagateError eval(c0, n)
   let b = getConstOrdinalValue propagateError eval(c0, n)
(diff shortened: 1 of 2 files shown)


src/nimony/expreval.nim
```diff
@@ -506,6 +623,19 @@ proc eval*(c: var EvalContext; n: var Cursor): Cursor =
...
+    of BitandX:
+      evalOrdBinOp(c, n, `and`)
+    of BitxorX:
+      evalOrdBinOp(c, n, `xor`)
+    of BitnotX:
+      evalBitnot(c, n)
+    of ShlX:
+      evalShiftOp(c, n, `shl`)
+    of ShrX, AshrX:
+      # considered the same for now
```
Araq: Nope, don't offer `ashr` until it's correct. "cannot eval ashr at compile-time" is **much** better than having to hunt subtle bugs.

metagn: The problem is that the `shr` overloads for integers use `AshrI` when the unsigned integers use `ShrI`, and `ashr` is only defined for signed integers. I will make it just only allow the respective type for each operation

PR #1764, opened 2026-04-04
CPS bugfix

Araq:


src/hexer/cps.nim
```diff
@@ -448,7 +502,14 @@ proc trPassiveCall(c: var Context; dest: var TokenBuf; n: var Cursor; sym: SymId
       if hasResult:
         dest.copyIntoKind AddrX, info:
           dest.copyTree target
-      contNextState c, dest, state, info
+      if returnsToCaller:
```
blackmius, line 505: we dont have specific test for this optimization

Araq: `tif` seems to cover it.

PR #1772, opened 2026-04-11
fix(cps): branching

blackmius:
# better yields and labels for ite

somewhere in commits I patched nifstreams.toString() and added there debug print for yields and labels

for this code it creates such mappings
```nim
proc main2() {.passive.} =
  if true:
    if true:
      a()
      if true:
        a()
        if true:
          a()
    else:
      a()
      discard
      discard
    c()
    discard
    discard
  elif true:
    c()
  if true:
    discard
  else:
    discard

main2()
```

```
(stmts
 (mflag :´r.0)
 (ite
  (true)
  (stmts
   (ite
    (true)
    (stmts
     (call<YIELD 1> a.0.teszpno6r1)
     (ite<LABEL 1>
      (true)
      (stmts
       (call<YIELD 2> a.0.teszpno6r1)
       (ite<LABEL 2>
        (true)
        (stmts
         (call<YIELD 3> a.0.teszpno6r1)) .<YIELD 3>)) .<YIELD 3>))
    (stmts
     (call<YIELD 4> a.0.teszpno6r1) .<LABEL 4> .<YIELD 3>))
   (call<LABEL 3><YIELD 5> c.0.teszpno6r1) .<LABEL 5> .<YIELD 6>)
  (stmts
   (ite
    (true)
    (stmts
     (call<YIELD 6> c.0.teszpno6r1)) .<YIELD 6>)))
 (ite<LABEL 6>
  (true)
  (stmts .)
  (stmts .))
 (kill ´r.0))
 ```
 
 # the compileStmtSeq now handles deep nested ifs
compileStmtSeq accumulates ite splits and emits states at the end

### CODE REVIEW

File: src/hexer/cps.nim [hunk 1]
```diff
@@ -1497,15 +1525,14 @@ proc trIteStmts(c: var Context; dest: var TokenBuf; n: var Cursor) =
     let state = c.currentProc.labels.getOrDefault(p, -1)
     if state != -1:
       # split inside branch — save position and stop emitting into this branch
-      c.inlineContState = state
-      c.inlineContCursor = n
+      c.createNewLocalProc.add (state, n)
       while n.kind != ParRi: skip n  # skip rest of branch
```
blackmius, line 1529: it needs afterNode here, if it was a call with a complex expressions inside it will stopped at expr parri not the call

src/hexer/cps.nim
```diff
@@ -1531,10 +1555,8 @@ proc trIte(c: var Context; dest: var TokenBuf; n: var Cursor) =
     dest.copyIntoKind ElseU, info:
       dest.copyIntoKind StmtsS, info:
         dest.add elseBuf
-  # If a split was detected inside the then-branch, signal it to the caller
-  if splitState >= 0:
-    c.inlineContState = splitState
-    c.inlineContCursor = splitCursor
+  let p = cursorToPosition(c.currentProc.cf, n)
```
blackmius, line 1558: here it think it is easier to relate on labeling and if ite had a yield inside then right after ite should be a new label which needs to be also added to createNewLocalProc and skipped in current compileStmtSeq

PR #1780, opened 2026-04-12
documented plugin system

Araq:


Araq:
Ping @planetis-m , please review.

doc/plugins.md
```diff
@@ -0,0 +1,456 @@
...
+template renderTree(s: typed) {.plugin: "deps/mrender".}
+
+renderTree:
+  let x = 1
+  echo x
+```
+
+Here the plugin receives the entire block as typed, semantically checked NIF.
+
+Plugins can be hidden inside imported modules so that callers don't see the
```
levovix0: This should be in the "Module plugins" section

doc/plugins.md
```diff
@@ -0,0 +1,456 @@
+# Plugins
+
+Plugins are Nimony's metaprogramming mechanism, replacing Nim 2's macro system.
+A plugin is a separate Nim program that transforms NIF (Nimony Intermediate Format)
+trees at compile time. Plugins run as external processes, communicating with the
+compiler through NIF files.
+
+## Overview
+
+There are three kinds of plugins:
```
levovix0: There are also "import plugins"

planetis-m:
Nothing stands out to me as a mistake. Looks good.

doc/plugins.md
```diff
@@ -0,0 +1,456 @@
...
+| `charLit(n)` | `char` | Character literal |
+| `eqIdent(n, name)` | `bool` | True if token matches `name` (works for idents and symbols) |
+
+**Source location helpers:**
+
+| Proc | Description |
+|------|-------------|
+| `isValid(info)` | True when info refers to a real source location |
+| `filePath(info)` | Source file path, or `""` |
+| `lineCol(info)` | `SourcePos(line, col)` (1-based), or `(0, 0)` |
```
planetis-m, line 382: I guess (0, 0) here means invalid.

doc/plugins.md
```diff
@@ -0,0 +1,456 @@
...
+```nim
+var n = loadPluginInput()
+if n.stmtKind == StmtsS: inc n
+# n now points to the first argument
+```
+
+A template plugin can also accept a code block via `typed` parameters:
+
+```nim
+template renderTree(s: typed) {.plugin: "deps/mrender".}
```
blackmius, line 207: looks like plugins lacks the way to be applied to a single function via pragma like macros do

Araq: Maybe in the current implementation it's broken, but in principle the rewrite rule `proc p() {.temp.} -> temp: proc p()` exists like it does in Nim.

doc/plugins.md
```diff
@@ -0,0 +1,456 @@
...
+proc renderTree*(tree: Tree): string        # debug rendering (no line info)
+```
+
+
+### NIF templates and interpolation
+
+For quick tree construction, use NIF string templates with `%~` interpolation:
+
+```nim
+let greeting = """(call echo $msg)""" %~ {"msg": ~"hello"}
```
blackmius, line 451: cant it work like json `%*` function work, is there a situation when i dont need to prepend a `~`?

planetis-m: Prepend is necessary, because for json its a macro, here we have to use functions only.

doc/plugins.md
```diff
@@ -0,0 +1,456 @@
...
+### Writing output
+
+```nim
+proc saveTree*(tree: Tree)                  # writes to paramStr(2)
+proc saveTree*(tree: Tree; filename: string)
+proc renderTree*(tree: Tree): string        # debug rendering (no line info)
+```
+
+
+### NIF templates and interpolation
```
blackmius, line 446: do you consider adding nim code interpolation?
like this one https://nim-lang.org/docs/macros.html#quote%2Ctyped%2Cstring

doc/plugins.md
ZoomRmc, line 1: At this stage this reads more like a specification, not a user-facing document.

Lacks the following:
- Why the architectural change
- What new possibilities does it open, what's not possible anymore
- What's now harder/easier to do
- Best practices for the developers
- Cost analysis (how expensive is a single subprocess for each template, how to bundle multiple templates in a same plugin?)
- Comparison of the Nim 2 AST structure with Nif, quirk by quirk. Like, in Nim a body is usually a statement list, but can be a single call. Is it *always* `StmtsS` with Nif?

 ---

- "How plugins are compiled and run" is positioned before any introduction to Nif structure. "Tree" is used everywhere, even though Nif is explicitly not a tree but a token stream. T
- Terms like `StmtsS` dropped and used without explanation (like "`Node` is wrapped in a `StmtsS` node" without giving any explanation what constitutes a node).

planetis-m:
Add something like:

`Tree` is the mutable copy-on-write builder. Copying a Tree shares the payload; the next mutation detaches it.
`Node` is an owned read handle into a frozen snapshot. Copying a Node creates another read handle to the same snapshot.

doc/plugins.md
Araq: Many, though of course not all of these questions are solved by simply linking to doc/tags.md, the one source of truth when it comes to Nimony's architecture.

Araq:
@planetis-m Zoom has a valid point though, maybe we can find somewhat better names for Tree and Node (though they are not too bad).

doc/plugins.md
ZoomRmc: I know you tend to lean towards not "re-epxlaining the world" but an introductory paragraph that explains the required mental context in a tightly compressed form never hurt anybody, as long as the document's structure is sane so it's easily skippable.

Its main raison d'être, as I see it, is actually to serve as a self-check for the reader. If one goes "yeah, yeah, get on with it", than great, skip ahead. Otherwise, if it's "oh, I'm not sure if I'm actually following what they're saying here", then you just follow the links, in this case to `doc/tags.md` and others.

For invested developers failing on the cursed side of the curse of knowledge is statistically more probable, so it's optimal to always try to counter-balance when writing docs. The negative impact (bloat) is solved with deliberate structure, footnotes/hints/asides/<summaries> and links.

Take it from the guy who is on the other side more often than not.

planetis-m:
NifBuilder and NifCursor is perhaps the most technically correct names.

> @planetis-m Zoom has a valid point though, maybe we can find somewhat better names for Tree and Node (though they are not too bad).

doc/plugins.md
Araq: Makes sense.

PR #1782, opened 2026-04-14
feat(cps): enable passive closures

blackmius:
it turned out, it just needed to lift corotype to top level so vtables could handle it


blackmius:
not for long 

this crashes
```nim
proc e() {.passive.} =
    echo "e"

proc q(x: int): proc() {.passive,closure.} =
    return proc(){.passive,closure.} =
        e()

proc main2() =
    var x = q(5)
    x()
main2()
```

Araq:
It improves the status-quo and is clean, I'll take it.

PR #1783, opened 2026-04-15
feat(cps): cps now handles complex closures

blackmius:
this approach is changing what needs to be changed after liftlambda pass
1. converts each proctype passive to its init wrapper form
2. supports calls to tuple[fn, env] where fn is passive
3. converts assignments, locals, symbols of passive function to its init wrapper
4. lift nested passive procs to top level


tests/nimony/cps/tclosure.nim
```diff
@@ -5,4 +5,62 @@ proc main1() =
...
+
+
+proc a4(x: int) {.passive.} =
+    echo "a4: ", x
+
+proc a5(x: int) {.passive.} =
+    echo "a5: ", x
+
+proc main5() =
+    var x: (proc(x: int) {.passive.}) = a4
```
blackmius: this is a hack, because var x = a4 converts to proc declaration instead of proctype
```
(var :x.1 . .
    (proc :a4.0.teszpno6r1 . . .
     (params
      (param :x.0 . .
       (i +64) .)) .
     (pragmas
      (passive)) .
     (stmts
      (stmts
       (stmts
        (cmd write.0.syn1lfpjv stdout.0.syn1lfpjv "a4\3A "))
       (stmts
        (cmd write.2.syn1lfpjv stdout.0.syn1lfpjv x.0))
       (cmd write.7.syn1lfpjv stdout.0.syn1lfpjv '\0A')))) a4.0.init.teszpno6r1)
```

it looks like nimsem needs some patches so it will replace such things

blackmius:
crashes when running closure constructor inside passive proc

```nim
proc a2(x: int): proc (q: int) {.passive,closure.} =
    var b = x;
    return proc (q: int) {.passive,closure.} =
        b += q
        echo "a2 ", b

proc main3() {.passive.}=
    let x = a2(10)
    x(10)

main3()
```

blackmius:
also it has no check if non passive function is assigned to variable of passive type and vice versa, like it is done with closures

```
test.nim(12, 9) Error: type mismatch: got: proc (x: int64) {.closure.} but wanted: proc (x: int64)
```

tests/nimony/cps/tclosure.nim
Araq: I've noticed this problem too, nimsem needs to be more consistent about proctype, yes.

tests/nimony/cps/tclosure.nim
```diff
@@ -5,4 +5,104 @@ proc main1() =
...
+
+proc main7() {.passive.} =
+    var b = 0
+    let x = a()
+    x()
+    x()
+
+main7()
+
+# proc main8() {.passive.} =
```
blackmius: The passive closure constructor needs to lift an inner function. The NJVL pass operates on the entire procedure body and applies multiple times, so it needs to extract functions first and only then perform the transformation.

blackmius:
it becomes messy with all transformations for closure and non closure variants, a little refactor wouldn't hurt

Araq:
well CPS runs after closure elimination, I don't understand why it even needs to consider closures, they are all gone by the time it runs.

blackmius:
> well CPS runs after closure elimination, I don't understand why it even needs to consider closures, they are all gone by the time it runs.

because you need to replace all closure calls that calls passive closure with correct arguments
```
(call
  (tupat x.0 +0)
  (tupat x.0 +1))
```
should be added return pointer, and caller

and these calls can be everywhere, asgn, locals

and this additional checks if a call closure sums up

also this PR handles proctypes, so you can assign proc to variable and then reassign it to another, and same as closures you cant know which Continuation type it need so you need to always pick init wrapper (originally made for methods)

and also when x.0 is initialising it has proctype of default proc, without continuation arguments, which also requires patching

Araq:
We should fix proctypes in the frontend first so that we see how many of these workarounds can be avoided.

blackmius:
> We should fix proctypes in the frontend first so that we see how many of these workarounds can be avoided.

for proctype patching i added another pass, which runs at the end

https://github.com/nim-lang/nimony/pull/1783/changes#diff-934eed5a4c0e5733c8faba77d5b33fafa474b21d1613a41d9486940ee0cdb1f9R1667

the problem was that checks were performed on every assignment and local

if it is a call, it might be a passive call, but what kind of passive call closure is it? then it needs to be transformed, while still remaining a `tupleat`. if its a call to a variable holding an init wrapper, it also needs to stay as is. if its an explicit call to a passive function, it needs to be transformed with optimizations (maybe creating the continuation on the stack)

if its an assignment to a passive function without a call, then an init wrapper symbol needs to be created from it

and this logic exists in three different places, each with a different approach

overall, this clearly needs a refactor

Araq:
> overall, this clearly needs a refactor

Yup.

blackmius:
what lasts: (all patches are in nimsem, so maybe it should be in another PR)

1. patch nimsem that it will become more consistent about proctype
``` nim
import std/syncio
proc a4(x: int) {.passive.} =
    echo "a4: ", x
proc a5(x: int) {.passive.} =
    echo "a5: ", x
var x = a4
x(5)
x = a5
x(42)
```
2. check passive pragma on assignment 
``` nim
import std/syncio
proc a4(x: int) {.passive.} =
    echo "a4: ", x
proc a5(x: int) =
    echo "a5: ", x
var x: (proc(x: int) {.passive.}) = a4
x(5)
x = a5
x(42)
```
3. infer passive pragma from return type
``` nim
import std/syncio
proc a(x: int): proc (q: int): int =
    var b = x;
    return proc (q: int): int {.passive,closure.} =
        b += q
        return b
var q = a(10)
```

src/hexer/cps.nim
```diff
@@ -1517,13 +1503,141 @@ proc generateContinuationProcImpl(): Cursor =
...
+  let s = extractVersionedBasename(pool.syms[procId])
+  result = pool.syms.getOrIncl(s & ".lifted." & $index & "." & c.thisModuleSuffix)
+
+proc liftPassiveClosures(c: var Context; n: var Cursor): TokenBuf =
+  var dest = createTokenBuf(16)
+  var nested = 0
+  var needsLift: seq[tuple[n: Cursor, parent: int, name: SymId]] = @[]
+  var procStack: seq[tuple[pos: int, depth: int]] = @[]
+  var procRename = initTable[SymId, SymId]()
+  # search for nested passive procs
```
Araq: Why would there be such things? Lambda-lifting must eliminate them.

blackmius: it does not, move to lambdalift?

Araq: Yes.

src/hexer/lambdalifting.nim
```diff
@@ -776,6 +831,7 @@ proc elimLambdas*(pass: var Pass) =
       tre(c, pass.dest, n2)
     pass.dest.takeParRi n2
     endRead(oldDest)
+    pass.dest = liftProcs(c, beginRead(pass.dest))
```
Araq: Well no, embed the transformation in the general `tre` transformation.

blackmius: i dont think it is possible in one pass

blackmius: the best i can come up with is:
when you meet first proc - create new tokenBuf and continue to accumulate tokens there, then do the same lift

ignore creating new dest inside via some context flag

src/hexer/cps.nim
```diff
@@ -1517,6 +1503,62 @@ proc generateContinuationProcImpl(): Cursor =
       return t.body
   return default(Cursor)
 
+proc updatePassiveClosureProcTypes(c: var Context; dest: var TokenBuf; n: var Cursor) =
```
Araq: Seems easy to do in the same pass.

src/hexer/lambdalifting.nim
```diff
@@ -64,6 +64,8 @@ type
     typeCache: TypeCache
     thisModuleSuffix: string
     procStack: seq[SymId]
+    needsLift: seq[seq[int]]
```
Araq: what is that? why do we need it? we have an explicit `dest` parameter, use it.

blackmius: because i want transform this
``` nim
proc a():
  proc b() =
    proc d() =
      discard
    d()
  b()
  proc c() =
    discard
  c()
```

into this
``` nim
proc c() =
  discard
proc d() =
  discard()
proc b() =
  d()
proc a() =
  b()
  c()
```

blackmius: let me try some trick
``` nim
var newDest = createTokenBuf(16)
treProc c, newSest, n
dest.add newDest
```

Araq: It's not a trick, I told you it would work. :P

Araq:
Now it still needs sem.nim patches so that it produces proctype reliably.

PR #1789, opened 2026-04-17
added stdlib threadpool and a test

Araq:


blackmius:
i think "relays" here will be good fit

blackmius:
and it is still fd centric like old apis, try to integrate io_uring

Araq:
@blackmius I'm merging it this way but feel free to iterate over it.

PR #1792, opened 2026-04-17
Bootstrap more tier-1 leaves and fix enum array bounds

planetis-m:
This draft continues the bootstrap process from the previous branch, checking additional low-tier modules against `bin/nimony c`. The branch includes:

- Bootstrapping of `src/lib/platform.nim`, `src/nimony/langmodes.nim`, `src/lib/stringtrees.nim`, and `src/lib/tooldirs.nim`
- Adding regression coverage for enum array-bound issues
- Extending the local `std/os` surface to support proper API usage in `tooldirs`

**Purpose:**
The goal is to reduce the Tier 1 bootstrap backlog and address frontend issues where compiler support is the correct solution, rather than source-level workarounds. A key fix in this branch is support for enum-based constant evaluation in array bounds (e.g., `succ(low(E)) .. high(E)`), which previously failed during constant evaluation. For `tooldirs`, the fix involved adding missing `std/os` functions, respecting Nim 3’s string semantics for C-facing code.

**Impact:**
- Closes additional bootstrap boxes in the Tier 1 plan
- Maintains normal `std/os` usage in `src/lib/tooldirs.nim`
- Reduces a frontend constant-evaluation bug to a covered regression
- Brings `gear2/modnames` closer to compiling, now blocked only by missing `echo` support

**Root Causes Addressed:**
- Incorrect reduction of enum conversions during constant evaluation for array bounds
- Insufficient local `lib/std/os.nim` surface for bootstrap modules relying on basic app-dir and path helpers
- Nim 3 string semantics requiring explicit handling for OS APIs

**Validation:**
- Successfully bootstrapped and tested multiple modules
- No regressions: `0 / 21 bootstrap regressions`


tests/nimony/modules/tinitorder.output
```diff
@@ -1,4 +1,4 @@
-a
 b
+a
```
Araq, line 1: I know my PR does the same but this is a serious regression, module init order is not subject to change.

planetis-m: fixed! I am out for the next 5hours!

PR #1796, opened 2026-04-18
Align nimonyplugins internals with nifcursors style (#1799)

planetis-m:
- Rename `TreePayload` → `NifBuilderObj`, add `NifBuilderOwner` ptr alias matching the `CursorOwnerObj`/`CursorOwner` pattern from nifcursors.
- Rename `counter` → `rc` for the reference count field.
- Restructure `=destroy` to symmetric `dec` + `== 0` check, matching Cursor's destructor.
- Remove unused `NifCursor.owner: NifBuilder` field (never set or read; data stays alive via Cursor's own COW mechanism).
- Wrap `TagId` as an opaque object with `raw` field, aligned with `SymId`.
- Add `withTree` overload without `info` parameter (defaults to NoLineInfo).

💘 Generated with Crush

Assisted-by: GLM-5.1 via Crush <crush@charm.land>

### DIFF
--- a/src/nimony/lib/nimonyplugins.nim
+++ b/src/nimony/lib/nimonyplugins.nim
@@ -66,7 +66,7 @@ proc `=dup`*(x: NifBuilder): NifBuilder {.nodestroy.} =
 
 proc initNifBuilderObj(buf: sink TokenBuf): NifBuilderOwner =
   result = cast[NifBuilderOwner](alloc0(sizeof(NifBuilderObj)))
-  result.rc = 0
+  result.rc = 1
   result.buf = ensureMove(buf)
 
 proc copyBuffer(buf: TokenBuf): TokenBuf =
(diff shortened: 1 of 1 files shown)


src/nimony/lib/nimonyplugins.nim
```diff
@@ -56,17 +56,17 @@ proc `=copy`*(dest: var NifBuilder; src: NifBuilder) =
...
   if result.p != nil:
-    inc result.p.counter
+    inc result.p.rc
 
-proc initPayload(buf: sink TokenBuf): ptr TreePayload =
-  result = cast[ptr TreePayload](alloc0(sizeof(TreePayload)))
-  result.counter = 0
+proc initNifBuilderObj(buf: sink TokenBuf): NifBuilderOwner =
+  result = cast[NifBuilderOwner](alloc0(sizeof(NifBuilderObj)))
+  result.rc = 0
```
planetis-m: 🤔

Araq:
> Add withTree overload without info parameter (defaults to NoLineInfo).

No, not being able to trace back where a node came from isn't a good idea.

planetis-m:
Alright, its now gone! I did not touch validate* procs, I have no idea if they should be updated or removed.

PR #1805, opened 2026-04-21
Cps closure iterators

blackmius:


blackmius:
needs to rewrite after https://github.com/nim-lang/nimony/pull/1811

PR #1816, opened 2026-04-25
Sumtype narrowing, WIP

### LINKED ISSUE #1779: Case objects value access bug
With case objects we can write:
```nim
import std/syncio

type
  MyObj = object
    case k: bool
    of false:
      a: int
    of true:
      b: int

var obj = MyObj(k: false, a: 42)
echo obj.b # prints 42
```
Or
```nim
import std/syncio

type
  MyObj2 = object
    case
    of B1:
      a: int
    of B2:
      b: int

var obj2 = B1(a: 42)
echo obj2.b # prints 42
```

It's invalid and for Nim 2 it is FieldDefect:
```
Hint: /home/asvi/nim/nimony_bug/bugrepro [Exec]
/home/asvi/nim/nimony_bug/bugrepro.nim(10) bugrepro
/home/asvi/.local/share/grabnim/nim-devel/lib/system/fatal.nim(62) sysFatal
Error: unhandled exception: field 'b' is not accessible for type 'MyObj' using 'k = 0' [FieldDefect]
```

In the case of Nimony, this should either be a Defect as well, or as an alternative implementation the compiler must verify that the discriminator is correct by:
- Using control flow analysis (case and if proving selected discriminator value).
- Ensuring the discriminator remains unchanged (I’m not entirely sure, but perhaps this can be achieved through contracts_njvl.nim / contracts.nim)

Otherwise, return an error: failed to verify field `b` access, consider add `if obj.k`

ASVIEST:
Fixes https://github.com/nim-lang/nimony/issues/1779
It also implements `{.cast(uncheckedAssign).}` that disables checks added by this PR

### DIFF
--- a/src/hexer/lifter.nim
+++ b/src/hexer/lifter.nim
@@ -792,17 +792,24 @@ proc genProcDecl(c: var LiftingCtx; sym: SymId; typ: TypeCursor) =
     let bodyStart = c.dest.len
     copyIntoKind(c.dest, StmtsS, c.info):
       maybeAddResultDecl c, paramA, typ
-      let beforeUnravel = c.dest.len
-      if a.typeKind == RefT:
-        unravelRef(c, a, paramTreeA, paramTreeB)
-      else:
-        unravelDispatch(c, typ, paramTreeA, paramTreeB)
-        if c.dest.len == beforeUnravel:
-          var t = typ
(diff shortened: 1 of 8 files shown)


Araq:
Outdated and adds a feature without any discussion about it.

PR #1827, opened 2026-04-27
Compiler speedup

ASVIEST:
nifmake used to use a DAG depth based approach that each layer blocks, the problem is that DAG for compiler is quite close to linear, so it can't get into the far corners (i.e. C code generation) early, in fact compiling all the C code will happen at about the same time at the end, so waiting for C compilation can't be hidden behind other processes, resulting in big slowdown. It's main problem, other (smaller) problem was with https://github.com/nim-lang/nimony/commit/cb0f724, which led to unexpected copies, claude analyzed callgrind output and found proc that frequently triggered a shrink, which caused a COW. This was the commonType function. The obvious fix would be to implement freeze and thaw operations, but they would be 'unsafe' so compiler now handles it this via `=wasMoved` in `endRead` instead

bench results:
After this PR:
img
Before this PR:
screenshot
i.e 3.1x speedup

### DIFF
--- a/src/lib/nifcursors.nim
+++ b/src/lib/nifcursors.nim
@@ -228,11 +228,15 @@ proc prepareMutation*(b: var TokenBuf) {.inline.} =
   ## existing cursors. Releases the TokenBuf's rc ref to the old owner;
   ## if no cursors remain (rc was 1), cleans up the old owner entirely.
   if b.owner != nil:
-    let newData = cast[Storage](alloc(sizeof(PackedToken) * b.cap))
-    copyMem(newData, b.data, sizeof(PackedToken) * b.len)
-    decRcAndFree(b.owner)
-    b.owner = nil
-    b.data = newData
+    if b.owner.rc == 1:
+      dealloc(b.owner)
+      b.owner = nil
+    else:
+      let newData = cast[Storage](alloc(sizeof(PackedToken) * b.cap))
+      copyMem(newData, b.data, sizeof(PackedToken) * b.len)
+      decRcAndFree(b.owner)
+      b.owner = nil
+      b.data = newData
 
 proc freeze*(b: var TokenBuf) {.inline.} = discard
 proc thaw*(b: var TokenBuf) {.inline.} = discard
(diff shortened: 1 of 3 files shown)


src/lib/nifcursors.nim
```diff
@@ -228,11 +228,15 @@ proc prepareMutation*(b: var TokenBuf) {.inline.} =
...
   ## if no cursors remain (rc was 1), cleans up the old owner entirely.
   if b.owner != nil:
-    let newData = cast[Storage](alloc(sizeof(PackedToken) * b.cap))
-    copyMem(newData, b.data, sizeof(PackedToken) * b.len)
-    decRcAndFree(b.owner)
-    b.owner = nil
-    b.data = newData
+    if b.owner.rc == 1:
+      dealloc(b.owner)
+      b.owner = nil
```
Araq, line 233: I don't understand this logic.

Araq:
> The obvious fix would be to implement freeze and thaw operations, but they would be 'unsafe' so compiler now handles it this via =wasMoved in endRead instead

No, please bring back freeze and thaw.

src/nifmake/nifmake.nim
```diff
@@ -310,8 +313,131 @@ proc recordCmdTime(profile: var ProfileData; cmdName: string; sec: float) =
...
+
+proc `==`(a, b: WorkerId): bool {.borrow.}
+
+proc steal(p: var WorkerPool, fromWorkerId: WorkerId): TaskId =
+  result = TaskId(-1)
+  for id in p.workerIds:
+    if id != fromWorkerId and p.wokerDeques[id.int].len > 0:
+      return p.wokerDeques[id.int].popLast()
+
+proc eval(s: var Scheduler; dag: var Dag; dependents: seq[seq[int]];
```
Araq, line 346: `seq[seq[int]]` -- what is this shit?

src/nifmake/nifmake.nim
```diff
@@ -322,54 +448,45 @@ proc runDag(dag: var Dag; opt: set[CliOption]; profile: ptr ProfileData = nil):
...
-        proc beforeRunEvent(idx: int) =
-          progress[idx] = Running
-          if profile != nil: startTimes[idx] = getMonoTime()
-
-        proc afterRunEvent(idx: int; p: Process) =
-          progress[idx] = Finished
-          if profile != nil:
-            let sec = toSeconds(getMonoTime() - startTimes[idx])
-            profile[].recordCmdTime(cmdNames[idx], sec)
+    # Work-stealing scheduler: one worker owns one process.
```
Araq, line 451: Well but the depth-based traversal was obviously correct and this work-stealing scheduler of yours is incomprehensible.

Araq:
Never mind, found better solutions for the problems you found. Thanks!

PR #1834, opened 2026-04-29
LLVM backend stub wrappers for C macros

bridgeQiao:
## Summary

The LLVM backend cannot directly reference C macros as symbols. C macros include but are not limited to: variable-like macros (e.g. `stderr`, `stdout`, which on some platforms expand to internal symbols like     
 `__stderrp` rather than linkable global variables) and constant-like macros (e.g. `_IOFBF`, `EOF`). Referencing `@stderr` directly in LLVM IR produces undefined symbols.                                               
                                                                               
 nifc LLVM codegen now detects `importc + header` global variable declarations and generates a C wrapper function for each (e.g. `module_get_stderr()`), compiled into `module.wrap.o` and linked into the final binary.

### Why this lives in nifc

 C macros being non-representable as LLVM symbols is a backend compatibility issue. nifc, as the backend code generator, encapsulates this platform difference — it only exposes the `module.wrap.c` artifact filename
 to the upper nifler layer (deps.nim / nifmake), without leaking any C FFI semantics.

### Why empty .wrap.c files exist

 This minimizes the surface area of the change. `generateStubFile` function is called for every module and always writes a file. `deps.nim` unconditionally generates cc rules and link inputs for `module.wrap.c`. (Maybe could parse at nif's metadata? I will check it :)

## Changes

 `src/nifc/llvmcodegen.nim` — Detects importc + header global variables, collects StubWrapper entries, generates .wrap.c. Emits `declare ptr @module_get_xxx()` for wrapped symbols.

 `src/nifc/llvmgenexprs.nim` — In genExprLLVM Symbol branch: if symbol is in wrappedSyms, generates `call ptr @wrapper()` instead of `load ptr, ptr @symbol`.

 `src/nimony/deps.nim` — Unconditionally generates cc rules for .wrap.c, adds .wrap.o to link inputs, declares .wrap.c as an additional output of the nifc step.

### Example generated files

 .ll diff (using stderr as example):

```llvm
 -@stderr = external global ptr
 +declare ptr @module_get_stderr()
  ...
 -  %t1 = load ptr, ptr @stderr
 +  %t1 = call ptr @module_get_stderr()
```

 module.wrap.c:

```c
 // Auto-generated stub wrappers for C macros
 #include <stdio.h>

 FILE* module_get_stderr(void) {
   return stderr;
 }
```

 final.build.nif diff:

```lisp
  (do cc
    (input "nimcache/module.c")
    (output "nimcache/module.o"))
 +(do cc
 +  (input "nimcache/module.wrap.c")
 +  (output "nimcache/module.wrap.o"))
  (do nifc
    (input "nimcache/module.c.nif")
    (output "nimcache/module.ll")
 +  (output "nimcache/module.wrap.c"))
```

 Link step gains .wrap.o input:

```lisp
  (do link
    (input "nimcache/module.o")
 +  (input "nimcache/module.wrap.o")
    (output "nimcache/module"))
```

### DIFF
--- a/src/nifc/llvmcodegen.nim
+++ b/src/nifc/llvmcodegen.nim
@@ -124,6 +124,7 @@ type
     symName: string
     wrapperName: string
     header: string
+    cType: string
 
   LLVMGenFlag* = enum
     gfMainModule
@@ -189,6 +190,36 @@ proc error(m: MainModule; msg: string; n: Cursor) {.noreturn.} =
   when defined(debug):
     echo getStackTrace()
   quit 1
+
+proc computeCType(c: var LLVMCode; typ: Cursor): string =
+  var t = typ
+  case t.typeKind
+  of PtrT, AptrT:
+    inc t
+    if t.typeKind == VoidT:
+      return "void*"
+    if t.kind == Symbol:
+      let decl = c.m.getDeclOrNil(t.symId)
+      if decl != nil and decl.extern != StrId(0):
+        return pool.strings[decl.extern] & "*"
+    return "void*"
+  of IT, UT, FT:
+    result = case t.typeKind
+             of IT: "int"
+             of UT: "unsigned int"
+             else: "double"
+    inc t
+    if t.kind == IntLit: inc t
+    while t.kind != ParRi:
+      if t.pragmaKind == ImportcP:
+        inc t
+        if t.kind == StringLit:
+          result = pool.strings[t.litId]
+          break
+        skip t
+      else: skip t
+  else:
+    return "void*"
 
 # ---- Token helpers ----
 
@@ -504,15 +535,16 @@ proc genGlobalVarDeclLLVM(c: var LLVMCode; n: var Cursor; vk: VarKindLLVM; toExt
       # C macros (e.g. stderr, stdout) cannot be referenced as LLVM symbols;
       # generate a wrapper function that returns the macro's value.
       let wrapperName = c.stubModuleName & "_get_" & name
-      c.stubWrappers.add StubWrapper(
-        symName: name,
-        wrapperName: wrapperName,
-        header: headerStr
-      )
       c.wrappedSyms[lit] = wrapperName
       if wrapperName notin c.declaredExterns:
         c.declaredExterns.incl wrapperName
         c.addTo(c.externs, "declare " & typ & " @" & wrapperName & "()\n")
+        c.stubWrappers.add StubWrapper(
+          symName: name,
+          wrapperName: wrapperName,
+          header: headerStr,
+          cType: computeCType(c, d.typ)
+        )
     elif toExtern or isImport:
       c.addTo(c.globals, "@" & name & " = external " & tls & "global " & typ & alignSuffix & "\n")
     else:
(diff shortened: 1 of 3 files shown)


bridgeQiao:
test file can use `hello.nim`:
```c
import std / [syncio, assertions]

# importc + header tests for wrap.c generation

proc printf(format: cstring): cint {.importc: "printf", header: "<stdio.h>", varargs.}

var stderr {.importc: "stderr", header: "<stdio.h>".}: pointer
var stdout {.importc: "stdout", header: "<stdio.h>".}: pointer
var errno {.importc: "errno", header: "errno.h".}: cint

var IOFBF {.importc: "_IOFBF", header: "<stdio.h>".}: cint
var O_RDONLY {.importc: "O_RDONLY", header: "<fcntl.h>".}: cint

when isMainModule:
  echo "hello, world"
  discard printf("printf: %d\n".cstring, 42)
  echo "stderr ptr: ", cast[int](stderr)
  echo "_IOFBF = ", IOFBF
  echo "O_RDONLY = ", O_RDONLY
  echo "errno = ", errno
```

do `./bin/nimony l -r hello.nim`, will generate `nimcache/helg27y181/helg27y181.wrap.c`:
```c
// Auto-generated stub wrappers for C macros
#include <stdio.h>
#include "errno.h"
#include <fcntl.h>

void* helg27y181_get_stderr(void) {
  return (void*)stderr;
}

int helg27y181_get_errno(void) {
  return errno;
}

int helg27y181_get__IOFBF(void) {
  return _IOFBF;
}

int helg27y181_get_O_RDONLY(void) {
  return O_RDONLY;
}

FILE* helg27y181_get_stdout(void) {
  return stdout;
}
```
I notice the `printf` is `1` which c backend is correct `42`. But it's not the PR range.

Araq:
Interesting approach but I prefer something like https://github.com/nim-lang/Nim/blob/devel/tools/detect/detect.nim

bridgeQiao:
> Interesting approach but I prefer something like https://github.com/nim-lang/Nim/blob/devel/tools/detect/detect.nim

Thanks for the feedback!

Using `cc -E` to extract the actual symbols for direct replacement is indeed a very clean approach.

I actually experimented with symbol resolution previously, but I considered it only half-successful. In that attempt, I invoked `cc -E` to parse the symbols dynamically for every single `module.c.nif` file. Caching the resolved symbols globally should be right.

bridgeQiao:
Using the parsed symbol mapping turned out to be more complicated than I expected.
Here's my current high-level idea (**Only idea**, demo didn't work well):

1. add `--detect` option to `nifc`;
  1.1 Use `clang -E` to obtain the symbol mapping. (`#define NIMONY_PROBE(id, expr) #id | expr`, but only clang works)
  1.2 Use `printf` (or similar) to get the remaining enum constant symbols. (use `__builtin_constant_p(id)` and `id == expr`)
2. Generate the final artifact `symconsts.txt`, which will serve as an `input dependency` for nifc at `module.final.build.nif`;
3. During LLVM backend code generation, perform symbol replacement;
  3.1 `Constants` (e.g. `_IOFBF`) are replaced directly with their values.
  3.2 `External symbols` and `Global variable` (e.g. `printf` `stderr`) are replaced by name.
  3.3 `Function calls` (e.g. `errno`) are converted to standard call instructions.

By the way, although the `stub wrapper functions` doesn’t look very clean, it is indeed simple and does not rely on any additional compiler for preprocessing or compilation steps.

Araq:
Just hardcode the values, take them from Rust where they can seen without having to run a preprocessor. FreePascal also has them.

PR #1835, opened 2026-04-29
try to speed up CIs with more caching

Araq:


Araq:
No benefit, 68m. Let's try it one more time.

PR #1839, opened 2026-04-30
Contracts: static cycle preventions

Araq:


doc/nocycles.md
```diff
@@ -0,0 +1,351 @@
...
+- **Documented escape valves** (`.cursor` and `requires: disjoint`)
+  that map onto language features that already exist.
+
+## What This Does Not Buy
+
+- Doubly-linked rings and graphs are not expressible under the rule
+  with all-owning fields. Either:
+  - mark the back-edge field `.cursor` and arrange for the container
+    to keep the nodes alive some other way, or
+  - keep the type as a cycle-collected `ref` — these types opt into a
```
blackmius, line 338: what is cycle-collected ref? i mean is it something else than just ref? can we have an example of double linked ring?

Araq: It's stuff the AI made up.

Araq:
Too early for this. Also should be a plugin.

PR #1842, opened 2026-05-02
dagon: better doc generator; WIP

Araq:


ZoomRmc:
Can we have a tool that's not so tightly coupled with html as Nim2 nimdoc?

I have a vibe-coded prototype that emits docs in the original md+rst format from nif+original code and it's already pretty useful. However, it's buggy and probably stale (a few months old).

As far as I understand, we basically have two kinds of documentation: free-standing doc-comments, and those tied to symbol declarations. Emitting those in a structured way and then having an ability to pass those to backends for html/plain text/json/original md+rst would be great.

Current coupling with html is to simplify referencing, right?

Araq:
The docgen will have support for both NIF output and HTML output.

I don't know what else to generate instead, JSON is just terrible for it as it's not a document format and every other tree-like format is just like HTML, I don't really understand the "tight coupling with HTML" argument, you can parse the HTML and extract the tree structure. Nim even ships with an HTML parser in its stdlib...

ZoomRmc:
> I don't know what else to generate instead.

What I want is to be able to easily convert .nim to .rst, inverting relations between docs and code. Removing everything except documentation comments and putting declarations inside code blocks.

src/dagon/dagon.nim
```diff
@@ -0,0 +1,725 @@
...
+  let t = asTypeDecl(n)
+  recordEntry(idx, TypeS, t.name.symId, doc)
+  emitDeclItem(b, t.name.symId):
+    emitTag(b, "code"):
+      emitKw(b, "type")
+      emitText(b, " ")
+      emitName(b, basename(pool.syms[t.name.symId]))
+      if t.body.kind == ParLe:
+        emitOp(b, " = ")
+        emitKw(b, pool.tags[t.body.tagId])
```
blackmius, line 340: it would be cool if doc blocks could be under each exported field in the type

PR #1851, opened 2026-05-08
runtime: strings use an OOM cookie for easier detection

Araq:


lib/std/system/stringimpl.nim
```diff
@@ -16,6 +16,17 @@ const
 
 const LongStringDataOffset = 3 * sizeof(int)  ## byte offset of LongString.data from start
 
+## OOM cookie: "\nD^OOM\0" packed inline (slen=7). Detect with `isOom(s)`.
```
Araq:
```suggestion
# OOM cookie: "\nD^OOM\0" packed inline (slen=7). Detect with `isOom(s)`.
```

PR #1860, opened 2026-05-15
.passive procs and .closure iterators: progress

Araq:


tests/nimony/iter/tmulti_yield_nested.nim
```diff
@@ -0,0 +1,29 @@
...
+iterator pair2(x: int): int =
+  if x > 0:
+    yield 2
+  else:
+    for k in 0 ..< x:
+      yield k
+
+proc total(): int =
+  var acc = 0
+  for a in pair(3):
```
blackmius, line 23: test how closure iterator will work inside another closures iterator

PR #1875, opened 2026-05-21
LLVM backend update virtualParRi feature

bridgeQiao:
- Replace manual `skip parRi` logic with `into`/`loopInto` patterns.
- Verify generated `.ll` files against commit 3466cd248277bec4c1830a5401e64cdf13b26312. 

The `examples/*.nim` generated `.ll` content is identical, with the exception of minor changes in the lib library.


bridgeQiao:
test **variant objects**, the generated `.ll` is same:

```nim
type
  ShapeKind* = enum skCircle, skRect
  Shape* = object
    case kind*: ShapeKind
    of skCircle:
      radius*: float64
    of skRect:
      width*, height*: float64
```

PR #1892, opened 2026-05-25
Fix crash when passing module as argument (nim-lang/nimony#1891)

### LINKED ISSUE #1891: Crash in error reporting when passing module to function
This produces a crash instead of a proper error message:
```nim
import std/[syncio]
echo syncio
```

```
PS C:\dev\nimony> nimony check .\tests\nimony\calls\tcall_module_arg.nim
C:\dev\nimony\src\nimony\nimsem.nim(169) nimsem
C:\dev\nimony\src\nimony\nimsem.nim(150) handleCmdLine
C:\dev\nimony\src\nimony\nimsem.nim(70) processModules
C:\dev\nimony\src\nimony\sem.nim(7442) semcheck
C:\dev\nimony\src\nimony\sem.nim(7217) semcheckCore
C:\dev\nimony\src\nimony\sem.nim(7004) phase3
C:\dev\nimony\src\nimony\sem.nim(6981) semToplevelStmts
C:\dev\nimony\src\nimony\sem.nim(739) semStmt
C:\dev\nimony\src\nimony\sem.nim(6644) semExpr
C:\dev\nimony\src\nimony\semcall.nim(1225) semCall
C:\dev\nimony\src\nimony\semcall.nim(941) resolveOverloads
C:\dev\nimony\src\nimony\semcall.nim(145) semTemplateCall
C:\dev\nimony\src\nimony\sem.nim(6513) semExpr
C:\dev\nimony\src\nimony\sem.nim(725) semStmtsExpr
C:\dev\nimony\src\nimony\sem.nim(739) semStmt
C:\dev\nimony\src\nimony\sem.nim(6742) semExpr
C:\dev\nimony\src\nimony\sem.nim(727) semStmtsExpr
C:\dev\nimony\src\nimony\sem.nim(6644) semExpr
C:\dev\nimony\src\nimony\semcall.nim(1225) semCall
C:\dev\nimony\src\nimony\semcall.nim(1037) resolveOverloads
C:\dev\nimony\src\nimony\sigmatch.nim(209) addErrorMsg
C:\dev\nimony\src\nimony\sigmatch.nim(158) getErrorMsg
C:\dev\nimony\src\nimony\renderer.nim(2128) typeToString
C:\dev\nimony\src\nimony\renderer.nim(2078) renderTree
C:\dev\nimony\src\lib\nifcursors.nim(817) toString
C:\dev\nimony\src\lib\nifcursors.nim(739) span
C:\dev\nimony\src\lib\nifcursors.nim(152) kind
C:\dev\nimony\src\lib\nifcursors.nim(149) load
C:\dev\Nim\lib\std\assertions.nim(41) failedAssertImpl
C:\dev\Nim\lib\std\assertions.nim(34) raiseAssert
Error: unhandled exception: C:\dev\nimony\src\lib\nifcursors.nim(149, 3) `c.p != nil and c.rem > 0`  [AssertionDefect]
FAILURE: C:\dev\nimony\bin\nifmake.exe --verbose --base:.\tests\nimony\calls -j run nimcache\tca6ryqzw1.build.nif
```

I'm not sure what the best way to fix this would be, maybe `proc gtype` in renderer.nim should handle the module case (although that doesn't really make sense in this case as there is no module type afaik, and probably doesn't handle other similar crashes).

Maybe `proc renderTree` should just handle the case where n is at the end:
```nim
proc renderTree(n: Cursor, renderFlags: RenderFlags = {}, renderType = false): string =
  var g: SrcGen = initSrcGen(renderFlags)
  var n = n
  if renderType:
    var c: Context = initContext()
    gtype(g, n, c)
  else:
    gsub(g, n, isTopLevel = true)
  result = g.buf
  if result.len == 0:
    result = toString(n, false) # here n can be at the end, leading to a crash
```

or maybe this should handle the cursor at end case:
```nim
proc span*(c: Cursor): int =
  result = 0
  var c = c
  if c.kind == ParLe: # <--- crash when c at end, maybe just return 0?
```

Nimaoth:
fixes #1891

### DIFF
--- a/src/nimony/renderer.nim
+++ b/src/nimony/renderer.nim
@@ -2067,6 +2067,7 @@ proc gsub(g: var SrcGen; n: var Cursor, fromStmtList = false, isTopLevel = false
 
 proc renderTree(n: Cursor, renderFlags: RenderFlags = {}, renderType = false): string =
   var g: SrcGen = initSrcGen(renderFlags)
+  let orig = n
   var n = n
   if renderType:
     var c: Context = initContext()
@@ -2075,7 +2076,7 @@ proc renderTree(n: Cursor, renderFlags: RenderFlags = {}, renderType = false): s
     gsub(g, n, isTopLevel = true)
   result = g.buf
   if result.len == 0:
-    result = toString(n, false)
+    result = toString(orig, false)
 
 proc asNimCode*(n: Cursor; renderFlags: RenderFlags = {}): string =
   var m0: PackedLineInfo = NoLineInfo
--- a/tests/nimony/calls/tcall_module_arg.msgs
+++ b/tests/nimony/calls/tcall_module_arg.msgs
@@ -1,0 +1,12 @@
+tests/nimony/calls/tcall_module_arg.nim(5, 1) Trace: instantiation from here
+lib/std/syncio.nim(144, 5) Error: Type mismatch at [position]
+write stdout, syncio
+[2] expected: string but got: (module) (declared in lib/std/syncio.nim(48, 1))
+[2] expected: bool but got: (module) (declared in lib/std/syncio.nim(51, 1))
+[2] expected: int64 but got: (module) (declared in lib/std/syncio.nim(55, 1))
+[2] expected: int32 but got: (module) (declared in lib/std/syncio.nim(58, 1))
+[2] expected: uint64 but got: (module) (declared in lib/std/syncio.nim(61, 1))
+[2] expected: uint32 but got: (module) (declared in lib/std/syncio.nim(64, 1))
+[2] (module) does not match constraint T (declared in lib/std/syncio.nim(67, 1))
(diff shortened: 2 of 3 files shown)


src/nimony/renderer.nim
```diff
@@ -2067,6 +2067,7 @@ proc gsub(g: var SrcGen; n: var Cursor, fromStmtList = false, isTopLevel = false
 
 proc renderTree(n: Cursor, renderFlags: RenderFlags = {}, renderType = false): string =
   var g: SrcGen = initSrcGen(renderFlags)
+  let orig = n
   var n = n
```
bridgeQiao, line 2071: should change this shadow `n` be better?

Araq: Nah, it's fine. We generally prefer shadowing over "oops, used the wrong variable". This can backfire as seen in this bug, but there is no universal rule.

PR #1900, opened 2026-05-27
WIP: tree based optimizers for NIFC

Araq:


src/nifc/opt_unused/cse.nim
```diff
@@ -0,0 +1,804 @@
...
+## intervening call or write through an aliased pointer can change the
+## value at that location, forcing us to recompute. Caching the
+## **address** instead is robust — the address of `x.field[i]` is a
+## function only of `x`'s storage location and `i`'s value, neither of
+## which a foreign call can change unless their addresses have escaped.
+##
+## So we rewrite:
+##
+## ```
+## y = x.field[i].fieldB
```
blackmius, line 23: this is true only for single threaded code

Araq: No, addresses on the stack are stable even for multi-threaded code.

blackmius: i dont see guards that the variable is from local stack 

```nim
proc test(x: var seq[ref SomeType]) =
  var y = x[0].field
  y = x[0].field
```

Araq: Well it's "WIP", there is all sort of stuff missing and wrong right now.

PR #1902, opened 2026-05-28
llvm backend: full examples/*.nim support on x86_64-linux & arm64-apple independently

bridgeQiao:
all `exampls/*.nim` works at x86_64-linux and arm64-apple (each platform tested independently, not cross-compilation/cross-platform)


bridgeQiao:
Known bug: On the arm64-apple platform, clang -O1 misoptimizes sext when used with literal values. fix in progress (hard game).

bridgeQiao:
It seems the original IT type also has this issue, for example `echo "test: ", int(2323'i32)`, generate:

```llvm
%t55 = load ptr, ptr @__stdoutp
%t56 = trunc i64 2323 to i32
%t57 = sext i32 %t56 to i64
call void @write_2_syn1lfpjv(ptr %t55, i64 %t57)
```

the output: 
```text
test: 8397799680
```

Therefore this path doesn't work: `UT i64 -> trunc -> sext`. Maybe need a new PR to fix it :)

src/nifc/cimportsyms.nim
```diff
@@ -0,0 +1,462 @@
...
+      constVal*: string
+    of imkCall:
+      callFuncName*: string
+
+template S(v: string): ImportMacroEntry = ImportMacroEntry(kind: imkReplace, resolvedName: v)
+template I(v: string): ImportMacroEntry = ImportMacroEntry(kind: imkConstant, constVal: v)
+template C(v: string): ImportMacroEntry = ImportMacroEntry(kind: imkCall, callFuncName: v)
+
+# macOS import map
+const macOSImportMap* = block:
```
Araq: These import maps are not llvmgen's job. Fix it in Nimony's stdlib instead, that's why it diverged from Nim's. This might also imply to fix the related errorcodes project. So be it.

bridgeQiao: I moved the symbol map into the stdlib, more specifically `syncio.nim` and `panics.nim`, but `tests/nimony/ffi/tpacked.nim` generates different C code on different platforms. 

`./src/hastur test` lacks the ability to compare results across different platforms.

Araq: Simplify the test so that it doesn't check the produced C code, only the outputs of the program.

bridgeQiao:
When calling variadic functions in LLVM IR, the `...` signature must be explicitly specified. That's mean:

`call void @fprintf(ptr %t0, ptr @.str.0, i64 %t1)`

should be:

`call void (ptr, ptr, ...) @fprintf(ptr %t0, ptr @.str.0, i64 %t1)`

PR #1909, opened 2026-06-01
Fixed compile-time `is` for union types and generic bodies

gmpreussner:
While union-type matching in `sigmath.nim` was largely fine, there were issues with `semIs` that rejected successful matches, causing tests like `int8 is SomeInteger` to fail. I added unit tests that also cover various types of enums.

More details in https://github.com/nim-lang/nimony/pull/1909/changes/aa6c9ddc39db77c5285b396a2b2efcd4dc4011df

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -6292,7 +6292,7 @@ proc semIs(c: var SemContext; dest: var TokenBuf; it: var Item) =
     rhs = semLocalType(c, dest, it.n)
   skipParRi it.n
   dest.shrink beforeExpr # delete LHS and RHS
-  if c.routine.inGeneric > 0 or containsGenericParams(lhs.typ) or containsGenericParams(rhs):
+  if containsGenericParams(lhs.typ) or containsGenericParams(rhs):
     # Keep the unpreprocessed operand expressions so template/generic
     # instantiation can substitute formals and re-run `semIs`. Defer whenever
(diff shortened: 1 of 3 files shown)


src/nimony/sigmatch.nim
```diff
@@ -1397,8 +1397,16 @@ proc singleArgImpl(m: var Match; f: var Cursor; arg: CallArg) =
         if not matched:
           m.error InvalidMatch, f, arg.typ
         skip f
+    of TypekindT, OrdinalT, ConceptT, SymkindT:
```
Araq: Too dangerous to do it here in general, this should be in `semIs`'s logic.

src/nimony/sem.nim
```diff
@@ -6288,15 +6292,25 @@ proc semIs(c: var SemContext; dest: var TokenBuf; it: var Item) =
     rhs = semLocalType(c, dest, it.n)
   skipParRi it.n
   dest.shrink beforeExpr # delete LHS and RHS
-  if containsGenericParams(lhs.typ) or containsGenericParams(rhs):
+  if c.routine.inGeneric > 0 or containsGenericParams(lhs.typ) or containsGenericParams(rhs):
```
Araq:
```suggestion
  if containsGenericParams(lhs.typ) or containsGenericParams(rhs):
```

PR #1915, opened 2026-06-04
WIP: Added check for proc redefinitions

gmpreussner:
Nimony treats proc names as overloadable but did not yet check for existing overloads with the same signature.


src/nimony/semdecls.nim
```diff
@@ -82,6 +82,51 @@ proc findMatchingForwardDecl*(c: var SemContext; symId: SymId; implDecl: Cursor)
...
+    else: discard
+    inc pos
+
+proc checkRoutineRedefinition(c: var SemContext; dest: var TokenBuf; declStart: int;
+                              symId: SymId; kind: SymKind; hasBody: bool; info: PackedLineInfo): string =
+  ## Reject a routine declaration whose signature matches an existing overload,
+  ## mirroring Nim's `searchForProc` / `wrongRedefinition` (see procfind.nim).
+  ## Forward declarations with empty bodies are exempt.
+  result = ""
+  if kind notin RoutineKinds: return
```
Araq: Why would that be the case? Stupid check, remove it.

src/nimony/semdecls.nim
```diff
@@ -82,6 +82,51 @@ proc findMatchingForwardDecl*(c: var SemContext; symId: SymId; implDecl: Cursor)
...
+  let lit = symToIdent(symId)
+  var newBuf = copyRoutineDeclAt(dest, declStart)
```
Araq: Why would that be required?

src/nimony/semdecls.nim
```diff
@@ -82,6 +82,51 @@ proc findMatchingForwardDecl*(c: var SemContext; symId: SymId; implDecl: Cursor)
...
+  var newDecl = beginRead(newBuf)
+  try:
+    var scope = c.currentScope.up
+    if scope == nil: scope = c.currentScope
```
Araq: What kind of scope traversal is this?

src/nimony/semdecls.nim
```diff
@@ -82,6 +82,51 @@ proc findMatchingForwardDecl*(c: var SemContext; symId: SymId; implDecl: Cursor)
...
+    let ignoreStyle = IgnoreStyleFeature in c.features
+    while scope != nil:
+      for k in stylesOfScope(scope, lit, ignoreStyle):
+        for sym in scope.tab.getOrDefault(k):
+          if sym.name == symId or sym.kind != kind: continue
+          let loaded = tryLoadSym(sym.name)
+          if loaded.status != LacksNothing: continue
+          if not signaturesMatch(loaded.decl, newDecl): continue
```
Araq: All very expensive. We could compute a canonical string from a proc signature instead and simply use a hash table here.

src/nimony/semdecls.nim
```diff
@@ -886,7 +931,17 @@ proc semProcImpl(c: var SemContext; dest: var TokenBuf; it: var Item; kind: SymK
...
+        let redefMsg = checkRoutineRedefinition(c, dest, declStart, symId, kind, hasBody, info)
+        if redefMsg.len > 0:
+          dest.addParLe(StmtsS, info)
+          dest.buildTree ErrT, info:
+            dest.add dotToken(info)
+            dest.add strToken(pool.strings.getOrIncl(redefMsg), info)
+          dest.addParRi()
+          skip it.n
+        else:
+          dest.takeTree it.n
```
Araq: This part here, however, is as good as it gets. Good job!

src/nimony/semdecls.nim
gmpreussner: Defensive programming in case someone calls `checkRoutineRedefinition` for something other than routines, but I agree, since this proc is private and we control the flow inside this module, the check is redundant. I will remove it.

src/nimony/semdecls.nim
gmpreussner: Hmm... I agree, but it might require a slightly more invasive change. I couldn't find any existing helpers that would be a good fit. Would it be OK to track the canonical strings more generally in a field of `Scope` in `semtabs.nim`? We could then recompute the hashes as scopes are entered and exited. Could forward declaration checks benefit from this as well? For the duplicate detection we only care about the routine name and its shape - I think this is true for forward decls as well?

src/nimony/semdecls.nim
gmpreussner: When this runs, the compiler is inside the parameter scope opened for the routine being declared. `currentScope` is for the locals only, `currentScope.up` is where overloaded routine names live. We want to check the overload sets in the outer scopes, not the parameter locals. If no parameter scope was opened, or we are at the outmost scope, there is no parent and we fall back to the scope that holds the symbol table.

Again, this was defensive programming, but since the method is private and we can control the call sites, I will remove the check.

src/nimony/semdecls.nim
gmpreussner: Another option would be to keep a signature hash cache in `SemContext`, then add a few helper functions in `semdecl.nim`. I think I would prefer that from an esthetic point of view. It would be minimally invasive and self-contained without touching existing features.

gmpreussner: Updated the code with the second idea.

gmpreussner:
I'm gonna take another look at this one; not happy with the amount of changes in this PR.

gmpreussner:
@Araq I updated this PR for latest Nimony and rewrote the lookup. This new implementation looks a lot better than the previous one and is more compact as well, but you'll have to check if it is sound. I also removed the lookup cache as it was useless for typical code, required extra state in `SemContext`, and made the implementation more complex.

I'm now doing the following:
1. Walk the scope chain
2. For each scope, check if identifier for the new definition already exists
3. For each candidate, load the prior decl and check if the signatures match

Signatures match if:
- typevars are the same (also considers type constraints)
- param types are the same
- return types are the same

Parameter names are irrelevant, and types are compared exactly so that distinct instantiations remain valid overloads.

Computational complexity should be something like `O(d × s × k × |sig|)` with:
- `d` = scope depth
- `s` = style variants (usually 1)
- `k` = overload count for that basename in that scope
- `|sig|` = size of the signature trees being compared

The vast majority of lookups are rejected immediately by identifier mismatch.

It looks like this is pretty much the same shape as in Nim's `searchForProcAux` in `procfind.nim`

gmpreussner:
After adding support for generic procs this got pretty messy again. I expected to be able to reuse existing functionality in Nimony instead of implementing a bunch of new helper functions, but it looks like nothing in overload resolution quite fits the requirements for redefinition checks.

Araq:
Make it so that when we add the symbol to the symbol table it reports back with `HasExisting` or something like that so that the first entry does not have to pay for all this machinery.

gmpreussner:
This is way outdated now; should revisit with the latest Nimony version.

PR #1919, opened 2026-06-05
Automatic widening for int and float types 

gmpreussner:
Built-in integer and float types were not automatically widened yet. For example, `echo 5'i8` would fail because the `write` proc is defined for `int`, not `int8`. With this change, integer and float types will be widened to the best match, if necessary.


Araq:
This can be simplified further.

src/nimony/sigmatch.nim
```diff
@@ -984,6 +984,24 @@ proc matchSymbol(m: var Match; f: Cursor; arg: CallArg) =
...
+  else: return false
+  var p = plat
+  inc p
+  cmpTypeBits(context, bits, p) == 0
+
+proc incIntegralWidenCost(m: var Match; kind: TypeKind; bits: Cursor; intLit = false) =
+  if isPlatformNumeric(m.context, kind, bits):
+    inc m.intConvCosts
+  elif intLit and kind in {IntT, UIntT}:
+    inc m.intLitCosts
```
Araq:
```suggestion
  if intLit and kind in {IntT, UIntT}:
    inc m.intLitCosts
  elif isPlatformNumeric(m.context, kind, bits):
    inc m.intConvCosts
```

Issue #1922, opened 2026-06-05
issues with generics


blackmius:
# creating new sequence of generic procs

```nim
type A[T] = object
  v: seq[proc(x: T)]
proc a[T](): A[T] = A[T](v: @[])
discard a[int]()
```

```sh
daniil@ubuntu24:~/projects/nimony$ ./bin/nimony c -r test.nim
test.nim(4, 15) Trace: instantiation from here
lib/std/system/seqimpl.nim(189, 18) Error: type mismatch: got: seq but wanted: seq
nifmake: /home/daniil/projects/nimony/bin/nimsem m --isMain nimcache/teszpno6r1.p.nif
FAILURE: /home/daniil/projects/nimony/bin/nifmake --base:'' -j run nimcache/teszpno6r1.build.nif
```

# iterating over sequence of procs

```nim
type A = object
  v: seq[proc(x: int)]
proc a(x: A) =
  for f in x.v:
    f(10)
```

```sh
daniil@ubuntu24:~/projects/nimony$ ./bin/nimony c -r test.nim
test.nim(5, 5) Error: attempt to call routine: 'f'
test.nim(5, 6) Error: undeclared identifier: 'f'
nifmake: /home/daniil/projects/nimony/bin/nimsem m --isMain nimcache/teszpno6r1.p.nif
FAILURE: /home/danill/projects/nimony/bin/nifmake --base:'' -j run nimcache/teszpno6r1.build.nif
```

# creating sequence of procs

```nim
var a: seq[proc(x: int)] = @[]
```
```sh
daniil@ubuntu24:~/projects/nimony$ ./bin/nimony c -r test.nim
Traceback (most recent call last)
/home/daniil/projects/nimony/src/nimony/nimsem.nim(169) nimsem
/home/daniil/projects/nimony/src/nimony/nimsem.nim(150) handleCmdLine
/home/daniil/projects/nimony/src/nimony/nimsem.nim(70) processModules
/home/daniil/projects/nimony/src/nimony/semmain.nim(675) semcheck
/home/daniil/projects/nimony/src/nimony/semmain.nim(480) semcheckCore
/home/daniil/projects/nimony/src/nimony/contracts_njvl.nim(1482) analyzeContractsNjvl
/home/daniil/projects/nimony/src/nimony/contracts_njvl.nim(1442) traverseToplevel
/home/daniil/projects/nimony/src/nimony/contracts_njvl.nim(1450) traverseToplevel
[...]

PR #1932, opened 2026-06-07
fixing #1922

blackmius:
```nim
proc a[T](x: T) = discard
a(proc(): int = 0)
```

Root cause: In semdecls.nim:1018-1023, the anonymous proc handler in semProc set it.typ (the expression's type cursor) to point into the full proc declaration — including the name symbol (symdefToken(anonproc.xxx)) and body. When this cursor was used as a generic parameter binding during a[T](x: T) / a(proc(): int = 0), the full declaration (with symdefToken instead of DotToken for the name slot) was substituted into the param type position. The RoutineTypes type handler then encountered a non-dot name token and couldn't parse the type, crashing at takeParRi.

Fix (src/nimony/semdecls.nim:1014-1034): Instead of pointing it.typ into the proc declaration, the fix builds a proper (proctype ...) type by extracting just the params and return type from the processed declaration, wrapping them with a nilability tag and a pragmas dot slot — the correct canonical form for a routine type.

```nim
var a: seq[proc(x: int)] = @[]
```

Root cause: The NJVL converter (src/njvl/vl.nim:131) passes SymbolDef tokens through into expressions when the type proc(x: int) is embedded within expression trees (e.g., in @[] with type seq[proc(x: int)]). But traverseExpr in contracts_njvl.nim:798 treated SymbolDef as a bug, crashing with [Bug] symbol definition in expression.

Fix: src/nimony/contracts_njvl.nim:797 — changed the SymbolDef case from bug to inc pc to skip over these tokens, matching how the NJVL converter already handles them. SymbolDef inside expression-embedded types is metadata (e.g., param names in proc types) that doesn't need contract analysis.

```nim
type A = object
  v: seq[proc(x: int)]
proc a(x: A) =
  for f in x.v:
    f(10)
```

Root cause: The items iterator for openArray[T] returns var T. When iterating for f in x.v where x.v: seq[proc(x: int)], the loop variable f gets type (mut (proctype ...)) instead of bare (proctype ...). The typeKind in RoutineTypes checks then fail because MutT isn't in RoutineTypes, causing resolveOverloads to take the wrong code path and produce "attempt to call routine".

Fixes (4 files):
1. src/nimony/semcall.nim:788 — In the OchoiceX branch of resolveOverloads, strip type modifiers before checking RoutineTypes.
2. src/nimony/semcall.nim:820 — In the Symbol branch, strip type modifiers from cs.fn.typ before both the RoutineTypes check and the FnCandidate construction.
3. src/nimony/sigmatch.nim:1778-1781 — Strip type modifiers from fn.typ before skipToParams, otherwise the assert f.substructureKind == ParamsU fails when the type is (mut (proctype ...)).
4. src/nimony/decls.nim:74-79 — Fix skipProcTypeToParams to strip type modifiers before checking RoutineTypes, preventing assertion failures in derefs.nim:589 and other callers.

```nim
type A[T] = object
  v: seq[proc(x: T)]
proc a[T](): A[T] = A[T](v: @[])
discard a[int]()
```

The root cause was in typeToCanon in src/nimony/semdata.nim:241. This function creates canonical string keys for type deduplication and instantiation caching.
When two seq[proc(x: int)] type trees were processed, the param x inside the (proctype ...) got a different SymbolDef symId each time semParams ran (fresh allocation per declaration). typeToCanon was including the raw symId in the key, causing two structurally identical types to produce different canonical keys → cache miss → two separate seq[proc(x: int)] instantiations → sameTrees returned false in commonType → false "type mismatch" error.
The fix: for SymbolDef tokens (param names), emit a fixed placeholder !symdef instead of the symId. Param names don't affect type identity in Nim, and this ensures seq[proc(x: int)] always canonicalizes to the same key regardless of the param's internal symId allocation.

### DIFF
--- a/src/nimony/decls.nim
+++ b/src/nimony/decls.nim
@@ -74,8 +74,6 @@ proc skipToParams*(c: var Cursor) =
 proc skipProcTypeToParams*(t: Cursor): Cursor =
   ## Pure version: returns a cursor advanced past the prefix slots.
   result = t
-  if result.typeKind in TypeModifiers:
-    result = result.skipModifier
   if result.typeKind in RoutineTypes:
     skipToParams result
 
--- a/src/nimony/decls.nim
+++ b/src/nimony/decls.nim
@@ -74,6 +74,8 @@ proc skipToParams*(c: var Cursor) =
 proc skipProcTypeToParams*(t: Cursor): Cursor =
   ## Pure version: returns a cursor advanced past the prefix slots.
   result = t
+  if result.typeKind in TypeModifiers:
+    result = result.skipModifier
   if result.typeKind in RoutineTypes:
     skipToParams result
 
--- a/src/nimony/semcall.nim
+++ b/src/nimony/semcall.nim
@@ -785,7 +785,7 @@ proc resolveOverloads(c: var SemContext; dest: var TokenBuf; it: var Item; cs: v
         let sym = f.symId
         let s = fetchSym(c, sym)
         let typ = fetchCallableType(c, dest, f, s)
-        if typ.typeKind in RoutineTypes:
+        if typ.skipModifier.typeKind in RoutineTypes:
           let candidate = FnCandidate(kind: s.kind, sym: sym, typ: typ)
           m.add createMatch(addr c)
           sigmatchNamedArgs(m[^1], candidate, cs.args, genericArgs, cs.hasNamedArgs)
@@ -817,8 +817,8 @@ proc resolveOverloads(c: var SemContext; dest: var TokenBuf; it: var Item; cs: v
     # Keep in mind that proc vars are a thing:
     let sym = if cs.fn.n.kind == Symbol: cs.fn.n.symId else: SymId(0)
     let typ = cs.fn.typ
-    if typ.typeKind in RoutineTypes:
-      let candidate = FnCandidate(kind: cs.fnKind, sym: sym, typ: typ)
+    if typ.skipModifier.typeKind in RoutineTypes:
+      let candidate = FnCandidate(kind: cs.fnKind, sym: sym, typ: typ.skipModifier)
       m.add createMatch(addr c)
       sigmatchNamedArgs(m[^1], candidate, cs.args, genericArgs, cs.hasNamedArgs)
       considerTypeboundOps(c, m, cs.fnName, cs.args, genericArgs, cs.hasNamedArgs)
--- a/src/nimony/semcall.nim
+++ b/src/nimony/semcall.nim
@@ -785,8 +785,9 @@ proc resolveOverloads(c: var SemContext; dest: var TokenBuf; it: var Item; cs: v
         let sym = f.symId
         let s = fetchSym(c, sym)
         let typ = fetchCallableType(c, dest, f, s)
-        if typ.skipModifier.typeKind in RoutineTypes:
-          let candidate = FnCandidate(kind: s.kind, sym: sym, typ: typ)
+        let maybeProc = typ.skipModifier
+        if maybeProc.typeKind in RoutineTypes:
+          let candidate = FnCandidate(kind: s.kind, sym: sym, typ: maybeProc)
           m.add createMatch(addr c)
           sigmatchNamedArgs(m[^1], candidate, cs.args, genericArgs, cs.hasNamedArgs)
       else:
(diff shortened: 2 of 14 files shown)


src/nimony/semdecls.nim
```diff
@@ -1015,12 +1015,30 @@ proc semProc(c: var SemContext; dest: var TokenBuf; it: var Item; kind: SymKind;
     semProcImpl c, anons, it, kind, pass, name
     dest.add parLeToken(ExprX, info)
     dest.add parLeToken(StmtsS, info)
-    let anonTypePos = dest.len
     dest.add anons
     dest.addParRi()
     dest.add symToken(name, info)
     dest.addParRi()
-    it.typ = typeToCursor(c, dest, anonTypePos)
+    # build a proper type for it.typ (not the full proc declaration)
```
Araq: No, instead make the proctype handler more flexible.

blackmius: not sure what exactly did you mean, I tried to change handling in semLocalTypeImpl, and there a proc feeded insted proctype

```nim
proc a[T](x: T) = discard
a(proc(): int = 0)
```

the problem mostly because it has a body but there was a `wantDot c, dest, n` check

blackmius:
at least this now works

```nim
import std/syncio
type 
  Listener[T] = proc(x: T)
  Effect[T] = ref object
    value: T
    listeners: seq[Listener[T]]
proc effect[T](x: T): Effect[T] = Effect[T](value: x, listeners: @[])
proc get[T](e: Effect[T]): T = e.value
proc set[T](e: Effect[T], v: T) =
    e.value = v
    for fn in e.listeners:
        fn(v)
proc on[T](e: Effect[T], fn: Listener[T]) = e.listeners.add fn
let y = effect(0)
y.on proc(x: int) = echo "update:", x
let x = effect(10)
x.on proc(x: int) = y.set(x*2)

x.set(1)
x.set(5)
x.set(10)
```

src/nimony/sigmatch.nim
```diff
@@ -1776,6 +1776,8 @@ proc sigmatch*(m: var Match; fn: FnCandidate; args: openArray[CallArg];
   matchTypevars m, fn, explicitTypeVars
 
   var f = fn.typ
+  if f.typeKind in TypeModifiers:
+    f = f.skipModifier
```
Araq: Nope.

src/nimony/decls.nim
```diff
@@ -74,6 +74,8 @@ proc skipToParams*(c: var Cursor) =
 proc skipProcTypeToParams*(t: Cursor): Cursor =
   ## Pure version: returns a cursor advanced past the prefix slots.
   result = t
+  if result.typeKind in TypeModifiers:
+    result = result.skipModifier
```
Araq: Why would you need that? -> caller is broken.

src/nimony/semcall.nim
```diff
@@ -785,7 +785,7 @@ proc resolveOverloads(c: var SemContext; dest: var TokenBuf; it: var Item; cs: v
         let sym = f.symId
         let s = fetchSym(c, sym)
         let typ = fetchCallableType(c, dest, f, s)
-        if typ.typeKind in RoutineTypes:
+        if typ.skipModifier.typeKind in RoutineTypes:
```
Araq: If only there was a way to do `let maybeProc = typ.skipModifier` and then reuse that expression.

src/nimony/semcall.nim
```diff
@@ -817,8 +817,8 @@ proc resolveOverloads(c: var SemContext; dest: var TokenBuf; it: var Item; cs: v
     # Keep in mind that proc vars are a thing:
     let sym = if cs.fn.n.kind == Symbol: cs.fn.n.symId else: SymId(0)
     let typ = cs.fn.typ
-    if typ.typeKind in RoutineTypes:
-      let candidate = FnCandidate(kind: cs.fnKind, sym: sym, typ: typ)
+    if typ.skipModifier.typeKind in RoutineTypes:
```
Araq: Here maybe.

src/nimony/semcall.nim
```diff
@@ -817,8 +817,8 @@ proc resolveOverloads(c: var SemContext; dest: var TokenBuf; it: var Item; cs: v
...
+      let candidate = FnCandidate(kind: cs.fnKind, sym: sym, typ: typ.skipModifier)
```
Araq: And then over here too. Too bad we cannot extract common subexpressions in Nim.

src/nimony/decls.nim
blackmius: to skip mut wrapper, I am not sure if I can globally do something with it

```nim
var a: seq[proc(x: int)] = @[]
for fn in a:
  fn(10)
```

fn in this code resolves as
```
(mut
 (proctype
  (notnil)
  (params
   (param :x.0 . .
    (i 64).)). .))
``

blackmius: I can move skipModifier to trCall

PR #1939, opened 2026-06-09
Implemented concept inheritance via `concept of`

gmpreussner:
This change covers all use cases of `concept of` that I could think of, including multiple inheritance, interaction with templates, typedef parameters in procs and templates, and dot notation for call sites. I included several new unit tests in `tests/nimony/concepts` that represent minimal repros for the basic functionality I would expect, as well as all the issues I ran into along the way. I tried to keep it minimally invasive, which is why you'll see mostly additions to the existing code; only the tree comparison code was refactored to avoid duplication.


src/nimony/sigmatch.nim
```diff
@@ -475,6 +555,163 @@ proc matchesConstraint(m: var Match; f: SymId; a: Cursor): bool =
   assert typevar.kind == TypevarY
   result = matchesConstraint(m, typevar.typ, a)
 
+proc skipTypeSourceAnnot(n: var Cursor) =
+  ## Drop module/path slots in a type spelling before structural compare.
```
Araq: What the heck is that?

src/nimony/sigmatch.nim
```diff
@@ -475,6 +555,163 @@ proc matchesConstraint(m: var Match; f: SymId; a: Cursor): bool =
...
+  while n.kind == StringLit:
+    skip n
+  while n.kind == Symbol:
+    var name = pool.syms[n.symId]
+    if ',' in name or '/' in name or '~' in name:
```
Araq: No idea what that is supposed to do...

src/nimony/sigmatch.nim
gmpreussner: This is used to strip metadata annotations from return types when comparing signatures in concepts. It seems that the candidate implementations can differ in symbol IDs and module/path annotations even when they mean the same type. Without this, they'd be rejected.

src/nimony/sigmatch.nim
gmpreussner: See above; normalization for concept matching

src/nimony/sigmatch.nim
Araq: That's not how to do it... See hexer's nifcgen.nim how to deal with `(importc ..)` type annoations etc properly.

Araq:
I took over.

PR #1942, opened 2026-06-10
Fix float and int literals not being passable to float32 parameters

Nimaoth:
Allow passing float and int literals to functions expecting float32 if the literal can be represented as a float32 without precision loss.


Araq:
**if the literal can be represented as a float32 without precision loss.** YES!

PR #1947, opened 2026-06-13
adds std/algorithm

demotomohiro:
This PR brings a part of std/algorithm in Nim 2 to Nimony

### DIFF
--- a/lib/std/algorithm.nim
+++ b/lib/std/algorithm.nim
@@ -20,16 +20,7 @@ func `*`*(x: int; order: SortOrder): int {.inline.} =
   var y = order.ord - 1
   result = (x xor y) - y
 
-template `<-`(a, b) =
-  #when defined(gcDestructors):
-  when true:
-    a = move b
-  elif onlySafeCode:
-    shallowCopy(a, b)
-  else:
-    copyMem(addr(a), addr(b), sizeof(T))
-
-proc mergeAlt[T](a: var openArray[T]; b: var seq[T];
+proc mergeAlt[T](a, b: var openArray[T];
                  lo, m, hi: int;
                  cmp: proc (x, y: T): int;
                  order: SortOrder) =
@@ -37,15 +28,19 @@ proc mergeAlt[T](a: var openArray[T]; b: var seq[T];
   # 1 2 3 4 ## 5 6 7 8
   # -> O(n) for sorted arrays.
   # On random data this saves up to 40% of mergeAlt calls.
+
+  # all values in `b` is uninitialized
+  # don't call destructor to `b[i]`
   if cmp(a[m], a[m+1]) * order <= 0: return
   var j = lo
-  # copy a[j..m] into b:
+  # move a[j..m] into b:
   assert j <= m
   #when onlySafeCode:
   when true:
-    b.shrink(0)
+    var bb = 0
     while j <= m:
-      b.add a[j]
+      swap b[bb], a[j]
+      inc(bb)
       inc(j)
   else:
     copyMem(addr(b[0]), addr(a[j]), sizeof(T)*(m-j+1))
@@ -55,21 +50,23 @@ proc mergeAlt[T](a: var openArray[T]; b: var seq[T];
   # copy proper element back:
   while k < j and j <= hi:
     if cmp(b[i], a[j]) * order <= 0:
-      a[k] <- b[i]
+      swap a[k], b[i]
       inc(i)
     else:
-      a[k] <- a[j]
+      swap a[k], a[j]
       inc(j)
     inc(k)
   # copy rest of b:
   #when onlySafeCode:
   when true:
     while k < j:
-      a[k] <- b[i]
+      swap a[k], b[i]
       inc(k)
       inc(i)
   else:
     if k < j: copyMem(addr(a[k]), addr(b[i]), sizeof(T)*(j-k))
+
+  # all uninitialized values moved from `b` to `a` returned to `b`
 
 proc sort*[T](a: var openArray[T];
               cmp: proc (x, y: T): int;
@@ -114,8 +111,9 @@ proc sort*[T](a: var openArray[T];
       else: -1
     sort(d, myCmp)
     assert d == ["fo", "qux", "boo", "barr"]
+
   var n = a.len
-  var b = newSeqOfCap[T](n div 2)
+  var b = newSeqUninit[T](n div 2)
   var s = 1
   while s < n:
     var m = n-1-s
@@ -123,6 +121,8 @@ proc sort*[T](a: var openArray[T];
       mergeAlt(a, b, max(m-s+1, 0), m, m+s, cmp, order)
       dec(m, s*2)
     s = s*2
+
+  destroyUninit(b)
 
 proc sorted*[T](a: openArray[T]; cmp: proc(x, y: T): int;
                 order = SortOrder.Ascending): seq[T] =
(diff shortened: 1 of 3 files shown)


lib/std/algorithm.nim
```diff
@@ -0,0 +1,174 @@
...
+  #when onlySafeCode:
+  when true:
+    while k < j:
+      a[k] <- b[i]
+      inc(k)
+      inc(i)
+  else:
+    if k < j: copyMem(addr(a[k]), addr(b[i]), sizeof(T)*(j-k))
+
+proc sort*[T: HasDefault](a: var openArray[T];
```
Araq: Makes no sense, sort's job is to sort and it doesn't construct values hence the `T: HasDefault` restriction is wrong.

lib/std/algorithm.nim
```diff
@@ -0,0 +1,174 @@
...
+  ## * `sortedByIt template<#sortedByIt.t,untyped,untyped>`_
+  runnableExamples:
+    var d = ["boo", "fo", "barr", "qux"]
+    proc myCmp(x, y: string): int =
+      if x.len() > y.len() or x.len() == y.len(): 1
+      else: -1
+    sort(d, myCmp)
+    assert d == ["fo", "qux", "boo", "barr"]
+  var n = a.len
+  var b = newSeq[T](n div 2)
```
Araq: Needs a `newSeqUninit` or something like that.

Araq:
And likewise for the other procs that repeat this pattern.

lib/std/algorithm.nim
demotomohiro: Removed `HasDefault` concept.

lib/std/algorithm.nim
demotomohiro: Replaced it with `newSeqUninit`.
And added `nodestroy` pragma to procs that write to uninitialized seq so that destructor is not called with uninitialized values when T has destructor.

lib/std/algorithm.nim
```diff
@@ -0,0 +1,177 @@
...
+  var s = 1
+  while s < n:
+    var m = n-1-s
+    while m >= 0:
+      mergeAlt(a, b, max(m-s+1, 0), m, m+s, cmp, order)
+      dec(m, s*2)
+    s = s*2
+
+proc sorted*[T](a: openArray[T]; cmp: proc(x, y: T): int;
+                order = SortOrder.Ascending): seq[T] {.nodestroy.} =
```
Araq: `.nodestroy` is **Not** correct here!

lib/std/algorithm.nim
```diff
@@ -0,0 +1,177 @@
...
+      a = [2, 3, 1, 5, 4]
+      b = sorted(a, system.cmp[int])
+      c = sorted(a, system.cmp[int], Descending)
+      d = sorted(["adam", "dande", "brian", "cat"], system.cmp[string])
+    assert b == @[1, 2, 3, 4, 5]
+    assert c == @[5, 4, 3, 2, 1]
+    assert d == @["adam", "brian", "cat", "dande"]
+  result = newSeqUninit[T](a.len)
+  for i in 0 .. a.high:
+    result[i] = a[i]
```
Araq: Because this would lose RC values, for example. This is a raw per-byte assignment via `.nodestroy`!

Araq:
```suggestion
    result[i] = `=dup`(a[i])
```

tests/nimony/stdlib/talgorithm.nim
```diff
@@ -0,0 +1,26 @@
...
+block:
+  var d = [1, 11, 7, 3, 5]
+  let d2 = sorted(d, myCmpInt)
+  assert d2 == [1, 3, 5, 7, 11]
+
+  assert sorted(["boo", "fo", "barr", "qux"], myCmpStr) == ["fo", "qux", "boo", "barr"]
+
+block:
+  assert isSorted([1, 2, 3], myCmpInt)
+  assert not isSorted([3, 2, 1], myCmpInt)
```
Araq, line 26: Test it with strings that are put onto the heap and run it under valgrind cause your `.nodestroy` makes me nervous.

demotomohiro:
valgrind detected error `Conditional jump or move depends on uninitialised value(s)` in `mergeAlt` proc.
So I rewrote it to use `newSeqOfCap` and removed `nodestroy` pragma.

valgrind also detected error `Conditional jump or move depends on uninitialised value(s)` in `sorted` proc.
I read generated C code and it seems following code still calls destructor with uninitialised value when assign to `result[i]`:
```nim
  result = newSeqUninit[T](a.len)
  for i in 0 .. a.high:
    result[i] = `=dup`(a[i])
```
In generated C code, `putQ_7_Ild9iim_talfb69581` calls `nimStrDestroy((*X60Qlhs_1))` and `X60Qlhs_1` points to an element of `result`.
I rewrote it to:
```nim
  result = newSeqOfCap[T](a.len)
  for i in 0 .. a.high:
    result.add a[i]
```
Then, valgrind doesn't detect the error.

Araq:
Not good enough, see that

```nim

  result = newSeqUninit[T](a.len)
  for i in 0 .. a.high:
    result[i] = `=dup`(a[i])
```

works within a `.nodestroy`, it must not produce a destructor call for it.

demotomohiro:
Following code still calls destructor with uninitialized values.
valgrind detects 'Conditional jump or move depends on uninitialised value(s)'.
`nimStrDestroy` is called in generated C code.
```nim
import std/syncio

proc test(a: openArray[string]): seq[string] {.nodestroy.} =
  result = newSeqUninit[string](a.len)
  for i in 0 .. a.high:
    result[i] = `=dup`(a[i])

var s = test(["a"])
```
`nodestroy` pragma doesn't remove destructor calls in procedures called from the procedure with `nodestroy` pragma?

Araq:
> nodestroy pragma doesn't remove destructor calls in procedures called from the procedure with nodestroy pragma?

Yeah, well it should!

lib/std/algorithm.nim
```diff
@@ -0,0 +1,172 @@
...
+  ## * `sortedByIt template<#sortedByIt.t,untyped,untyped>`_
+  runnableExamples:
+    var d = ["boo", "fo", "barr", "qux"]
+    proc myCmp(x, y: string): int =
+      if x.len() > y.len() or x.len() == y.len(): 1
+      else: -1
+    sort(d, myCmp)
+    assert d == ["fo", "qux", "boo", "barr"]
+  var n = a.len
+  var b = newSeqOfCap[T](n div 2)
```
Araq: No. I told you to fix the Nimony compiler first so that `newSeqUninit` + `nodestroy` can be used!

demotomohiro:
Added `destroyUninit` func to `lib/std/system/seqimpl.nim` because existing `=destroy`*[T](s: seq[T]) calls destructor to uninitialized elements.

lib/std/system/seqimpl.nim
```diff
@@ -18,6 +18,13 @@ func `=destroy`*[T](s: seq[T]) =
       inc i
     dealloc s.data
 
+func destroyUninit*[T](s: var seq[T]) {.inline.} =
```
Araq: Nope.

lib/std/algorithm.nim
```diff
@@ -0,0 +1,172 @@
...
+  var b = newSeqUninit[T](n div 2)
+  var s = 1
+  while s < n:
+    var m = n-1-s
+    while m >= 0:
+      mergeAlt(a, b, max(m-s+1, 0), m, m+s, cmp, order)
+      dec(m, s*2)
+    s = s*2
+
+  destroyUninit(b)
```
Araq: Nope.

PR #1951, opened 2026-06-13
Fix imported custom pragma templates

planetis-m:
## Summary

  Fix Nimony custom pragma templates so exported `{.pragma.}` templates work across module imports, matching the behavior that existing Nim-backed plugin code relied on.

  This makes `import plugins` expose annotations such as `ensuresNif` from `nif_annotations`, instead of requiring users to textually include the annotation module.

  ## Root Cause

  Nimony recorded custom pragma templates only in the current module's `SemContext.customPragmaTemplates`. Imported module interfaces exposed symbols, converters, and
  exports, but did not carry the information that an exported template was declared with `{.pragma.}`.

  As a result, imported annotations like `{.ensuresNif: addedAny(dest).}` were rejected with `expected pragma`.

  ## Changes

  - Add a `custompragmas` section to `.idx.nif` files.
  - Record exported templates declared with `{.pragma.}` into that index section.
  - Load indexed custom pragma names through normal import filtering.
  - Prevent imported custom pragma templates from being treated as pragma macro invocations.
  - Update `nif_annotations` to remove the obsolete “include not import” comment and add `addedNested`.
  - Restore `plugins.errorTree` helpers expected by plugin code.
  - Add regression tests for imported custom pragmas and `import plugins` annotations.

  ## Validation

  - `git diff --cached --check`
  - `bin/nimony c -r tests/nimony/pragmas/timported_custompragma.nim`
  - `bin/nimony c -r tests/nimony/plugins/tplugin_annotations.nim`
  - `nim c -r src/hastur build all`

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -444,12 +444,15 @@ proc renderTree*(tree: NifBuilder): string        # debug rendering (no line inf
 
 ```nim
 proc errorTree*(msg: string): NifBuilder
-proc errorTree*(msg: string; at: Node): NifBuilder
-proc errorTree*(msg: string; at, orig: Node): NifBuilder
+proc errorTree*(msg: string; info: LineInfo): NifBuilder
+proc errorTree*(msg: string; at: NifCursor): NifBuilder
+proc errorTree*(msg: string; at, orig: NifCursor): NifBuilder
 ```
 
-Return an `ErrT` node that the compiler reports as a compile-time error. The `at`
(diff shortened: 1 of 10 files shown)


Araq:
Are you sure custom pragmas are exported in Nim? That's not how I remember things to work.

planetis-m:
Alright seems to be taken care of. Not sure when errorTree was removed and what's the reasoning and alternative.

PR #1958, opened 2026-06-14
Support resolving fields in idetools

Nimaoth:
Add support finding definitions for dot expressions in `idetools.nim`

### DIFF
--- a/src/nimony/idetools.nim
+++ b/src/nimony/idetools.nim
@@ -75,7 +75,7 @@ proc findAndAddFieldDefinition(c: var IdeContext, n: var Cursor, name: string, n
 proc tr(c: var IdeContext, n: var Cursor) =
   ## Traverse the tree and collect definitions and usages matching the sym in the context
   case n.stmtKind
-  of ScopeS, StmtsS:
+  of ScopeS:
     c.typeCache.openScope()
     n.loopInto():
       tr(c, n)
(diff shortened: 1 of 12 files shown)


src/nimony/idetools.nim
```diff
@@ -40,57 +41,228 @@ proc foundSymbol(tok: PackedToken; mode: TrackMode) =
...
+      inc sym
+      if sym.kind == SymbolDef and sym.symId == nameSym:
+        c.usages.add(Usage(n: sym, containingType: containingType))
+    skip n
 
-        if nested == 0: break
+proc tr(c: var IdeContext, n: var Cursor) =
+  ## Traverse the tree and collect definitions and usages matching the sym in the context
+  case n.stmtKind
+  of ScopeS, StmtsS:
```
Araq: ONLY ScopeS opens a new scope!!!

src/nimony/idetools.nim
```diff
@@ -40,57 +41,228 @@ proc foundSymbol(tok: PackedToken; mode: TrackMode) =
...
-    inc n
+    unsafeDec result
+
+proc findLocal(file: string; sym: SymId; toTrack: PackedLineInfo; mode: TrackMode, offset: int) =
+  var buf = parseFromFile(file)
+
+  var n = beginRead(buf)
+
+  var symCursor = buf.cursorAt(offset)
+  let container = symCursor.getParent()
```
Araq: `getParent` is ugly and you can do without.

Nimaoth: parent is now tracked when finding the symbol in the first pass by keeping a stack of parent token indices

PR #1966, opened 2026-06-17
hexer/coro: lower non-stmts ite branches as a single statement

tokyovigilante:
A .passive proc that early-returns from inside a while loop (after a suspension point) failed in hexer with '[Error] unknown statement: (not ...)'. The single-exit rewrite of the same loop worked, so only this control-flow shape was broken.

Root cause: NJ lowers `while c: body` to `while true: if c: body else: break`, and the synthesised `else: break` becomes a guard `ite` placed DIRECTLY in the else slot of the `if c` — i.e. the branch is a single statement, not a `(stmts ...)` list. trGoto's IteV handler (for an ite that contains a suspension point) lowered each branch by descending into it with `into` and iterating its children. For the bare `ite` branch that emitted the ite's *condition* — `(not g)` — as a standalone statement, which nifc rejects.

Fix: in the suspension-point IteV path, only iterate children when a branch is a `(stmts ...)`; otherwise lower the branch as one statement via trGoto. Applies to both then and else slots.

Full nimony suite green (cps 24/24, iter, closures, exceptions, errmsgs)


Araq:
> Root cause: NJ lowers while c: body to while true: if c: body else: break, and the synthesised else: break becomes a guard ite placed DIRECTLY in the else slot of the if c — i.e. the branch is a single statement, not a (stmts ...) list. 

NJ should see that the else-section has `stmts`.

PR #1967, opened 2026-06-17
sem: don't split generic instances on a type arg's nil annotation

tokyovigilante:
Under {.feature: "lenientnils".}, `var names: seq[cstring]` failed to compile: gcc rejected the generated C with 'incompatible types when assigning to type seq_0_X from seq_0_Y' — two structurally-identical seq[cstring] structs.

Root cause: the generic-instance cache key (typeToCanon) serialized the full type tree including `notnil`/`nil`/`unchecked` annotations, so `seq[(cstring)]` (bare) and `seq[(cstring (unchecked))]` — the same type; the bare form arises for an empty `@[]` element while the explicit annotation carries the lenientnils default — keyed to two cache entries, two instance symbols, two C structs. The backend mangle already strips nil annotations, so it then couldn't tell the two apart to assign between them.

Fix: add a `stripNil` option to typeToCanon and pass it at the two code-generating instance caches (instantiatedTypes, instantiatedProcs) so nil-ness no longer splits an instantiation. The general type cache (typeMem/typeToCursor) keeps the exact key, so the nil-checker still distinguishes notnil from nilable types (notnil tests stay green).

Full nimony suite green.


src/nimony/sem.nim
```diff
@@ -584,7 +584,7 @@ proc requestRoutineInstance*(c: var SemContext; origin: SymId;
                             typeArgs: TokenBuf;
                             inferred: Table[SymId, Cursor];
                             info: PackedLineInfo): ProcInstance =
-  let key = typeToCanon(typeArgs, 0)
+  let key = typeToCanon(typeArgs, 0, stripNil = true)
```
Araq, line 587: I doubt it's correct, `sem.nim` is the frontend.

src/nimony/semtypes.nim
```diff
@@ -485,7 +485,7 @@ proc semInvoke(c: var SemContext; dest: var TokenBuf; n: var Cursor) =
       not isNominal(decl.body.typeKind)):
     # we have to be eager in generic type instantiations so that type-checking
     # can do its job properly:
-    let key = typeToCanon(dest, typeStart)
+    let key = typeToCanon(dest, typeStart, stripNil = true)
```
Araq, line 488: Likewise, you are removing nil-information for the frontend here.

Araq:
> seq[(cstring)] (bare) and seq[(cstring (unchecked))]

There shall be no bare `cstring`, that is the root cause that needs to be fixed.

Araq:
Fixed it my way.

PR #1979, opened 2026-06-19
Exporting re-imported export now also looks for matching module

### LINKED ISSUE #1972: Exporting re-imported export looks for symbol instead of module name
Consider the following example:

**foo.nim**
```
type Foo* = object
  value*: float

func newFoo*(x: float): Foo =
  Foo(value: x)
```

**reexporter.nim**
```
import foo

export foo
```

**wrapper.nim**
```
import reexporter
import foo

export foo
```

**main.nim**
```
import wrapper
import std/syncio

let a = newFoo(1.0)
echo a.value
```

# Build
```
nimony c main.nim
```
```
wrapper.nim(4, 8) Error: undeclared identifier: foo
```

gmpreussner:
Fixes https://github.com/nim-lang/nimony/issues/1972

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -1025,6 +1025,14 @@ proc tryBuiltinDot(c: var SemContext; dest: var TokenBuf; it: var Item; lhs: Ite
     c.buildErr dest, info, "identifier after `.` expected"
     result = InvalidDot
   else:
+    # Module-qualified `module.sym` before type-based dispatch; see dotLhsModuleSym.
+    let moduleSym = dotLhsModuleSym(lhs)
+    if moduleSym != SymId(0):
+      result = MatchedDotSym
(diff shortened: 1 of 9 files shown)


src/nimony/sem.nim
```diff
@@ -1025,6 +1025,17 @@ proc tryBuiltinDot(c: var SemContext; dest: var TokenBuf; it: var Item; lhs: Ite
     c.buildErr dest, info, "identifier after `.` expected"
     result = InvalidDot
   else:
+    # Module-qualified `module.sym` must be resolved before type-based dot
+    # dispatch. Otherwise a local wrapper with the same name as an imported
```
Araq: I disagree with this order. Local names should be prioritized over module names which are in the global scope. We need a different solution here.

Issue #1964, opened 2026-06-16
New-style concept implementation very slow


gmpreussner:
We are currently using brute force for semchecking concept constraints. This gets extremely slow for larger and nested (`concept of`) concepts because the number of repeated lookups grows rapidly.

I locally tested a hacked together caching mechanism that provides relief, but Araq will look into a better/cleaner implementation using DFA.

gmpreussner:
We integrated the caching mechanism in https://github.com/nim-lang/nimony/pull/1989

Further improvements may be possible using a DFA, but that will be a separate task

PR #1989, opened 2026-06-22
Added minimally invasive LRU cache for concept matching

gmpreussner:
The cache is currently being used for:
- Body cache: whole concept body check for concrete types
- Routine Impl cache: Per-requirement overload resolution
- Candidates cache: Name lookup across modules/scopes

The bulk of the implementation is in _conceptcache.nim_. The rest of the compiler only calls a small hook surface. Some minor refactoring was necessary in _sigmatch.nim_

Profile counters are available with `-d:nimonyProfileConcepts` for measuring cache hit rates on real workloads.

### DIFF
--- a/src/nimony/conceptcache.nim
+++ b/src/nimony/conceptcache.nim
@@ -56,34 +56,6 @@ type
     candidatesCacheOrder*: seq[CandidatesCacheKey]
     metadata*: Table[SymId, ConceptMetadata]
 
-when defined(nimonyProfileConcepts):
-  var
-    conceptBodyChecks* = 0
-    conceptBodyCacheHits* = 0
-    conceptRoutineAvailableCalls* = 0
-    conceptRoutineImplCacheHits* = 0
-    conceptCandidateScans* = 0
-    conceptCandidateCacheHits* = 0
-    matchConceptRoutineSigCalls* = 0
-
-  proc printConceptProfile*() =
-    echo "concept profile:"
-    echo "  body_checks=", conceptBodyChecks, " body_cache_hits=", conceptBodyCacheHits
-    echo "  routine_available=", conceptRoutineAvailableCalls,
-         " routine_impl_hits=", conceptRoutineImplCacheHits
-    echo "  candidate_scans=", conceptCandidateScans,
-         " candidate_cache_hits=", conceptCandidateCacheHits
-    echo "  sig_match_calls=", matchConceptRoutineSigCalls
-else:
-  template conceptBodyChecks*() = discard
-  template conceptBodyCacheHits*() = discard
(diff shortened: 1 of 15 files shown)


gmpreussner:
The concept semchecking is already considerably faster since recent refactors. The additional speedup of this caching solution isn't as good as I was hoping - only about 30-50% in my test cases.  I marked this as a draft for a general review of the overall approach.

https://github.com/nim-lang/nimony/issues/1964

savannt:
Following up on the "better/cleaner implementation using DFA" note in #1964 — I prototyped that direction as a complement to this cache, and wanted to share the results since you asked for an overall-approach review.

**Idea:** precompile each concept requirement's parameter-type trees into a reusable structural pattern (the concept's `Self` typevar becomes a capture bound to the concrete type), then use a cheap *reject / maybe* verdict to gate the full `matchConceptRoutineSig`. The contract is **no false negatives**: it rejects only on a definite, conversion-free structural mismatch; anything involving a typevar on either side stays "maybe"; parameters drive rejection, the return type doesn't (the real matcher's return check is intentionally lenient).

This targets a different layer than the LRU cache. The cache removes *repeated identical* checks; the pre-filter makes each *un-cached* scan cheap — and on what I measured, that un-cached matcher scan is the dominant residual cost even with the candidate list fully cached.

**Numbers** (120-type non-inheritance stress benchmark, validated with a `-d:conceptDfaCheck` shadow build that runs both paths and counts):
- **0 false negatives across 29,100 candidate checks**
- the gate skips **79% of `matchConceptRoutineSig` calls** (29,100 → 6,120); the candidate-list cache was already ~99% hit, so this is purely the residue the cache can't touch

Honest caveat: end-to-end that's only ~5% on this benchmark, because concept matching is a small slice of total compile time when the concepts are flat. The design is really aimed at the *nested* `concept of` blowup that #1964 describes, where matcher calls grow super-linearly — but I couldn't measure that case locally (my host Nim predates `concept of` parsing), so I'd value a sanity check on a real nested workload.

The two compose cleanly (cache in front, pre-filter behind). The prototype is one commit on top of this PR if it's useful to look at: savannt/nimony@8500d913c10c282d8ee46d25854e8839fecafacc — happy to turn it into a proper draft PR if you'd like.

gmpreussner:
@Araq can you take a look at https://github.com/savannt/nimony/commit/8500d913c10c282d8ee46d25854e8839fecafacc? I'm not qualified to make a call.

gmpreussner:
> Honest caveat: end-to-end that's only ~5% on this benchmark, because concept matching is a small slice of total compile time when the concepts are flat.

This is expected

Araq:
Make the cache a property of `SemContext` instead of these global/threadlocal variables.

Issue #1946, opened 2026-06-13
Final IR design


Araq:
# Final IR

## Control flow constructs

There are only these control flow constructs:

- `loop`: Format `(loop <body>)`. An infinite back-edge loop with no condition slot (see below).
- `ite`: Format `(ite <cond> <then> <else>)`
- `case`: Format as existing (quite complex with range handling etc.)
- `lab` + `jmp`: Format `(lab L)` and `(jmp L)`. The sole structural multi-exit — the one mechanism for
  leaving a region early — admitted **only in their *structured* form** (defined precisely in *What
  "structured" means* below). `jmp` is the leaving transfer; `lab` is the join it lands on. Structured
  `lab`/`jmp` is exactly as expressive, and exactly as analyzable, as lexically-scoped multi-level `break`
  would be — it is that mechanism spelled with names instead of brackets.
- `try` + `finally` + `except`: the cleanup/handler constructs that survive lowering (see below).

Plus two primitive exits, `return` and `raise`. These stay primitive because they carry data and interact
with cleanup/dispatch: `return` carries the result and runs the proc epilogue; `raise` carries the
exception, is type-dispatched by `except`, and crosses `finally`. They *are* jumps — to the proc-epilogue
label and to the nearest-handler label respectively — but they are kept as named exits because they carry a
payload and cross cleanup, so they are not plain structural transfers. The clean line: **pure structural
transfers become `jmp`/`lab`; only data-or-cleanup-carrying exits stay primitive.**

### What "structured" means

`lab`/`jmp` are *not* general labels-and-gotos. They are admitted into final-IR only under two invariants
that, together, keep the control-flow graph a reducible, properly-nested DAG (plus loop back-edges) — which
is what makes every pass a total tree traversal (the exit-summary composition in *Analysis: exit summaries*)
with no CFG construction and no fixpoint. A verifier rejects any `lab`/`jmp` that violates either invariant.

1. **Forward-only.** Every `(jmp L)` targets a `(lab L)` that appears *later* in the statement stream;
   there are no backward jumps. The sole exception is the loop back-edge, which belongs to `loop` alone (see
   below). The payoff is the multi-join property: when a forward traversal reaches `(lab L)`, **every**
   predecessor of `L` has already been visited, so the join's arity and its joined state are fully known in
   that one pass. (`arity(L) = (1 if fall-through into L is reachable else 0) + count of (jmp L)`.)

[...]

PR #1998, opened 2026-06-23
njvl: lower large ordinal cases to a balanced binary search

tokyovigilante:
A `case` over a large enum (or its generated `$`) was lowered to a linear chain of `of`-comparisons in nj.nim's trCase, producing an N-deep nested `ite`. The recursive NJVL passes — vl.nim trIte, contracts_njvl traverseIte — and nifc genstmts then recurse once per branch and overflow the compiler's call-depth limit.

The threshold is ~650 branches, seen in the large Vulkan enum VkStructureType (1145 members).

Lower a case to a balanced binary-search tree once it has more than 16 discrete single-value ordinal branches: parse the `of` labels into (ordinal, body), sort by ordinal, and emit a `(lt …)/(eq …)` dispatch tree that nests only O(log N) deep, so every downstream pass handles it unmodified. Enum-field labels are still symbolic at the NJVL stage, so the ordinal is read from the field's `(efld … (tup counter name))` declaration (as expreval does). Ranges, multi-value branches, string selectors and small cases fall back to the existing linear lowering unchanged.

The catch-all (explicit `else`, or the final branch of an exhaustive case) is emitted at the tree's no-match leaves; if it declares a local we keep the linear path so the same symbol is never defined twice.

### DIFF
--- a/src/njvl/nj.nim
+++ b/src/njvl/nj.nim
@@ -9,12 +9,12 @@
 
 ## Compiles Nimony IR to a simpler IR called [NJVL](doc/njvl.md) that does not contain jumps.
 
-import std / [tables, hashes, sets, assertions, syncio]
+import std / [tables, hashes, sets, assertions, syncio, algorithm]
 when defined(nimony):
   {.feature: "lenientnils".}
 include ".." / lib / nifprelude
 include ".." / lib / compat2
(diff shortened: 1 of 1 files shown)


Araq:
NJ is dead btw, to be replaced by https://github.com/nim-lang/nimony/issues/1946 which is not "final" either, however...

tokyovigilante:
Ok will keep this local for now and and eye on #1946

PR #1999, opened 2026-06-23
std/json: tolerate out-of-range integer literals

tokyovigilante:
parseValue aborted the whole parse via parseBiggestInt's `quit` when an integer literal exceeded int64 (e.g. C's UINT64_MAX in libclang/opir header dumps). Add fitsInt64 to range-check the literal first and fall back to a float atom for out-of-range values — the standard JSON approach — instead of aborting.

Lossy for magnitudes > 2^53, but the parse never aborts. Boundary cases (int64.high/low fit; ±1 and uint64.max become floats) covered in tests/nimony/stdlib/tjson.nim.


Araq:
Hack. Instead make `parseBiggestInt`'s return value more useful or something like that.

PR #2004, opened 2026-06-24
Pass template name to plugin input for shared-plugin dispatch

planetis-m:
Template-plugin input now begins with the invoked template's name as a bare identifier, so multiple templates can share one plugin binary and dispatch on which was called. Adds templateName()/templateArgs() helpers to the plugins module. Updates existing template plugins to use the new helpers and adds a shared-plugin test.


src/nimony/sem.nim
```diff
@@ -2780,6 +2780,65 @@ proc isIdentCall(c: var SemContext; dest: var TokenBuf; beforeCall: int): bool {
...
+  ## When a `for` loop's iterator resolves to a routine with a `.plugin`
+  ## pragma, invoke the plugin with the for-loop context instead of doing
+  ## normal iterator resolution. The plugin receives:
+  ##   (stmts <iter-name> <call-args...> <loop-vars> <loop-body>)
+  ## and its output replaces the entire for loop, then gets re-semanticked.
+  result = false
+  if dest.len <= beforeCall + 1: return
+  let callKind =
+    if dest[beforeCall].kind == ParLe and rawTagIsNimonyExpr(tagEnum(dest[beforeCall])):
+      cast[NimonyExpr](tagEnum(dest[beforeCall]))
```
Araq: Ugly, use the NIF API properly.

src/nimony/sem.nim
```diff
@@ -2780,6 +2780,65 @@ proc isIdentCall(c: var SemContext; dest: var TokenBuf; beforeCall: int): bool {
...
+    else:
+      NoExpr
+  if callKind notin CallKinds: return
+  if dest[beforeCall + 1].kind != Symbol: return
+  let sym = fetchSym(c, dest[beforeCall + 1].symId)
+  let res = declToCursor(c, dest, sym)
+  if res.status != LacksNothing: return
+  if not isRoutine(res.decl.symKind): return
```
Araq: return, return, return, return... Bad style.

src/nimony/sem.nim
planetis-m: It was taken care of.

src/nimony/sem.nim
planetis-m: The guards were joined to make less branches, but I dont know if you prefer structured control flow.

PR #2022, opened 2026-06-26
Consistent plugin API: rename, type helper, re-sem docs

planetis-m:
Makes the plugin API names and documentation consistent across all plugin kinds.

- **Rename `templateName`/`templateArgs` → `pluginName`/`pluginCallArgs`** — these helpers work for both template and for-loop plugins, so the names should reflect that. Old names are kept as deprecated `template` aliases so existing plugins don't break.
- **Add `loadTypeDefinitions()`** — wraps `loadPluginInput(paramStr(3))` so type plugin authors no longer need to know the `paramStr` convention.
- **Document the re-semantacking contract** on `saveTree` — template and for-loop plugin output is re-semantacked (can use raw identifiers like `addIdent "echo"`), while module and type plugin output is not (must be fully semanticked NIF). This is now explicit in the docstring rather than implicit.

All 12 tracked plugin tests + pluginpaths test pass.

### DIFF
--- a/src/nimony/lib/plugins.nim
+++ b/src/nimony/lib/plugins.nim
@@ -806,6 +806,9 @@ proc saveReplacer*(t: Replacer; filename = paramStr(2)) =
 
 proc loadPluginInput*(filename = paramStr(1)): NifCursor =
   ## Loads a NIF file and returns a root `NifCursor` for reading it.
+  ##
+  ## For type plugins, use `loadTypeDefinitions()` to read the second input
+  ## file (`paramStr(3)`) that carries the triggering type definitions.
   var inp = nifstreams.open(filename)
   try:
     var tree = createTree(fromStream(inp))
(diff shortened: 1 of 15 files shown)


Araq:
we should also take the opportunity and change `(stmts fn args)` to `(call fn args)` IMHO, it's clearer even though the `stmts` root node is also quite standardized by now...

planetis-m:
> we should also take the opportunity and change `(stmts fn args)` to `(call fn args)` IMHO, it's clearer even though the `stmts` root node is also quite standardized by now...

Yes! I agree. Also module plugins do not expose the module name. But also I would consider if extra information like the name could be passed as an extra parameter instead, and keep the original `(stmt args...` structure.

planetis-m:
@Araq I tried the  (stmts ...)  ->  (call ...)  suggestion. While it makes sense for template plugins in isolation, for-loop plugin input is  (stmts iter-name call-args loop-vars body), loop vars and body aren't call arguments, so  (call ...)  is semantically wrong there. I kept  (stmts ...)  for all plugin kinds since it's honest and keeps the helper API checking a single tag.

I also prototyped a module name prefix for module plugins (same dispatch pattern) but dropped it as it copies the whole module buffer on every invocation for a use case with no real users.

PR #2026, opened 2026-06-27
Feat/forcall structured input

planetis-m:
## Summary

  Restructures for-loop plugin input from a flat positional list to
   explicitly grouped components, eliminating fragile forward scans.

  ### Before
            `(stmts <name> <args...> <vars> <body>)`
  — components are positionally delimited but structurally
   indistinguishable. Every plugin re-implements forward scans to find
   where call args end and loop vars begin. The parfor plugin even had a
   bug where atom arguments broke the `ParLe`-only scan.

  ### After
           `(forcall <name> (callargs <args>...) (unpackflat ...) <body>)`
  — each component is a bounded subtree directly addressable via
   `firstChild`+`skip`.

  ### Changes
             - **New NIF tags**: `ForcallU`, `CallargsU` (NimonyOther)
  - **Compiler** (`sem.nim`): emits structured `(forcall ...)` shape
  - **Plugin API** (`plugins.nim`): `pluginCallArgs` enters `(callargs)`,
    `forLoopVars` skips to `(unpackflat)`, `forLoopBody` skips one more
  - **Parfor plugin**: `collectArgs` went from 11-line scan to 5-line
    iteration; `loopVarSym` no longer descends through a wrapper
  - **Hexer pipeline**: exhaustive case statements updated

  No `(loopvars ...)` wrapper — `(unpackflat ...)` is already uniquely
   identifiable, so wrapping it added a nesting level for nothing.

  All 12 plugin tests + pluginpaths + both parfor tests pass.


planetis-m:
I could eliminate ForcallU, then the helpers would detect the shape by looking at the second child, if it's  (callargs ...) , it's for-loop; otherwise it's template args. But its a heuristic versus a clean tag.

planetis-m:
The API is now cleanly split:

| Helper | Kind | Returns |
|---|---|---|
| `pluginName(n)` | template/for-loop | trigger name |
| `callArgs(n)` | template only | first arg from `(stmts ...)` |
| `forLoopCallArgs(n)` | for-loop only | `(callargs ...)` node for `.into:` |
| `forLoopVars(n)` | for-loop only | `(unpackflat ...)` |
| `forLoopBody(n)` | for-loop only | body |

PR #2028, opened 2026-06-28
Port plugins to nifcore

planetis-m:
## Summary

  - Port the Nimony plugin API from `nifcursors`/`nifstreams` to `nifcore`.
  - Use native `TagLit`, bounded cursors, and move-only `TokenBuf` builders.
  - Share pre-seeded pools across plugin inputs and generated output.
  - Densify plugin input locations to preserve effective source information.
  - Add optional line-info omission to `nifcoreparse.toString`.
  - Update plugin documentation and tests for the native API.

  ## API changes

  - `ParLe` becomes `TagLit`; `ParRi` checks become `hasMore`.
  - `StringLit` becomes `StrLit`.
  - `addParLe`/`addParRi` become `openTree`/`closeTree`.
  - `NifBuilder` is now move-only.
  - `addTree` consumes child builders.

  ## Validation

  - Plugin tests: 12/12 passed.
  - Plugin-path tests: 1/1 passed.
  - Nifcore tests: 2/2 passed.
  - Host Nim compilation and execution of `tnifcoreparse` passed.
  - `git diff --check` passed.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -423,7 +423,7 @@ snapshot transparently detaches its token storage.
 |------|-------------|
 | `takeTree(t, n)` | Copy the current subtree from `n` into `t`, advancing `n` |
 | `addSubtree(t, n)` | Copy subtree from `n` into `t` without advancing |
-| `addTree(t, child)` | Consume and append an entire child builder |
+| `addTree(t, child)` | Append an entire child builder without consuming it |
 
 **Structured building:**
 
@@ -496,7 +496,7 @@ saveReplacer(r)
 The core operations are:
 - `keep r, Kind` — copy one child verbatim, asserting its kind
 - `drop r, Kind` — skip one child without emitting, asserting its kind
-- `replace r, Kind, node` — skip one child of `Kind`, emit a replacement (`NifCursor` or `NifBuilder`)
+- `replace r, Kind, node` — skip one child of `Kind`, emit a replacement (`NifCursor` or borrowed `NifBuilder`)
 - `keepTag r:` — descend into a compound node (copy tag, process children, close)
 - `loopKeepTag r:` — copy tag, iterate all children, close
 - `peek r:` — read-ahead analysis without consuming (cursor is restored)
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -7,11 +7,12 @@ compiler through NIF files.
 
 ## Overview
 
-There are four kinds of plugins:
+There are five kinds of plugins:
 
 | Kind | Declaration | Scope |
 |------|-------------|-------|
 | **Template plugin** | `template foo(...) {.plugin: "path".}` | Invoked at each call site |
+| **For-loop plugin** | `iterator foo(...) {.plugin: "path".}` | Rewrites a `for` loop using the iterator |
 | **Module plugin** | `{.plugin: "path".}` as statement | Transforms the entire module |
 | **Type plugin** | `type T {.plugin: "path".} = ...` | Invoked for every module that uses `T` |
 | **Import plugin** | `import (path/foo) {.plugin: "std/v2".}` | Imports the module `path/foo` **from the plugin** `std/v2` |
@@ -43,15 +44,14 @@ And in `deps/mplugin1.nim`:
 ```nim
 import plugins
 
-proc transform(n: Node): NifBuilder =
+proc transform(n: NifCursor): NifBuilder =
   result = createTree()
   let info = n.info
-  var n = n
-  if n.stmtKind == StmtsS: inc n
+  var arg = callArgs(n)
   result.withTree StmtsS, info:
     result.withTree CallS, info:
       result.addIdent "echo"
-      result.takeTree n
+      result.takeTree arg
 
 var inp = loadPluginInput()
 saveTree transform(inp)
@@ -126,16 +126,13 @@ an expression, or a substructure) and only then call the appropriate accessor.
 
 ### Building with the enum classes
 
-`createTree` and `withTree` accept any of the five enum types directly:
+`withTree` accepts any of the five enum types directly:
 
 ```nim
 result.withTree StmtsS, info:                 # NimonyStmt
   result.withTree CallX, info:                # NimonyExpr
     result.addIdent "echo"
     result.addStrLit "hello"
-
-let t = createTree(ObjectT,                    # NimonyType
-  createTree(FldU, nameSym, intType))          # NimonyOther
 ```
 
 Passing the enum rather than a raw string means a typo becomes a compile error
@@ -189,13 +186,13 @@ template generateEcho(s: string) {.plugin: "deps/mplugin1".}
 generateEcho("Hello, world!")
 ```
 
-The input `Node` is wrapped in a `StmtsS` node containing the arguments. A typical
-plugin starts by skipping past this wrapper:
+The input is wrapped in a `StmtsS` node containing the template name and
+arguments. Use `callArgs` to obtain a bounded cursor over the arguments:
 
 ```nim
-var n = loadPluginInput()
-if n.stmtKind == StmtsS: inc n
-# n now points to the first argument
+let input = loadPluginInput()
+var args = callArgs(input)
+# args now points to the first argument
 ```
 
 A template plugin can also accept a code block via `typed` parameters:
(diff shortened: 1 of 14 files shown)


planetis-m:
I asked GPT5.5 high for a review:

  Highest priority:

  - balancedTokens no longer preserves its old behavior. Previously it excluded the root and advanced the cursor past the scanned subtree. Now it includes the root and
    leaves the caller’s cursor at the last visited tag. That is a silent semantic break. src/nimony/lib/plugins.nim:301

  - errorTree(msg, at) dereferences at.info when at is exhausted. With nifcore, exhaustion is rem == 0, not a readable ParRi, so this can assert. src/nimony/lib/
    plugins.nim:196

  - tag(n) changed from returning the error tag for non-tree tokens to asserting through cursorTagId. Also, $SymId and $TagId now render numeric IDs instead of names.
    These compatibility changes were not intentional or documented. src/nimony/lib/plugins.nim:68

  API improvements:

  - export nifcore exposes too much: raw buffer constructors, pool-local IDs, and operations that can create cursors/builders incompatible with the plugin-global pools.
    Then helpers such as rawTag, tagText(TagId), symText(SymId), and filePath silently assume those global pools. Selective exports or an opaque plugin wrapper would
    encode that invariant. src/nimony/lib/plugins.nim:18

  - withTree lost its enum constraint and now accepts any expression until conversion fails or produces an invalid tag. It should retain the original enum union. src/
    nimony/lib/plugins.nim:112

  - The two createTree(kind, children) overloads are untested, unused, and no longer source-compatible because children changed from builders to cursors. They should
    either be redesigned intentionally or removed. src/nimony/lib/plugins.nim:739

  Efficiency/clarity:

  - enterPluginScope/leavePluginScope copy reference-counted cursors unnecessarily. Native nifcore into already provides the exact bounded semantics and is already used
    successfully by plugin tests. src/nimony/lib/plugins.nim:134

  - Line-free rendering still decodes line information, maintains current positions, allocates a parent stack, and pushes/pops it for every tag. includeLineInfo = false
    needs a dedicated fast serialization path. src/lib/nifcoreparse.nim:153

  - addTree and bindSymHelper walk top-level cursors even though same-pool buffers could be appended with one bulk copy. A core appendBuffer operation would remove both
    loops. src/nimony/lib/plugins.nim:153

  - pragmaKind performs a mutating getOrIncl for lookup. It should use getKeyId; otherwise unknown identifiers consume the limited 9-bit tag namespace. src/nimony/lib/
    plugins.nim:93

  The existing tests cover the main plugin flows but missed the compatibility helpers, exhausted cursors, balancedTokens, foreign-pool misuse, and the convenience
  constructors. No files were changed during this review.

src/nimony/lib/plugins.nim
```diff
@@ -924,32 +848,13 @@ proc saveTree*(tree: NifBuilder) =
...
-    kind: K; info: LineInfo; children: openArray[NifBuilder]): NifBuilder =
-  ## Produces a new tree node of `kind` and line info `info` containing `children`.
-  result = createTree()
-  prepareMutation(result)
-  result.p[].buf.add parLeToken(kind, info)
-  for child in children:
-    result.add(child)
-  result.p[].buf.addParRi()
+  var output = ensureMove(tree)
+  writeTree(output, paramStr(2))
```
planetis-m, line 852: I think we should not hide the fact that toString now takes a var parameter.

planetis-m: sink/var overloads dont work in nimony so there's nothing I can do for now.

planetis-m:
@Araq its ready for review.

doc/plugins.md
```diff
@@ -329,16 +325,29 @@ import plugins
...
+proc loadPluginInput*(filename = paramStr(1)): NifCursor
 ```
 
-Returns a `Node` positioned at the root of the NIF tree. For type plugins, load
-the type definitions with `loadPluginInput(paramStr(3))`.
+Returns a `NifCursor` positioned at the root of the NIF tree. Type plugins use
+`loadPluginInput()` for the module and `loadTypeDefinitions()` for the
+definitions that triggered the plugin. Plugin inputs are densified while
+loading, so every positioned value carries its effective source location rather
+than only nifcore's sparse line-info changes.
```
Araq:
```suggestion
definitions that triggered the plugin.
```

doc/plugins.md
```diff
@@ -381,26 +390,25 @@ A `Node` is a read-only cursor into a frozen NIF tree. It advances forward only.
...
 **Appending tokens:**
 
 | Proc | Description |
 |------|-------------|
-| `addParLe(t, tag, info)` | Opening `(tag` with optional line info |
-| `addParLe(t, tagString, info)` | Same, from textual tag name |
-| `addParRi(t)` | Closing `)` |
+| `openTree(t, tag, info)` | Begin a tagged tree with optional line info |
+| `openTree(t, tagString, info)` | Same, from a textual tag name |
+| `closeTree(t)` | Seal the most recently opened tree |
```
Araq:
```suggestion
| `closeTree(t)` | Close the most recently opened tree |
```

doc/plugins.md
```diff
@@ -542,9 +551,9 @@ The `into`/`loopInto` templates are for pure analysis; the
...
-| `hasMore(n)` | `NifCursor` | true while there are more children before `)` |
-| `into n:` | `NifCursor` | enter node, run body for children, leave |
+| `hasMore(n)` | `NifCursor` | true while the bounded cursor has more values |
+| `into n:` | `NifCursor` | enter a `TagLit`, run the body for its children, leave |
 | `loopInto n:` | `NifCursor` | enter node, iterate all children, leave |
-| `balancedTokens n:` | `NifCursor` | deep-scan all compound nodes in subtree |
-| `keepTag r:` | `Replacer` | copy tag to output, run body, close + skip `)` |
-| `loopKeepTag r:` | `Replacer` | copy tag, iterate all children, close + skip `)` |
+| `balancedTokens n:` | `NifCursor` | scan descendant compound nodes, then advance past the subtree |
+| `keepTag r:` | `Replacer` | copy tag to output, run body, seal tree and advance |
```
Araq:
```suggestion
| `keepTag r:` | `Replacer` | copy tag to output, run body, close tree and advance |
```

doc/plugins.md
```diff
@@ -542,9 +551,9 @@ The `into`/`loopInto` templates are for pure analysis; the
...
+| `loopKeepTag r:` | `Replacer` | copy tag, iterate all children, seal tree and advance |
```
Araq:
```suggestion
| `loopKeepTag r:` | `Replacer` | copy tag, iterate all children, close tree and advance |
```

src/nimony/lib/plugins.nim
```diff
@@ -691,52 +611,56 @@ proc replace*(t: var Replacer; expected: ChildKind; replacement: NifCursor) =
...
   t.dest.addSubtree(replacement)
 
-proc replace*(t: var Replacer; expected: NimonyStmt; replacement: NifBuilder) =
+proc replace*(t: var Replacer; expected: NimonyStmt;
+              replacement: sink NifBuilder) =
   ## Skip one child, asserting a specific statement tag, emit replacement.
   assertTag(t.src, expected)
   t.src.skip()
-  t.dest.add(replacement)
+  t.dest.addTree(ensureMove(replacement))
```
Araq: Silly.

src/nimony/lib/plugins.nim
```diff
@@ -691,52 +611,56 @@ proc replace*(t: var Replacer; expected: ChildKind; replacement: NifCursor) =
...
 
 proc replace*(t: var Replacer; expected: NimonyStmt; replacement: NifCursor) =
   ## Skip one child, asserting a specific statement tag, emit replacement.
   assertTag(t.src, expected)
   t.src.skip()
   t.dest.addSubtree(replacement)
 
-proc replace*(t: var Replacer; expected: NimonyStmt; replacement: NifBuilder) =
+proc replace*(t: var Replacer; expected: NimonyStmt;
+              replacement: sink NifBuilder) =
```
Araq: That is not how `sink` works.

src/nimony/lib/plugins.nim
```diff
@@ -691,52 +611,56 @@ proc replace*(t: var Replacer; expected: ChildKind; replacement: NifCursor) =
...
 
 proc replace*(t: var Replacer; expected: NimonyExpr; replacement: NifCursor) =
   ## Skip one child, asserting a specific expression tag, emit replacement.
   assertTag(t.src, expected)
   t.src.skip()
   t.dest.addSubtree(replacement)
 
-proc replace*(t: var Replacer; expected: NimonyExpr; replacement: NifBuilder) =
+proc replace*(t: var Replacer; expected: NimonyExpr;
+              replacement: sink NifBuilder) =
```
Araq: Likewise.

src/nimony/lib/plugins.nim
```diff
@@ -691,52 +611,56 @@ proc replace*(t: var Replacer; expected: ChildKind; replacement: NifCursor) =
...
   assertTag(t.src, expected)
   t.src.skip()
-  t.dest.add(replacement)
+  t.dest.addTree(ensureMove(replacement))
 
 template keepTag*(t: var Replacer; body: untyped) =
   ## Copy the opening tag from input to output, run `body` for children,
-  ## close the node and advance past the input's closing `)`.
-  assert t.src.kind == ParLe, "keepTag requires cursor at ParLe"
+  ## seal the output node and advance past the input subtree.
```
Araq:
```suggestion
  ## close the output node and advance past the input subtree.
```

PR #2030, opened 2026-06-28
Added support for `or` relation in type constraints

### LINKED ISSUE #2029: `or` sumtypes do not work as type constraints
The following fails to compile in Nimony:

```
import std/syncio

type
  A = concept
    proc `+`(a, b: Self): Self
    proc `*`(a, b: Self): Self

  B = concept
    proc `-`(a, b: Self): Self
    proc `*`(a, b: Self): Self

func foo[T: A | B](a, b: T): T =
  a * b # Error: Type mismatch at [position]

echo foo(1, 2)
```

This seems to be a known limitation as `mtypeclassmatch.nim` is currently disabled.

gmpreussner:
Fixes https://github.com/nim-lang/nimony/issues/2029

### DIFF
--- a/src/nimony/semcall.nim
+++ b/src/nimony/semcall.nim
@@ -185,7 +185,7 @@ iterator findConceptsInConstraint(typ: Cursor): Cursor {.sideEffect.} =
       if section.body.typeKind == ConceptT:
         yield section.body
       inc typ
-    elif typ.typeKind == AndT:
+    elif typ.typeKind in {AndT, OrT}:
       inc typ
       inc nested
     else:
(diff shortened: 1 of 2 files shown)


Araq:
Not sound. The analysis has to look for the overlapping operations of `A | B` and only allow for those. At which point you are better off with a dedicated concept that lists what the generic actually needs.

savannt:
Opened #2056 as a sound alternative that addresses @Araq's concern: it exposes only the **intersection** of the operands' operations for `or` (and the union for `and`), so an operation present in just one alternative is rejected. It also handles the original `int` example from #2029 and arbitrary nesting like `(A and B) or C`.

Araq:
Succeeded by the better implementation.

PR #2040, opened 2026-06-29
Add NIF-native genSym support for plugins

planetis-m:
## Summary

  - Add `genSym()` to the plugin API, returning ordinary `SymId` values usable with `addSymDef` and `addSymUse`.
  - Pass NIF’s standard `.unusedname` directive through plugin input and output.
  - Register generated symbols as fresh when loading plugin output.
  - Add parser support for reading `.unusedname`.
  - Document and test collision-free symbol generation across repeated plugin invocations.

  ## Motivation

  Plugins need hygienic local symbols analogous to Nim macro `genSym`.

  NIF already specifies `.unusedname` for communicating the first symbol name unavailable in the current file. Using that mechanism avoids plugin-specific marker tags
  and integrates symbol generation with Nimony’s existing symbol allocation.

  ## Compatibility

  Plugin output without `.unusedname` remains accepted, preserving compatibility with existing and cached plugins.

  ## Testing

  - `nim c -r src/hastur build nimony`
  - `nim c -r src/hastur test tests/nimony/nifcore/tnifcoreparse.nim`
  - `nim c -r src/hastur test tests/nimony/plugins` — 15/15 passed
  - Fresh-cache Nimony compilation and execution of `tplugin_gensym.nim`


Araq:
For the love of good make the toolchain finally use the NIF spec which its infinite wisdom forsaw this very problem:

- `.unusedname`: Defines the first available symbol for a code generator that does not occur in the current file.

PR #2044, opened 2026-06-29
std/sequtils: add the module (22 symbols, incl. the *It family)

savannt:
## What

Adds `lib/std/sequtils` — the classic module, missing from the Nimony stdlib.
**22 symbols**, fully tested (`tests/nimony/stdlib/tsequtils.nim`, passes `hastur test`).

**procs:** `repeat`, `count`, `deduplicate`, `minIndex`, `maxIndex`, `map`,
`filter`, `keepIf`, `all`, `any`, `apply`, `cycle`, `zip`, `unzip`
**templates:** `foldl`, `foldr`, `anyIt`, `allIt`, `countIt`, `mapIt`,
`filterIt`, `keepItIf`

## Note on the bundled compiler fix

The `mapIt`/`filterIt`/`keepItIf` templates infer their result element type via
`typeof`, which requires the sem fix **"typeof yields the value type"** (also
submitted standalone as #2048). It is merged into this branch so the module
builds green; if #2048 lands first, this rebases cleanly to just the
`lib/std` + test changes.

## Nimony idioms found while porting

- generic `==`/`<` need concept constraints (`Equatable`, `Comparable`);
- build result seqs with `@[]` + `add` (not `newSeq[T](n)`/`setLen`, which need
  `T: HasDefault`);
- `find`/`contains` already live in `system` (openarrays);
- the `*It` family uses `{.untyped.}` + `let it {.inject.} = s[i]` (index
  iteration, since `for it {.inject.} in s` isn't accepted yet);
- a self-referencing `runnableExamples` recurses inside an `{.untyped.}`
  template, so examples are inlined in doc comments.

## Deferred

`concat` (shadowed by a greedy system `concat()` varargs template) and `toSeq`
(needs `for it {.inject.} in iter`).


lib/std/sequtils.nim
```diff
@@ -0,0 +1,343 @@
...
+func delete*[T](s: var seq[T]; first, last: Natural) =
+  ## Deletes the elements `s[first..last]` (inclusive) in place.
+  runnableExamples:
+    var a = @[10, 11, 12, 13, 14]
+    a.delete(1, 2)
+    assert a == @[10, 13, 14]
+  var res: seq[T] = @[]
+  for i in 0 ..< s.len:
+    if i < first or i > last: res.add s[i]
+  s = res
```
Araq: Slow. Do it in-place. This typically involves `.nodestroy` and explicit destruction of the dropped items for best performance.

lib/std/sequtils.nim
```diff
@@ -0,0 +1,343 @@
...
+    dest.insert(@[2, 2], 1)
+    assert dest == @[1, 2, 2, 1, 1]
+  var res: seq[T] = @[]
+  for i in 0 ..< pos:
+    res.add dest[i]
+  for i in 0 ..< src.len:
+    res.add src[i]
+  for i in pos ..< dest.len:
+    res.add dest[i]
+  dest = res
```
Araq: Slow. Do it in-place. This typically involves `.nodestroy` and explicit destruction of the dropped items for best performance.

lib/std/sequtils.nim
```diff
@@ -0,0 +1,343 @@
...
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## Operations on sequences (and `openArray`s), in the spirit of functional
+## programming: `map`, `filter`, `zip`, folds and friends.
+##
+## This is the Nimony port. It currently covers the callback- and
+## value-oriented core; the `mapIt`/`filterIt` template family is not ported
+## yet.
```
Araq:
```suggestion
## value-oriented core.
```

savannt:
Thanks for the review — addressed and pushed:

- **`delete` / `insert` now mutate in place** (no fresh `seq`).
  - `delete(first, last)`: swap the survivors after `last` down over the gap so the dropped elements end up in the tail, then `shrink` — which destroys exactly those dropped elements. No allocation, order preserved.
  - `insert(src, pos)`: `growUnsafe` once, shift the tail up by `src.len` back to front, then fill the gap with `=dup`ed copies of `src`. Marked `.nodestroy` so the raw slot moves aren't rewritten into copies (mirrors `system.grow`).
  - Verified with `seq[string]` and `seq[seq[int]]` (real destructors): interior/head/tail/whole-range deletes and front/end/into-empty inserts round-trip with no leaks or double-frees.
- Applied the module doc-comment suggestion.
- **CI fix:** dropped `tests/nimony/overload/tconcat.nim`. It still asserted the old `system.concat` string behaviour (`concat() == ""`), which was removed in #2053 in favour of `&` chaining — that was the red build. The user-overload coexistence it also covered already lives in `tconcatoverload.nim` on master. Also merged current master.

PR #2045, opened 2026-06-29
std/heapqueue: add a binary min-heap

savannt:
## What

Adds `lib/std/heapqueue` — a generic binary **min-heap / priority queue**
(`HeapQueue[T]`) over a `seq`, another module missing from the Nimony stdlib.

API: `initHeapQueue`, `len`, `[]`, `push`, `pop`, `clear`, `toHeapQueue`.
The element type is `Comparable`. `heap[0]` is always the smallest element, so
`pop` returns elements in ascending order.

## Nimony-specific choice

Sifting moves elements with **`swap`**, not `s[i] = s[j]`: Nimony's alias
analysis rejects assigning one element of a `var` seq from another (`mutable
argument aliases with immutable parameter`). This is the same `swap`-based idiom
`algorithm.sort` uses. `pop` removes the last element via `system.pop`, which
(unlike `setLen`) needs no `HasDefault` constraint.

## Test

`tests/nimony/stdlib/theapqueue.nim` pushes a shuffled set, drains via `pop`,
and asserts ascending order, plus `toHeapQueue` and `clear`. Passes via
`hastur test`.

### DIFF
--- a/lib/std/heapqueue.nim
+++ b/lib/std/heapqueue.nim
@@ -1,0 +1,103 @@
+#
+#
+#            Nim's Runtime Library
+#        (c) Copyright 2024 Nim contributors
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## A min-heap (priority queue) implemented as a binary heap over a `seq`.
+##
+## The smallest element (per `<`) is always at index 0, so `pop` returns
+## elements in ascending order. The element type must be `Comparable`.
+
+import std/[assertions]
+
+type
+  HeapQueue*[T] = object
+    ## A priority queue. Use `initHeapQueue` to create one.
+    data: seq[T]
+
+func initHeapQueue*[T](): HeapQueue[T] =
+  ## Creates a new empty heap.
+  HeapQueue[T](data: @[])
+
+func len*[T](heap: HeapQueue[T]): int {.inline.} =
+  ## Number of elements in `heap`.
+  heap.data.len
+
+func `[]`*[T](heap: HeapQueue[T]; i: Natural): T {.inline.} =
+  ## Accesses the `i`-th element of `heap`'s internal array. `heap[0]` is the
+  ## smallest element.
+  heap.data[i]
+
+proc siftUp[T: Comparable](heap: var HeapQueue[T]; p: int) =
+  ## Bubbles the element at `p` up toward the root until the heap property
+  ## holds. Uses `swap` to move elements (Nimony rejects `s[i] = s[j]` on a
+  ## `var` seq as an aliasing copy).
+  var pos = p
+  while pos > 0:
+    let parent = (pos - 1) shr 1
+    if heap.data[pos] < heap.data[parent]:
+      swap(heap.data[pos], heap.data[parent])
+      pos = parent
+    else:
+      break
+
+proc siftDown[T: Comparable](heap: var HeapQueue[T]; p: int) =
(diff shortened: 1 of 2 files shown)


lib/std/heapqueue.nim
```diff
@@ -0,0 +1,146 @@
...
+
+func initHeapQueue*[T](): HeapQueue[T] =
+  ## Creates a new empty heap.
+  HeapQueue[T](data: @[])
+
+func len*[T](heap: HeapQueue[T]): int {.inline.} =
+  ## Number of elements in `heap`.
+  heap.data.len
+
+func `[]`*[T](heap: HeapQueue[T]; i: Natural): T {.inline.} =
```
Araq:
```suggestion
func `[]`*[T](heap: HeapQueue[T]; i: Natural): var T {.inline.} =
```

lib/std/heapqueue.nim
```diff
@@ -0,0 +1,146 @@
...
+    if left < endpos and heap.data[left] < heap.data[smallest]:
+      smallest = left
+    if right < endpos and heap.data[right] < heap.data[smallest]:
+      smallest = right
+    if smallest == pos:
+      break
+    swap(heap.data[pos], heap.data[smallest])
+    pos = smallest
+
+proc push*[T: Comparable](heap: var HeapQueue[T]; item: T) =
```
Araq:
```suggestion
proc push*[T: Comparable](heap: var HeapQueue[T]; item: sink T) =
```

lib/std/heapqueue.nim
```diff
@@ -0,0 +1,146 @@
...
+proc toHeapQueue*[T: Comparable](xs: openArray[T]): HeapQueue[T] =
+  ## Builds a heap from the elements of `xs`.
+  runnableExamples:
+    var h = toHeapQueue([3, 1, 2])
+    assert h.pop() == 1
+  result = initHeapQueue[T]()
+  for i in 0 ..< xs.len:
+    result.push(xs[i])
+
+iterator items*[T](heap: HeapQueue[T]): T =
```
Araq:
```suggestion
iterator items*[T](heap: HeapQueue[T]): var T =
```

lib/std/heapqueue.nim
```diff
@@ -0,0 +1,146 @@
...
+    discard heap.data.pop()
+  else:
+    swap(heap.data[index], heap.data[last])
+    discard heap.data.pop()
+    # the moved-in element belongs either up or down; restoring both ways is
+    # safe because exactly one direction applies (the rest of the heap is valid).
+    siftUp(heap, index)
+    siftDown(heap, index)
+
+proc replace*[T: Comparable](heap: var HeapQueue[T]; item: T): T =
```
Araq:
```suggestion
proc replace*[T: Comparable](heap: var HeapQueue[T]; item: sink T): T =
```

lib/std/heapqueue.nim
```diff
@@ -0,0 +1,146 @@
...
+  ## Pops and returns the smallest element, then pushes `item` — more efficient
+  ## than `pop` + `push`. Requires a non-empty heap. The returned value may be
+  ## larger than `item`.
+  result = heap.data[0]
+  heap.data[0] = item
+  siftDown(heap, 0)
+
+proc pushpop*[T: Comparable](heap: var HeapQueue[T]; item: T): T =
```
Araq:
```suggestion
proc pushpop*[T: Comparable](heap: var HeapQueue[T]; item: sink T): T =
```

PR #1968, opened 2026-06-17
std: add sum-type Opt[T] and Result[T,E]

tokyovigilante:
New stdlib modules on Nimony's native sum types:

  std/opt    — Opt[T] {None, Some}: some/none[T]/isSome/isNone/get(default)/
               unsafeGet (raises BadOperation on None). None carries a phantom
               field and none[T]() constructs through `result` (generic
               inference can't synthesise a fieldless variant's T).
  std/either — Either[L, R] {Left, Right} + Result[T, E] = Either[E, T]:
               left/right/ok/err, isLeft/isRight/isOk/isErr, getLeft/getRight/
               get/error (total), unsafeGet*. Right/Ok = value, Left/Err = error.

~~Toolchain limitation documented: a two-parameter sum type's constructor can't infer its non-argument parameter from the result type yet, so Either/Result constructors take both explicitly (ok[int,string](v)); single-parameter opt.some needs none. Drop the annotations when inference improves.~~
[EDIT] Toolchain fixes so that inference and empty `None[T]` enum branch works.

Verified: full-API smoke over int + a custom object — Opt some/none/get, Result ok/err/get/error, Either left/right, unsafeGet raises on the empty case.

PR #2047, opened 2026-06-29
std/options: add Nim-2-compatible Option[T]

savannt:
Ports `std/options` to Nimony: the Nim-2-compatible, flag-based `Option[T]`.

> **Relationship to #1968.** #1968 adds a *sum-type* `Opt[T]` (a native `case` object), which its author explicitly frames as "distinct from a Nim 2-compatible `Option[T]`." This PR is that counterpart: a flag-based `Option[T]` (`val` + `has: bool`) with the familiar Nim-2 surface (`map`/`filter`/`flatMap`/`flatten`). The two can coexist (`std/opt` vs `std/options`, `Opt` vs `Option`), but I'm happy to consolidate or rename if maintainers prefer a single optional type.

## API

- Construction: `some`, `none`, `option`
- Predicates: `isSome`, `isNone`
- Access: `get` (raise-on-empty), `get(opt, default)`, `unsafeGet`
- `$`
- Combinators: `map`, `filter`, `flatMap`, `flatten`

## Nimony specifics

- Generic bodies are bound eagerly, so operations on the element type are gated by concepts: `$` via a local `Stringable`, and the empty-`Option` producers via `HasDefault` (satisfied by essentially every concrete type).
- Equality reuses the compiler-provided structural `==` for objects: an empty `Option` always stores `default(T)`, so two empty Options compare equal and an empty one never equals a populated one.
- Exceptions are not wired up yet, so `get` on an empty `Option` terminates the program (mirroring `assert`) rather than raising `UnpackDefect`.

## Test

`tests/nimony/stdlib/toptions.nim` covers construction, accessors, equality, stringify, and the combinators. Passes via `hastur test`.

## Note — `Stringable`

`$` is gated by a module-local `Stringable` concept (eager generic binding means `$` on the element type must be constrained). It's kept local here, but I've proposed promoting it into `system` alongside `Equatable`/`HasDefault`/`Orderable` in **#2051**. That exported concept coexists with this private copy (the local definition shadows the imported one), so there's no merge-order dependency between the two PRs; once #2051 lands, the local copy here can simply be removed.

### DIFF
--- a/lib/std/options.nim
+++ b/lib/std/options.nim
@@ -1,0 +1,104 @@
+#
+#
+#            Nim's Runtime Library
+#        (c) Copyright 2024 Nim contributors
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## This module implements types which encapsulate an optional value.
+##
+## A value of type `Option[T]` either contains a value `x` (represented as
+## `some(x)`) or is empty (`none(T)`).
+##
+## This is the Nimony port. Since exceptions are not yet wired up, `get` on an
+## empty `Option` terminates the program (mirroring `assert`) rather than
+## raising `UnpackDefect`. Equality is the compiler-provided structural
+## comparison: an empty `Option` always stores the default value, so two empty
+## `Option`s compare equal and an empty one never equals a populated one.
+
+import std/syncio
+
+type
+  Option*[T] = object
+    ## An optional type that may or may not contain a value of type `T`.
+    val: T
+    has: bool
+
+  Stringable = concept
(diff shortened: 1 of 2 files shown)


lib/std/options.nim
```diff
@@ -0,0 +1,104 @@
...
+
+type
+  Option*[T] = object
+    ## An optional type that may or may not contain a value of type `T`.
+    val: T
+    has: bool
+
+  Stringable = concept
+    ## A type that can be rendered with `$`.
+    func `$`(x: Self): string
```
Araq:
```suggestion
```

PR #2049, opened 2026-06-29
std/deques: add a ring-buffer Deque[T]

savannt:
Ports `std/deques` to nimony: a double-ended queue (`Deque[T]`) backed by a power-of-two ring buffer over `seq[T]`. Amortized O(1) push/pop at both ends, O(1) indexing.

### API
`initDeque` / `toDeque`, `addFirst` / `addLast`, `popFirst` / `popLast`, `peekFirst` / `peekLast`, `[]` / `[]=`, `len`, `clear`, `contains` (`in`/`notin`), `$`, and `items` / `mitems` / `pairs` iterators.

### Adaptations to nimony
nimony binds generic proc bodies eagerly, so operations on the element type are gated by concepts rather than resolved at instantiation:
- `HasDefault` where the backing `seq` is (re)allocated (`initDeque`, the grow path, `addFirst`/`addLast`, `toDeque`);
- `Equatable` for `contains`;
- a local `Stringable` concept for `$` (see the note below — proposed for promotion into `system` in #2051; same local approach as the `std/options` port #2047).

Other deltas from the Nim version, all consistent with the rest of the nimony stdlib:
- getters (`peekFirst`/`peekLast`/`[]`) return `T` by value instead of `lent T` (borrowing a `lent` result from a by-value parameter is rejected);
- exceptions aren't wired up yet, so accessing an empty deque or an out-of-range index terminates the program (mirroring `assert`) instead of raising `IndexDefect` — the same `quit`-based stand-in used elsewhere (e.g. `parseutils`, the `std/options` port).

### Tests
`tests/nimony/stdlib/tdeques.nim` covers ordering, `addFirst`/`addLast` across the ring's wrap boundary, growth past the initial capacity (100 elements into a cap-2 deque), `[]`/`[]=`, `peek`/`pop` at both ends, `contains`, the iterators, `mitems` in-place mutation, string elements, `toDeque`, and `clear`. Passes via `hastur` (`SUCCESS tests/nimony/stdlib/tdeques.nim`).

### Note — `Stringable`

The `$` gate uses a module-local `Stringable` concept. I've proposed adding a canonical `Stringable` to `system` in **#2051** (alongside `Equatable`/`HasDefault`). It coexists with this private copy (the local definition shadows the imported one), so there's no merge-order dependency; once #2051 lands, the local copy here can simply be removed.

### DIFF
--- a/lib/std/deques.nim
+++ b/lib/std/deques.nim
@@ -1,0 +1,207 @@
+#
+#
+#            Nim's Runtime Library
+#        (c) Copyright 2024 Nim contributors
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## An implementation of a `deque`:idx: (double-ended queue).
+## The underlying implementation uses a `seq`.
+##
+## This is the Nimony port. The container is a growable, power-of-two ring
+## buffer: elements are appended/prepended in amortized O(1) and indexed in
+## O(1). Since exceptions are not yet wired up, accessing an element of an
+## empty `Deque` (or an out-of-range index) terminates the program (mirroring
+## `assert`) rather than raising `IndexDefect`.
+
+import std/syncio
+
+const
+  defaultInitialSize = 4
+
+type
+  Deque*[T] = object
+    ## A double-ended queue backed by a ring buffer whose capacity is always a
+    ## power of two. Logical element `i` lives at `data[(head + i) and mask]`.
+    data: seq[T]
+    head, tail, count, mask: int
+
+  Stringable = concept
+    ## A type that can be rendered with `$`.
+    func `$`(x: Self): string
+
+func raiseEmpty() {.noinline, noreturn.} =
+  {.cast(noSideEffect).}: quit "Deque is empty"
+
+func raiseIndex() {.noinline, noreturn.} =
+  {.cast(noSideEffect).}: quit "Deque index out of bounds"
+
+func nextPowerOfTwo(x: int): int =
+  ## Smallest power of two `>= x` (at least 1).
+  result = 1
+  while result < x:
+    result = result shl 1
+
+func initDeque*[T: HasDefault](initialSize: int = defaultInitialSize): Deque[T] =
+  ## Creates a new empty `Deque`.
+  ##
+  ## `initialSize` is rounded up to the next power of two and used as the
+  ## initial capacity, avoiding reallocations when the final size is known.
+  runnableExamples:
+    var a = initDeque[int]()
+    assert a.len == 0
+  let cap = nextPowerOfTwo(initialSize)
+  result = Deque[T](data: newSeq[T](cap), head: 0, tail: 0, count: 0, mask: cap - 1)
+
+func len*[T](d: Deque[T]): int {.inline.} =
+  ## Returns the number of elements of `d`.
+  d.count
+
+func growIfNeeded[T: HasDefault](d: var Deque[T]) =
+  ## Doubles the capacity if the ring buffer is full, re-laying the elements
+  ## out contiguously starting at index 0.
+  if d.count < d.data.len:
+    return
+  let oldCap = d.data.len
+  let newCap = if oldCap == 0: defaultInitialSize else: oldCap * 2
+  var newData = newSeq[T](newCap)
+  var i = 0
+  var j = d.head
+  while i < d.count:
+    newData[i] = move(d.data[j])
+    j = (j + 1) and d.mask
+    inc i
+  d.data = newData
+  d.head = 0
+  d.tail = d.count
+  d.mask = newCap - 1
+
+func addLast*[T: HasDefault](d: var Deque[T]; item: sink T) =
+  ## Adds an `item` to the end of `d`.
+  runnableExamples:
+    var a = initDeque[int]()
+    for i in 1 .. 3: a.addLast(i)
+    assert $a == "[1, 2, 3]"
+  growIfNeeded(d)
+  d.data[d.tail] = item
+  d.tail = (d.tail + 1) and d.mask
+  inc d.count
+
+func addFirst*[T: HasDefault](d: var Deque[T]; item: sink T) =
+  ## Adds an `item` to the beginning of `d`.
+  runnableExamples:
+    var a = initDeque[int]()
+    for i in 1 .. 3: a.addFirst(i)
+    assert $a == "[3, 2, 1]"
+  growIfNeeded(d)
+  d.head = (d.head - 1) and d.mask
+  d.data[d.head] = item
+  inc d.count
+
+func peekFirst*[T](d: Deque[T]): T =
+  ## Returns the first element of `d`. Terminates the program if `d` is empty.
+  if d.count == 0:
+    raiseEmpty()
+  result = d.data[d.head]
+
+func peekLast*[T](d: Deque[T]): T =
+  ## Returns the last element of `d`. Terminates the program if `d` is empty.
+  if d.count == 0:
+    raiseEmpty()
+  result = d.data[(d.tail - 1) and d.mask]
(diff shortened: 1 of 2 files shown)


lib/std/deques.nim
```diff
@@ -0,0 +1,207 @@
...
+type
+  Deque*[T] = object
+    ## A double-ended queue backed by a ring buffer whose capacity is always a
+    ## power of two. Logical element `i` lives at `data[(head + i) and mask]`.
+    data: seq[T]
+    head, tail, count, mask: int
+
+  Stringable = concept
+    ## A type that can be rendered with `$`.
+    func `$`(x: Self): string
```
Araq:
```suggestion
```

lib/std/deques.nim
```diff
@@ -0,0 +1,207 @@
...
+
+func raiseEmpty() {.noinline, noreturn.} =
+  {.cast(noSideEffect).}: quit "Deque is empty"
```
Araq: Use `assert`

lib/std/deques.nim
```diff
@@ -0,0 +1,207 @@
...
+
+func raiseIndex() {.noinline, noreturn.} =
+  {.cast(noSideEffect).}: quit "Deque index out of bounds"
```
Araq: Use `assert`

lib/std/deques.nim
```diff
@@ -0,0 +1,207 @@
...
+  runnableExamples:
+    var a = initDeque[int]()
+    for i in 1 .. 3: a.addFirst(i)
+    assert $a == "[3, 2, 1]"
+  growIfNeeded(d)
+  d.head = (d.head - 1) and d.mask
+  d.data[d.head] = item
+  inc d.count
+
+func peekFirst*[T](d: Deque[T]): T =
```
Araq:
```suggestion
func peekFirst*[T](d: Deque[T]): var T =
```

lib/std/deques.nim
```diff
@@ -0,0 +1,207 @@
...
+  ## Returns the first element of `d`. Terminates the program if `d` is empty.
+  if d.count == 0:
+    raiseEmpty()
+  result = d.data[d.head]
+
+func peekLast*[T](d: Deque[T]): T =
```
Araq:
```suggestion
func peekLast*[T](d: Deque[T]): var T =
```

lib/std/deques.nim
```diff
@@ -0,0 +1,207 @@
...
+func popLast*[T](d: var Deque[T]): T =
+  ## Removes and returns the last element of `d`. Terminates the program if
+  ## `d` is empty.
+  if d.count == 0:
+    raiseEmpty()
+  d.tail = (d.tail - 1) and d.mask
+  result = move(d.data[d.tail])
+  dec d.count
+
+func `[]`*[T](d: Deque[T]; i: int): T =
```
Araq:
```suggestion
func `[]`*[T](d: Deque[T]; i: int): var T =
```

lib/std/deques.nim
```diff
@@ -0,0 +1,207 @@
...
+    var a = initDeque[int]()
+    for i in 1 .. 3: a.addLast(i)
+    assert 2 in a
+    assert 7 notin a
+  for i in 0 ..< d.count:
+    if d.data[(d.head + i) and d.mask] == item:
+      return true
+  result = false
+
+iterator items*[T](d: Deque[T]): T =
```
Araq:
```suggestion
iterator items*[T](d: Deque[T]): var T =
```

lib/std/deques.nim
```diff
@@ -0,0 +1,207 @@
...
+  ## Yields every element of `d`, from first to last.
+  for i in 0 ..< d.count:
+    yield d.data[(d.head + i) and d.mask]
+
+iterator mitems*[T](d: var Deque[T]): var T =
+  ## Yields every element of `d` by reference, allowing modification in place.
+  for i in 0 ..< d.count:
+    yield d.data[(d.head + i) and d.mask]
+
+iterator pairs*[T](d: Deque[T]): (int, T) =
```
Araq:
```suggestion
iterator pairs*[T](d: Deque[T]): (int, var T) =
```

lib/std/deques.nim
```diff
@@ -0,0 +1,207 @@
...
+  result = "["
+  var first = true
+  for i in 0 ..< d.count:
+    if not first:
+      result.add ", "
+    first = false
+    result.add $d.data[(d.head + i) and d.mask]
+  result.add "]"
+
+func toDeque*[T: HasDefault](xs: openArray[T]): Deque[T] =
```
Araq: `HasDefault` constraint must be dropped.

savannt:
Done: `quit`→`assert` at the access sites; rebased onto master (the branch predated the `Stringable` concept the `$` uses, so it no longer compiled). The `var T` accessor suggestions were already applied.

On dropping `HasDefault`: it can't come off while the ring uses `newSeq` (which requires it). Switching to `newSeqUninit` and `=wasMoved`-ing the unused slots hits a backend bug — `[Bug] call must have been bound to a location` in nj.nim on `=wasMoved(buf[i])` (wasMoved of a seq-element access). Dropping the constraint cleanly seems to need a raw `UncheckedArray` backing with a custom `=destroy` that only frees the live `[head, head+count)` slots. Want me to do that reimplementation, or is there a simpler primitive you'd prefer I use?

Araq:
> `[Bug] call must have been bound to a location` in nj.nim

Fix this bug, the time to workaround compiler bugs is over.

savannt:
Fixed the compiler bug rather than working around it — #2081. `nj.trExpr` was rejecting an lvalue call in a builtin's operand position; it now binds it via `trCall`. With that in, dropping `HasDefault` here works (`newSeqUninit` + `=wasMoved` the unused slots). I'll add that to this PR once #2081 lands.

lib/std/deques.nim
```diff
@@ -0,0 +1,192 @@
...
+    for i in 1 .. 3: a.addFirst(i)
+    assert $a == "[3, 2, 1]"
+  growIfNeeded(d)
+  d.head = (d.head - 1) and d.mask
+  d.data[d.head] = item
+  inc d.count
+
+func peekFirst*[T](d: Deque[T]): var T =
+  ## Returns the first element of `d`. Terminates the program if `d` is empty.
+  {.cast(noSideEffect).}: assert d.count > 0, "Deque is empty"
```
Araq: Use `.requires`.

lib/std/deques.nim
```diff
@@ -0,0 +1,192 @@
...
+  result = d.data[d.head]
+
+func peekLast*[T](d: Deque[T]): var T =
+  ## Returns the last element of `d`. Terminates the program if `d` is empty.
+  {.cast(noSideEffect).}: assert d.count > 0, "Deque is empty"
```
Araq: Likewise for all the others.

savannt:
Addressed all remarks and rebased onto master (which now has #2081):

- `.requires` contracts replace the `{.cast(noSideEffect).}: assert` guards on `peekFirst`/`peekLast`/`popFirst`/`popLast`/`[]`/`[]=`.
- **`HasDefault` dropped.** With #2081 in, the ring can be allocated with `newSeqUninit` and every slot marked moved-out via `=wasMoved` (`newRingBuf`), so a dead slot stays in the moved-from state — safe to destroy and to `=sink`-overwrite. Elements are never default-initialized, so `T` needs no default. Removed from `initDeque`/`growIfNeeded`/`addLast`/`addFirst`/`toDeque`.
- Dropped the unused `std/syncio` import.

`var T` accessors were already applied. Verified `tests/nimony/stdlib/tdeques.nim` plus a larger local test (string elements, 100-element growth crossing reallocation, `mitems`/`pairs`) all pass.

PR #2051, opened 2026-06-30
system: add a Stringable concept

savannt:
Adds a `Stringable*` concept to `system`, beside the existing `HasDefault` / `Equatable` / `Orderable` / `Comparable`:

```nim
type Stringable* = concept
  func `$`(x: Self): string
```

### Why

nimony binds generic proc bodies eagerly, so generic code that stringifies an abstract type parameter `T` must gate `$` behind a concept (a `mixin` does not defer resolution). With no shared concept, every such module reinvents a byte-identical private copy. Two already do:

- `std/options` (#2047) — `$*[T: Stringable](self: Option[T])`
- `std/deques` (#2049) — `$*[T: Stringable](d: Deque[T])`

This gives them (and future modules) one canonical concept to depend on, mirroring how `Equatable` already backs the generic `==`/`contains` machinery.

### No merge-order dependency

The exported concept **coexists** with those private copies — the local definition simply shadows the imported one — so this PR can merge in any order relative to #2047/#2049 without breaking them. Removing the now-redundant local copies from those modules is left as a trivial follow-up once this lands (and is intentionally *not* done here, since those files don't exist on `master` yet).

### Tests

`tests/nimony/stdlib/tstringable.nim` exercises the concept from a fresh module (no local copy) over `int`, `string`, and a user-defined object type. Verified with `nimony c` (`stringable: OK`).

### DIFF
--- a/lib/std/system.nim
+++ b/lib/std/system.nim
@@ -119,6 +119,12 @@ func `$`*(b: bool): string =
 func `$`*[T: enum](x: T): string {.magic: "EnumToStr", noSideEffect.}
   ## Converts an enum value to a string.
 
+type Stringable* = concept
+  ## A type that can be rendered as a `string` via the `$` operator. Generic
+  ## code that stringifies an abstract type parameter `T` should constrain it
+  ## with `[T: Stringable]`, since nimony binds generic bodies eagerly.
+  func `$`(x: Self): string
+
 func addr*[T](x: T): ptr T {.magic: "Addr", noSideEffect.}
   ## Returns the address of `x`.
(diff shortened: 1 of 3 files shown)


Araq:
Run `hastur --overwrite test tests/nimony/errmsgs/tdistinctinfos.nim` to make the CI green.

savannt:
Done — regenerated `tdistinctinfos.msgs` with `hastur --overwrite` (on top of the master merge, with the toolchain rebuilt). The only change is the two shifted declaration line numbers (`system.nim` 345->351, 351->357) from the lines the concept adds; the test passes locally now.

PR #2053, opened 2026-06-30
system: remove concat in favor of efficient & chaining

savannt:
## What

Removes the `concat` template from `system` entirely and switches the few
callers to chained `&`.

## Why

`system.concat` was declared as a bare 0-parameter `{.varargs.}` template:

```nim
template concat*(): string {.varargs.} =
  var res = ""
  for s in unpack():
    res.add s
  res
```

Because it takes no typed parameters, it matched **every** `concat(...)` call
and only failed later inside its body for non-string arguments. That shadowed
any user-defined `concat` overload (e.g. a `sequtils.concat` over
`openArray[T]`), so such an overload could never be selected.

Following @Araq's guidance on the previous revision of this PR: now that
chaining `&` is efficient, the right fix is to **drop `concat` from `system`
altogether** rather than retype it as `varargs[string]` (which would generate
worse code). Removing it eliminates the namespace conflict with `sequtils` at
the source.

## Changes

- Delete `template concat*` from `lib/std/system/stringimpl.nim`.
- Switch the in-tree callers to `&`:
  - `lib/std/dirs.nim` (`$w.dir & "/" & name`)
  - `examples/seqs_and_strings.nim`
  - `tests/nimony/arc/tconcat.nim`
  - `tests/nimony/methods/deps/mgenericinheritance.nim`
- Add `tests/nimony/overload/tconcatoverload.nim`: asserts a user-defined
  `concat[T](a, b: openArray[T])` overload now compiles and runs without being
  shadowed by `system`.

## Testing

`hastur test` green on the affected areas: overload (10/10), arc (40/40),
methods (8/8), stdlib (33/33, incl. `twalkdir`), examples (6/6).

### DIFF
--- a/examples/seqs_and_strings.nim
+++ b/examples/seqs_and_strings.nim
@@ -76,12 +76,11 @@ var s2 = s1
 s2[0] = 'X'
 assert s1 == "abc"   # s1 is unchanged
 
-# --- concat for efficient string building ---
+# --- string building with `&` ---
 
-# `&` allocates a new string each time.
-# `concat` is much more efficient for multiple pieces:
+# Chained `&` builds a new string from multiple pieces:
 let name = "world"
-let msg = concat("hello, ", name, "!")
+let msg = "hello, " & name & "!"
 assert msg == "hello, world!"
 
 # --- Bulk writes with beginStore / endStore ---
--- a/lib/std/dirs.nim
+++ b/lib/std/dirs.nim
@@ -155,7 +155,7 @@ proc fillDirEntry(w: var DirWalker; e: var DirEntry) =
 when defined(posix):
   proc pathComponentFromEntry(w: var DirWalker; name: string; dType: uint8): PathComponent =
     if dType == DT_UNKNOWN or dType == DT_LNK:
-      var fullPath = concat($w.dir, "/", name)
+      var fullPath = $w.dir & "/" & name
       var s = default Stat
       if lstat(fullPath.toCString, s) == 0:
         if S_ISLNK(s.st_mode):
--- a/lib/std/system/stringimpl.nim
+++ b/lib/std/system/stringimpl.nim
@@ -675,13 +675,7 @@ func newStringOfCap*(len: int): string =
   else:
     strOom result, LongStringDataOffset + len
 
-# ---- concat / & ----
-
-template concat*(): string {.varargs.} =
-  var res = ""
-  for s in unpack():
-    res.add s
-  res
+# ---- & ----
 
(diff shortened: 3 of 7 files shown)


Araq:
This is a bad solution. It makes `concat` produce much worse code. But now that chaining `&` has become efficient we can remove this `concat` altogether to keep it from conflicting with sequtils.

savannt:
Thanks — reworked along the lines you suggested. Instead of retyping `concat` as `varargs[string]`, I've **removed `concat` from `system` entirely** and switched the handful of in-tree callers to chained `&`. That drops the worse codegen concern and removes the conflict with `sequtils` at the source.

`hastur test` is green on the affected areas (overload, arc, methods, stdlib, examples). Force-pushed the rewrite.

Araq:
> hastur test is green on the affected areas (overload, arc, methods, stdlib, examples). Force-pushed the rewrite.

These things keep messing up the tests though that hardcode line information...

savannt:
Good news on the hardcoded-line-info front: rebased onto current `master` (which now has #2051), and **no golden tests need touching**. The `concat` block I removed lived at `stringimpl.nim` line ~675, but the `.msgs` candidate lists only reference declarations *above* it (`stringimpl.nim` 587/604 and `system.nim` 351/357, the latter already adjusted by the #2051 merge), so removing it shifts nothing they print.

Verified on a clean cache:
- `errmsgs` 29/29
- `overload` 10/10, `arc` 40/40, `methods` 8/8, `stdlib` 34/34

So this one's clear of the line-info churn — nothing to regenerate.

Araq:
[boot] stage 1: bin0/nimony c --silentMake --nimcache:nimcache/boot/nimsem --out:bin1/nimsem src/nimony/nimsem.nim
Error: src/nimony/sigmatch.nim(178, 11) Error: undeclared identifier: 'concat'
Error: src/nimony/sigmatch.nim(180, 11) Error: undeclared identifier: 'concat'
Error: src/nimony/sigmatch.nim(183, 11) Error: undeclared identifier: 'concat'
Error: src/nimony/sigmatch.nim(200, 11) Error: undeclared identifier: 'concat'
Error: src/nimony/sigmatch.nim(202, 11) Error: undeclared identifier: 'concat'
Error: src/nimony/sigmatch.nim(205, 11) Error: undeclared identifier: 'concat'
Error: src/nimony/sigmatch.nim(207, 11) Error: undeclared identifier: 'concat'
Error: src/nimony/sigmatch.nim(209, 11) Error: undeclared identifier: 'concat'
Error: src/nimony/sigmatch.nim(215, 11) Error: undeclared identifier: 'concat'

PR #2054, opened 2026-06-30
nimsem: refine nil-tracking through `and`/`or` short-circuit guards (#1984)

### LINKED ISSUE #1984: Nimony nil-tracking fails through `or` short-circuit with custom pragma
### Reproducer

```nim
{.pragma: sql, cdecl.}

proc getA(): nil cstring {.sql, importc.}
proc getB(): nil cstring {.sql, importc.}
proc use(a, b: cstring) {.sql, importc.}

proc bug() =
  let a = getA()
  let b = getB()
  if a == nil or b == nil:
    return
  use(a, b)
```

```
Error: cannot prove expression is not nil: a
Error: cannot prove expression is not nil: b
```

### Expected

After `if a == nil or b == nil: return`, both `a` and `b` are proven non-nil. `use(a: cstring, b: cstring)` should accept them.

### Actual

Nimony's nil-tracking fails to refine `nil cstring` → `cstring` through `or` short-circuit returns when a module-level custom pragma is active.

### Trigger

The module-level `{.pragma: sql, cdecl.}` pragma. Without it, identical code compiles fine. With it, nil-tracking degrades.

### Workaround

Separate the nil checks into individual branches:

```nim
if a == nil:
  return
if b == nil:
  return
use(a, b)  # OK
```

savannt:
Fixes #1984.

## The bug (and a correction to the report)

The issue is filed as a `{.pragma.}`-specific nil-tracking failure, but **the pragma is a red herring**. With every pragma stripped, this still fails:

```nim
proc getA(): nil cstring = nil
proc getB(): nil cstring = nil
proc use(a, b: cstring) = discard      # params are non-nil
proc bug() =
  let a = getA()
  let b = getB()
  if a == nil or b == nil:
    return
  use(a, b)   # Error: cannot prove expression is not nil: a / b
```

The real trigger is **any `and`/`or` combination of nil-checks**. The positive `and` form fails identically:

```nim
if a != nil and b != nil:
  use(a, b)   # same error
```

while the single-check form (`if a == nil: return`) and the split form (two separate `if`s) both work.

## Root cause

`and`/`or` short-circuit guards lower (via the goto→structured relooper) to a **join cfvar** `cf`: `cf = (a==nil) or (b==nil); if cf: return else: use(a, b)`. NJVL tracks two things — linear `le`/not-nil facts (`c.facts`, branch-scoped) and cfvar **init**-implications (`implications.nim`). The per-disjunct nil facts `a != nil`/`b != nil` are derived inside the disjunct sub-branches and **intersected away at each merge**; only `cf`'s value survives. Nothing connects `cf`'s truth value back to the operands' nil-ness, so the guarded code can't be refined. (The single/split forms work because they branch directly on `(eq nil x)`, whose negation is added to the else-branch facts.)

## Fix

A thin **cfvar-conditioned linear-fact** layer that bridges the init-implication system and the le/not-nil fact base. When an `ite` branch leaves via `jtrue cf`, the *other* branch's freshly-derived facts are recorded against `cf`'s truth value, and re-materialised when `cf` becomes known (inside the `if cf` the lowering produces).

Soundness is **structural**, gated by a per-cfvar count of `jtrue` sites:

| flag | jtrue sites | single path | captured polarity |
|---|---|---|---|
| `and` (conjunction) | 1 | `cf = true` | facts at the `jtrue` site, conditioned on `cf=true` |
| `or` (disjunction) | ≥ 2 | `cf = false` | per-branch negations at the merge, conditioned on `cf=false` |

The unsound polarity (a disjunction over multiple paths, where no single fact set holds) is **never** captured. Conditioned facts are invalidated on reassignment, dropped across loop-body boundaries, and reset per proc.

## Why it's safe

The change is **monotonic** — it only ever *adds* provable facts, and `implies` is monotonic — so it cannot turn a previously-accepted program into a rejected one (no false-reject regressions are possible). The only regression class would be a nil/contract-error test losing its error; the suite's nil/init error tests (`object/tcase2`, `errmsgs/tcommontype`) still error correctly.

Adversarial cases confirm no false-accepts: an `or`/`and` guard **without** an early return, a partial guard, an unrelated variable, the `and`-`else` branch, and reassignment-after-guard are all still correctly rejected.

## Tests

`tests/nimony/notnil/tandor.nim` covers both the `or`-with-return and positive-`and` forms (runs, asserting refined access succeeds). The `notnil`, `borrows`, `rtchecks`, `object`, and `sysbasics` suites pass.


Araq:
The NJ IR has outlived its purpose and https://github.com/nim-lang/nimony/issues/1946 is much better at this and will simplify the subsystem.

savannt:
Agreed — this fix layers a cfvar-conditioned fact bridge onto the NJ IR, and per #1946 that subsystem is being replaced. The Final IR handles this case structurally: the guarded transfer `(if cond (jmp L))` keeps both edges path-sensitive (`assume(cond)` / `assume(¬cond)`), so `and`/`or` nil-guards refine their operands with no special-casing. Closing in favor of that work. #1984 stays open and will be resolved by the Final IR.

PR #2058, opened 2026-06-30
nimsem: resolve overloaded routine passed to a proc-typed parameter (#1973)

### LINKED ISSUE #1973: Overloaded procedures doesn't match procedural type parameter
Following code fails to compile:
```nim
proc foo(x, y: string): int = 0
proc bar(x, y: string): int = 0
proc bar(x, y: int): int = 0

proc test(x: proc (x, y: string): int) = discard

# compiles without errors
test(foo)

# Error: Type mismatch
test(bar)
```

`sort` proc in algorithm module in Nim 2 uses `system.cmp[T]` passes it to another `sort` proc.

savannt:
Fixes #1973.

## Problem

Passing an **overloaded** routine where a `proc` type is expected fails:

```nim
proc bar(x, y: string): int = 0
proc bar(x, y: int): int = 0
proc test(x: proc (x, y: string): int) = discard

test(bar)   # Error: Type mismatch
```

A non-overloaded routine matches fine (`test(foo)` works), but an overloaded one is never resolved against the formal proc type. This blocks common higher-order patterns — e.g. passing `cmp` to `sort`.

## Fix

The compiler already resolves an overload choice against an expected **enum** type (`tryMatchEnumChoice` + an `hconv` resolved in `semConvArg`). This mirrors that for proc types:

- `tryMatchProcChoice` builds each candidate's structural proc type and runs the existing `procTypeMatch` against the formal, returning the **unique** match.
- In `sigmatch`'s routine-type branch, when the argument is an `Ochoice`/`Cchoice` and a unique overload matches, selection is deferred to `semConvArg` via an `hconv`, which emits the chosen symbol.

If zero or more than one candidate matches, the argument is left unresolved and the normal mismatch error fires — so the change is conservative (it only ever turns a previous error into a correct resolution).

## Tests

`tests/nimony/overload/toverloadedprocarg.nim` selects the `int` and `string` overloads of the same routine purely by the parameter's proc type, and runs to verify the *right* overload is chosen (not just that it compiles).

Verified no regressions across the proc/overload-related suites (`overload`, `calls`, `closures`, `lookups`, `templates`, `converter`, `macros`, `methods`, `iter`, `generics`, `sysbasics`, `typeconversions`).


src/nimony/sigmatch.nim
```diff
@@ -1174,6 +1174,45 @@ proc tryMatchEnumChoice*(choice: Cursor; enumTypeSym: SymId): SymId =
...
+      if procTypeOfRoutineSym(a.symId, buf):
+        var ac = beginRead(buf)
+        if ac.typeKind in RoutineTypes:
+          var trial = createMatch(context)
+          var fc = f
+          procTypeMatch(trial, fc, ac)
+          if not trial.err:
+            result = a.symId
+            inc matchCount
+    inc a
```
Araq:
```suggestion
    skip a
```

src/nimony/sigmatch.nim
```diff
@@ -1174,6 +1174,45 @@ proc tryMatchEnumChoice*(choice: Cursor; enumTypeSym: SymId): SymId =
...
+  if res.status != LacksNothing: return false
+  if res.decl.symKind notin RoutineKinds: return false
+  let r = asRoutine(res.decl)
+  buf.addParLe ProctypeT
+  buf.addDotToken() # nilability tag
+  buf.addSubtree r.params
+  buf.addSubtree r.retType
+  buf.addSubtree r.pragmas
+  buf.addParRi()
+  result = true
```
Araq:
```suggestion
  result = false
  if res.status == LacksNothing and res.decl.symKind in RoutineKinds:
    let r = asRoutine(res.decl)
    buf.addParLe ProctypeT
    buf.addDotToken() # nilability tag
    buf.addSubtree r.params
    buf.addSubtree r.retType
    buf.addSubtree r.pragmas
    buf.addParRi()
    result = true
```

PR #2060, opened 2026-06-30
std/complex: add complex numbers for 64-bit floats

savannt:
Adds `lib/std/complex` — complex numbers for 64-bit floats.

### What's included
- `Complex64` type (`re`, `im`) and a `Complex` alias for it
- Construction: `complex(re, im)`, `complex(re)`
- `==`, the four arithmetic operators for complex/complex **and** complex/scalar mixes, unary `-`
- `abs`, `abs2`, `arg`, `conjugate`, `inv`
- Transcendental: `sqrt`, `exp`, `ln`, `pow`
- `$`

### Design note
The module is **concrete over `float` (64-bit)** rather than generic over `SomeFloat` (as Nim 2's `std/complex` is). The stdlib's float `math`, comparison and `$` operations are defined per concrete float type and don't currently resolve through an abstract `SomeFloat` type parameter, so a generic `Complex[T]` body fails to typecheck (every `hypot`/`cos`/`==`/`$` call reports a constraint mismatch). Keeping it concrete matches how `lib/std/math` itself is written (concrete `float32`/`float64` overloads, no `SomeFloat` generics). If/when generic float resolution lands, this can be widened to `Complex[T: SomeFloat]` without breaking the API.

Results are built with object construction (`Complex64(re: ..., im: ...)`) so the init analysis sees `result` assigned on every path.

### Tests
`tests/nimony/stdlib/tcomplex.nim` — construction, arithmetic (incl. scalar mixes), `abs`/`abs2`/`arg`/`conjugate`/`inv`, `sqrt` (incl. `sqrt(-1) == i`), `exp`/`ln` round-trip, `pow`, and `$`. Passes `hastur test`.

### DIFF
--- a/lib/std/complex.nim
+++ b/lib/std/complex.nim
@@ -14,124 +14,136 @@
 ## part and `b` is the imaginary part (and `i` is the imaginary unit with the
 ## property `i*i == -1`).
 ##
-## The operations are provided for 64-bit floats (`Complex64`). A generic
-## `Complex[T: SomeFloat]` form (as in Nim 2) is not used here because the
-## stdlib's float `math`, comparison and `$` operations are concrete per float
-## type rather than generic over `SomeFloat`.
+## `Complex[T]` is generic over the floating-point component type `T`. Following
+## the modelling of `std/math`, each operation depends precisely on the concepts
+## it needs: the arithmetic operators come from `Arithmetic`, and the
+## transcendental functions add `HasSqrt`, `HasExp`, `HasLn`, `HasSin`,
+## `HasCos` and `HasArctan2`. So `Complex[float32]` and `Complex[float64]` both
+## work, and a user float type satisfying the relevant concepts works too.
 
 import std/math
 
 type
-  Complex64* = object
+  Complex*[T: SomeFloat] = object
+    ## A complex number with real part `re` and imaginary part `im`, both of
+    ## the floating-point type `T`.
+    re*, im*: T
+
+  Complex64* = Complex[float64]
     ## A complex number with 64-bit float components.
-    re*, im*: float
 
-  Complex* = Complex64
-    ## The default complex number type.
+  Complex32* = Complex[float32]
+    ## A complex number with 32-bit float components.
 
-func complex*(re, im: float): Complex64 =
+  Stringable = concept
+    proc `$`(x: Self): string
+
+func complex*[T: SomeFloat](re, im: T): Complex[T] =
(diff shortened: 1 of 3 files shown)


lib/std/complex.nim
```diff
@@ -0,0 +1,137 @@
...
+## on them.
+##
+## A complex number is a number of the form `a + b*i`, where `a` is the real
+## part and `b` is the imaginary part (and `i` is the imaginary unit with the
+## property `i*i == -1`).
+##
+## The operations are provided for 64-bit floats (`Complex64`). A generic
+## `Complex[T: SomeFloat]` form (as in Nim 2) is not used here because the
+## stdlib's float `math`, comparison and `$` operations are concrete per float
+## type rather than generic over `SomeFloat`.
```
Araq: Not good enough, it shall be `Complex[T]` and we need to be precise about `SomeFloat` which is likely a concept that requires the math operators. Then e.g. the `ln`, `sqrt` for Complex depends on `T: SomeNumber and HasLn` (`HasSqrt`). That is the clean modelling and Nimony's concepts are ready for it.

savannt:
Reworked to `Complex[T]` with concept-based modelling as suggested.

**`std/math`**: exported the existing `Arithmetic` concept and added per-function concepts `HasSqrt`, `HasExp`, `HasLn`, `HasSin`, `HasCos`, `HasHypot`, `HasArctan2` next to the functions they describe (same spirit as `HasDefault`). Generic code can now depend on exactly the operations it uses.

**`std/complex`**: `Complex*[T: SomeFloat]` (with `Complex64 = Complex[float64]` and `Complex32 = Complex[float32]`). Each op constrains `T` precisely:
- arithmetic/comparison ops: `SomeFloat and Arithmetic`
- `abs`: `SomeFloat and HasHypot`; `arg`: `SomeFloat and HasArctan2`
- `sqrt`: `... and HasHypot and HasSqrt`
- `exp`: `... and HasExp and HasSin and HasCos`
- `ln`: `SomeFloat and HasHypot and HasLn and HasArctan2`
- `pow`: union of `exp`'s and `ln`'s requirements

Generic literals use `T(0)`/`T(1)`/`T(0.5)`. The test now exercises both `float64` and `float32`.

Two judgment calls, happy to change: I kept the bound as `SomeFloat` rather than `SomeNumber` (matches Nim 2 and rules out `Complex[int]`), and reused the existing `Arithmetic` concept for the base operators instead of introducing a new one.

PR #2061, opened 2026-06-30
nimsem: warn on out-of-order object-constructor fields and named args (#1056)

savannt:
## What

Replaces the earlier source-order-preserving reordering of object constructors
with a compile-time **warning**, and applies the same warning to named call
arguments.

## Why

Per @Araq's review of the previous revision:

> It is lots of code and inconsistent with `someCall(argB = f(), argA = g())`
> which also does the reordering rewrite. We should flag both at compile-time
> ("reorder fields/named arguments to match evaluation order").

So both cases are now flagged instead of special-cased. The object constructor
keeps emitting fields in declaration order (values evaluate in that order); when
a flat object's fields — or a call's named arguments — are written in an order
that won't match their evaluation order, a warning advises reordering.

## Changes

- `semObjConstr`: track the source order of given fields; for a **flat** object
  (concrete, no base type, no variant part) whose fields are written out of
  declaration order, emit a warning. Inherited / variant / generic objects are
  conservatively not flagged (no false positives).
- `orderArgs` / `semcall`: detect when named arguments fill parameters out of
  source order on the **chosen** overload (speculative candidates are never
  reported) and warn. Template/macro calls are skipped.
- `SemContext.warn`: sem had no warning channel, so add a small one — printed
  via `Reporter` (exactly as `reportErrors` does for errors), deduped by source
  position, never written into the tree.
- Replace `tests/nimony/object/tobjconstr_evalorder` with a `.msgs` test that
  asserts both warnings fire and that the constructed values are still correct.

Net: **+96 lines** of compiler code (down from +143), no reordering machinery.

## Testing

`hastur test` green across the affected areas: object, overload, calls,
generics, templates, macros, const, consteval, converter, types, misc, methods,
basics, stdlib, exceptions, sets, strings, iter, casestmt — no failures and no
stray warnings.


Araq:
Not sure it's a good idea. It is lots of code and inconsistent with `someCall(argB = f(), argA = g())` which also does the reordering rewrite. We should flag both at compile-time ("reorder fields/named arguments to match evaluation order").

savannt:
Reworked along the lines you suggested. Instead of preserving source order in object constructors (the "lots of code" reordering), I now **flag both cases at compile time**:

- out-of-order **object-constructor fields** (flat objects), and
- out-of-order **named call arguments** (on the chosen overload),

each with a warning to reorder them to match evaluation order. The constructor keeps the existing declaration-order evaluation; nothing is silently reordered to preserve source order anymore. Inherited/variant/generic objects and template/macro calls are conservatively left unflagged to avoid false positives.

This needed a small `SemContext.warn` (sem had no warning channel) — printed through `Reporter`, deduped by source position, never written into the tree. Net is +96 lines of compiler code (down from +143), and `hastur test` is green across object/overload/calls/generics/templates/macros/const/stdlib/etc. with no stray warnings. Force-pushed.

src/nimony/sem.nim
```diff
@@ -4204,6 +4240,26 @@ proc semObjConstr(c: var SemContext; dest: var TokenBuf, it: var Item) =
...
   for field, pos in setFieldPositions:
     setFields[field] = cursorAt(fieldBuf, pos)
+
+  # `buildDefaultObjConstr` emits the fields in *declaration* order, so a value's
+  # side effects run in that order regardless of how the user wrote them. Rather
+  # than silently preserve source order (which would be inconsistent with how
+  # named call arguments are already reordered), flag the mismatch: warn when a
+  # flat object's fields are written in an order that won't match their
+  # evaluation order (nim-lang/nimony#1056).
+  if not isGenericObj and givenOrder.len >= 2:
```
Araq: Do the check in the main logic, not as another separate code path.

src/nimony/sigmatch.nim
```diff
@@ -2275,6 +2280,23 @@ proc orderArgs*(m: var Match; paramsCursor: Cursor; args: openArray[CallArg]): s
       inVarargs = true
     inc ai
 
+  # Detect whether named arguments changed the evaluation order: walking the
+  # parameters in declaration order, the source-argument index that fills each
+  # given (non-default, non-vararg-continuation) parameter must be increasing.
+  # If it isn't, a later parameter is filled by an earlier-written argument, so
+  # the values run out of source order — flagged by the caller (consistent with
+  # object constructors, nim-lang/nimony#1056).
+  block:
```
Araq: See if you can do this check in the main logic, not as a separate code path.

savannt:
Folded both checks into the existing traversals, no separate pass:

- **Object constructors:** `findObjFieldAux` now records each field's declaration index (`ObjField.pos`) as part of the lookup `semObjConstr` already does per field, so `semObjConstr` just tracks that the indices strictly increase inline — the separate `flatFieldDeclOrder` re-walk of the type is gone. As a bonus this now also covers variant objects correctly (only inherited objects are skipped, since their emission order interleaves base/derived fields and a per-level `pos` isn't a comparable rank there).
- **Named arguments:** the reorder detection now happens in `orderArgs`' main loop that walks the parameters to assemble the argument list, instead of a dedicated follow-up loop.

Rebuilt and reverified: object/overload/calls/generics/templates/macros/plugins/errmsgs all green with `--jobs:1`, and the runtime test still pins that out-of-order fields/args evaluate in declaration/parameter order. Force-pushed.

savannt:

Issue #1987, opened 2026-06-22
lengc/LLVM: template should appear as a function call during debugging


bridgeQiao:
at my working branch, debugging a `template` jumps directly to the source code:

```
(lldb) 
Process 3292 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = step over
    frame #0: 0x000000010001b468 seqs_and_strings`X60Qini_0_seqqcvqa1 at seqs_and_strings.nim:99:23
   96  	# endStore syncs it back. Forgetting it means s[0..6] returns
   97  	# wrong data after a write.
   98  	
-> 99  	var target = newString(10)
   100 	let p = beginStore(target, 5)   # get mutable pointer, ensure unique
   101 	p[0] = 'H'
   102 	p[1] = 'e'
Target 0: (seqs_and_strings) stopped.
(lldb) 
Process 3292 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = step over
    frame #0: 0x000000010001b47c seqs_and_strings`X60Qini_0_seqqcvqa1 at seqs_and_strings.nim:100:19
   97  	# wrong data after a write.
   98  	
   99  	var target = newString(10)
-> 100 	let p = beginStore(target, 5)   # get mutable pointer, ensure unique
   101 	p[0] = 'H'
   102 	p[1] = 'e'
   103 	p[2] = 'l'
Target 0: (seqs_and_strings) stopped.
(lldb) 
Process 3292 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = step over
    frame #0: 0x000000010001b498 seqs_and_strings`X60Qini_0_seqqcvqa1 at system.nim:34:10
   31  	template `[]=`*(x: cstring, i: int; elem: char) =
   32  	  (x[i]) = elem
   33  	template `[]=`*[T](x: ptr UncheckedArray[T], i: int; elem: T) =
-> 34  	  (x[i]) = elem
   35  	template `[]=`*[T](x: UncheckedArray[T], i: int; elem: T) =
   36  	  (x[i]) = elem
   37  	
Target 0: (seqs_and_strings) stopped.
```

This happens because the `template` lacks a flag or marker.

Araq:
No idea why you closed it, but yeah it should. It's just that it is/was hard to do.

bridgeQiao:
I realized the title wasn't quite appropriate, but I couldn't think of a more fitting one, so I closed it :)

[...]

PR #2067, opened 2026-06-30
lengc/LLVM: support basic debug functionality

bridgeQiao:
- Generate basic DI type information, support line-by-line debugging with printing of global and local variables
  - Add `llvmirmodel` to separate LLVM IR metadata from .ll string generation
  - Formatted with `nimpretty`, resulting in a large diff


bridgeQiao:
Two known bug, but not in this PR range: #1987 #2068

src/lengc/llvmserializer.nim
```diff
@@ -0,0 +1,278 @@
...
+import llvmirmodel
+
+const OrderingStr*: array[LLAtomicOrdering, string] = [
+  "unordered", "monotonic", "acquire", "release", "acq_rel", "seq_cst"]
+
+const RmwOpStr*: array[LLAtomicrmwOp, string] = [
+  "xchg", "add", "sub", "and", "or", "xor",
+  "nand", "min", "max", "umin", "umax"]
+
+proc serialize*(typ: LLType): string =
```
Araq:
```suggestion
proc serialize*(typ: LLType; result: var string) =
```

And for all helper procs. Avoids Exponential blow-ups.

Issue #1974, opened 2026-06-18
Phased semchecking limitation when resolving `when` statements


gmpreussner:
```nim
# Regression: top-level `let` must be visible when `when` conditions are
# folded in phase 2. Classic Nim resolves the `let` binding and folds
# `when a is SomeInteger` to false without typechecking the dead branch.

# Must work with sum types

type
  Foo = distinct float64

let a = 1.0.Foo

when a is SomeInteger: # Error: undeclared identifier: a
  discard $a

when typeof(a) is SomeInteger:
  discard $a # OK

# Must also work with concepts

type
  Stringable = concept
    proc `$`(x: Self): string

when a is Stringable: # Error: undeclared identifier: a
  discard $a

when typeof(a) is Stringable:
  discard $a # Error: Type mismatch at [position]
```

In classic Nim, the compiler resolves `a`’s type (`Foo`), folds the `is` check at compile time (`false`), and never typechecks `$a`.

In Nimony, top-level `let` and `when` are handled in different phases:
- Top-level `let` is only semchecked in phase 3 (`SemcheckBodies`) via `toplevelGuard`
- `when` conditions are evaluated in phase 2 (`SemcheckSignatures`) via `semWhen` / `semConstBoolExpr`

So when Nimony evaluates `when a is XYZ`, `let a` has not been registered yet, hence `undeclared identifier: a`. In the `Stringable` concept case this is also unrelated to whether `Foo` has `$`.

gmpreussner:
@Araq I have a fairly small fix for this that mirrors how `const` folding (correctly) works in this case. However, the concept part of this test depends on https://github.com/nim-lang/nimony/pull/1976, which I would like to get merged first.

gmpreussner:
After investigating some more, this turned out to be more complex than expected. There are two separate issues:
1. phase ordering for top-level `let` + `when`
2. dead-branch typechecking when a concept `is` check is (incorrectly) kept alive.

Top-level `let` can be addressed relatively easily by mirroring `const` folding behavior in _sem.nim_, but that's just a very narrow and incomplete fix. For example, if the `let` statement references any `var`, the latter is still processed in phase 3 of semchecking and not yet in scope for `let`, so that case fails.

Araq suggested that a proper fix for these and other related issues involves a refactor of the semcheck phases so that they can run on demand instead of in fixed order. That's a much more substantial undertaking though.

## Workarounds

For specific use cases the following workarounds are available:

[...]

Issue #2064, opened 2026-06-30
Heavy refactor / rewrite of nimsem


Araq:
- Make it use the nifcore APIs which allow for a much faster `skip` operation.
- Make the 3 phases over the input less rigid and give it an on-demand resolution flavor.

Araq:
> Which APIs? I've used the existing nifcursors structured helpers (into/loopInto/hasMore), since that's what nimsem is on today. Is that the right direction, or do you want these ported toward the nifcore APIs specifically?

Right direction. But we also need a `peekInto` template that is like `into` but does not enforce total traversal. Both the old nifcursors and the new nifcore implementation need this `peekInto`.

> Predicate-only swaps — progress or churn? Some of these only change while x.kind != ParRi → while x.hasMore while a manual skipParRi still follows. Useful as an intermediate step, or would you rather I land only the conversions that also remove the manual enter/exit (the into/loopInto ones)?

Predicates can use `peekInto` and are to be converted too.

> Core flat walkers — touch them or leave them to you? 

Need something like plugins.nim `template balancedTokens` but name it `linearScan`.

> The remaining unstructured loops are balanced walkers in core code (e.g. the hook-tag search in programs.nim, and walkers in contracts/derefs) whose semantics are subtler — some search all depths rather than direct children.

See above, these are linearScans.

> One PR or several? Keep this as a single growing PR, or split per area?

Several PRs, start a new one whenever I merged one.

> Is that the direction you want for the phase de-rigidification, and should I generalize it next

The direction is good, but this should be done in a separate PR, independent of the nifcore-refactoring. We will deal with potential merge conflicts as they arise.

PR #2070, opened 2026-06-30
nimsem: incremental structured cursor traversal (toward #2064)

savannt:
Incremental work toward #2064, per your steer to do it piecemeal ("every refactoring that does away with the unstructured `while`+`inc` loops helps"). Converts nimsem's manual `inc n` / `while n.kind != ParRi` / `inc n` cursor loops to the structured `into` / `loopInto` / `hasMore` helpers.

### Commits
- **idetools `tr`** — `TypeS` / `DotX` / `FldU` + `findAndAddFieldDefinition` → `into`/`loopInto`/`hasMore`.
- **nimsem structured loops** — the `while <cur>.kind != ParRi` loops across `sempragmas`/`semtypes`/`semdecls`/`semmagics`/`macro_plugin`/`nimony_model`/`renderer` → `hasMore`, plus two enter-and-scan loops → `loopInto`.
- **idetools fixed-slot handlers** — the `inc # ParLe` … `inc # ParRi` blocks (ProcS family, VarS/LetS/ConstS, ParamU) → `into`.
- **exprexec `collectSyms`** — `while true` + `nested`-counting balanced walk → recursive `loopInto`.

Behaviour-preserving; the full `tests/nimony` suite stays green after each commit (only the pre-existing `concepts` `concept of` parse failure remains).

### Open questions before I go further — would really appreciate your steer
1. **Which APIs?** I've used the existing `nifcursors` structured helpers (`into`/`loopInto`/`hasMore`), since that's what nimsem is on today. Is that the right direction, or do you want these ported toward the `nifcore` APIs specifically?
2. **Predicate-only swaps — progress or churn?** Some of these only change `while x.kind != ParRi` → `while x.hasMore` while a manual `skipParRi` still follows. Useful as an intermediate step, or would you rather I land only the conversions that also remove the manual enter/exit (the `into`/`loopInto` ones)?
3. **Core flat walkers — touch them or leave them to you?** The remaining unstructured loops are balanced walkers in core code (e.g. the hook-tag search in `programs.nim`, and walkers in `contracts`/`derefs`) whose semantics are subtler — some search *all depths* rather than direct children. Happy to convert them, but I'd want to confirm the intended semantics first; or I can leave the perf/semantically-sensitive ones to you and stick to the mechanical cases.
4. **One PR or several?** Keep this as a single growing PR, or split per area?


src/nimony/exprexec.nim
```diff
@@ -61,21 +61,16 @@ type
...
-  if n.kind != ParLe:
-    # atom:
-    if n.kind == Symbol: stack.add n.symId
+  case n.kind
+  of Symbol:
+    stack.add n.symId
+  of ParLe:
+    var child = n
+    child.loopInto():
+      collectSyms(child, stack)
```
blackmius: previously it wasnt recursive, i dont think tail call optimization works here

src/nimony/idetools.nim
```diff
@@ -63,8 +63,7 @@ type
     usages: seq[Usage]
 
 proc findAndAddFieldDefinition(c: var IdeContext, n: var Cursor, name: string, nameSym: SymId, containingType: Cursor) =
-  inc n
-  while n.kind != ParRi:
+  n.loopInto():
```
Araq:
```suggestion
  n.loopInto:
```

And everywhere else

src/nimony/exprexec.nim
Araq: Irrelevant, nifcore produces more recursions, true.

savannt:
Landed a first on-demand-across-phases step on this branch (b5699d23, fixes #1974): a signature-phase `when` that inspects a toplevel `let`/`var`'s type now drives just that decl's signature on demand instead of failing with `undeclared identifier`. The decl is still deferred to the body phase; only when a `when` condition actually references it does its signature get resolved (into a throwaway buffer, value carried verbatim so there's no macro expansion or forward-ref hazard), and the body phase reuses the same symbol. Modules that don't use the pattern stay byte-identical.

It's a point solution, though — the general shape would be activating `loadSymWithPhase` / `ensurePhase` (currently inert) as a real "drive symbol to phase N" driver, and routing this + `tryPromoteTemplateBody` + `semExprMissingPhases` through the one hook.

Is that the direction you want for the phase de-rigidification, and should I generalize it next — or focus on the nifcore fast-skip first?

Araq:
> Which APIs? I've used the existing nifcursors structured helpers (into/loopInto/hasMore), since that's what nimsem is on today. Is that the right direction, or do you want these ported toward the nifcore APIs specifically?

Right direction. But we also need a `peekInto` template that is like `into` but does not enforce total traversal. Both the old nifcursors and the new nifcore implementation need this `peekInto`.

> Predicate-only swaps — progress or churn? Some of these only change while x.kind != ParRi → while x.hasMore while a manual skipParRi still follows. Useful as an intermediate step, or would you rather I land only the conversions that also remove the manual enter/exit (the into/loopInto ones)?

Predicates can use `peekInto` and are to be converted too.

> Core flat walkers — touch them or leave them to you? 

Need something like plugins.nim `template balancedTokens` but name it `linearScan`.

> The remaining unstructured loops are balanced walkers in core code (e.g. the hook-tag search in programs.nim, and walkers in contracts/derefs) whose semantics are subtler — some search all depths rather than direct children.

See above, these are linearScans.

> One PR or several? Keep this as a single growing PR, or split per area?

Several PRs, start a new one whenever I merged one.

> Is that the direction you want for the phase de-rigidification, and should I generalize it next

The direction is good, but this should be done in a separate PR, independent of the nifcore-refactoring. We will deal with potential merge conflicts as they arise.

src/nimony/decls.nim
```diff
@@ -139,6 +139,22 @@ proc asLocal*(c: Cursor; mode = SkipInclBody): Local =
 proc asTypevar*(c: Cursor): Local {.inline.} =
   result = asLocal(c)
 
+proc localHasAtomExplicitType*(n: Cursor): bool =
```
Araq: What about `var x: T`? That one surely is a generic and `T` is an atom! The whole idea seems bonkers to me...

savannt:
Rebased this down to the nifcore / structured-traversal refactoring only. The on-demand phase-resolution work (#1974) that had accumulated on the tip is now split into its own PR (#2080), independent of this one, as you asked. This branch is back to the six structured-loop commits.

Next here, per your review: a `peekInto` template (like `into` but without enforced total traversal), converting the predicate loops to it, and a `linearScan` (modeled on plugins.nim `balancedTokens`) for the all-depth balanced walkers (the programs.nim hook search, contracts/derefs).

savannt:
Added the two pieces from your review:

**`peekInto`** (nifcursors + nifcore) — like `into` but does not enforce total traversal; it enters, runs the body, then skips whatever's left and advances past the `)` via a single `skip` from the rewound head (O(1) under the bounded-`rem` nifcore rep). Adopted at the two clean partial-consumption sites: the renderer cstring/pointer nilness rendering, and `implicitlyDiscardable`'s pragma scan (drops its manual "mop-up before early-exit").

On the other predicate loops from the earlier commit: most aren't `peekInto` shape — they're copy-walks (`takeTree` into a dest, so every token must be visited) or early-return *error* paths where the manual `skipParRi` precedes a proc `return` (wrapping them in `peekInto` would skip its epilogue through the return and leave the cursor mid-scope). I left those as `hasMore`. Happy to restructure them into whole-proc `peekInto` (moving the success logic into an `else`) if that's the shape you want.

**`linearScan`** — deep all-depth `ParLe` visitor (the counterpart to `loopInto`); converted the three hand-rolled balanced walkers: `tryLoadHook` / `tryLoadAllHooks` (type-hook pragma lookups) and the sembasics `(err)` search.

Two questions:
1. `linearScan` currently shares its shape with the existing `nifcursors.balancedTokens`. I kept it a separate name (as you asked) and didn't touch the validator, but happy to unify — alias, or rename `balancedTokens` → `linearScan` and migrate its validator call-sites.
2. Should `linearScan` also get a `nifcore` (NifCursor) version, matching how `peekInto` landed in both?

Full tests/nimony green (536/536 outside the pre-existing nifler `concepts` parse failure); validator stays clean on the `sem.nim` passFile.

Araq:
1. Name it `linearScan` in nifcore too and patch its use-sites. It is the better name.
2. Yes.

src/lib/nifcursors.nim
```diff
@@ -343,6 +343,31 @@ template into*(n: var Cursor; body: untyped) =
...
+    let j = jump(n.load)
+    let isOverflow = j == MaxJump
+    n.p = cast[ptr PackedToken](cast[uint](n.p) + sizeof(PackedToken).uint)
+    n.rem = if isOverflow: high(int) else: int(j)
+    body
+  else:
+    inc n          # skip opening tag
+    body
+  n = savedParLe   # rewind to the opening tag …
+  skip n           # … and skip the entire subtree past the closing `)`
```
Araq: That is terrible performance for the default, not-virtualParRi mode.

src/nimony/exprexec.nim
```diff
@@ -66,16 +66,9 @@ proc collectSyms(n: Cursor; stack: var seq[SymId]) =
...
-      of ParRi:
-        dec nested
-        if nested == 0: break
-      of ParLe: inc nested
-      of Symbol: stack.add n.symId
-      else: discard
-      inc n
+    n.loopInto:
+      collectSyms(n, stack)
+      skip n
```
Araq: Use a helper collectSyms that takes a `n: var Cursor` here so that this skip can be avoided.

src/nimony/typeprops.nim
```diff
@@ -218,21 +218,15 @@ proc lengthOrd*(c: var SemContext; typ: TypeCursor): xint =
...
-    inc n
+  if n.kind == Symbol:
+    let res = tryLoadSym(n.symId)
+    if res.status == LacksNothing and res.decl.tagEnum == TypevarTagId:
+      return true
+  elif n.kind == ParLe:
+    var n = n
+    n.loopInto:
+      if containsGenericParams(n): return true
+      skip n
```
Araq: Likewise, use `n: var TypeCursor` to avoid the skip operation.

savannt:
Addressed all three:

**`peekInto` finish** — reworked: the default (not-virtualParRi) mode no longer rewinds and re-skips the subtree; it mops up *forward* from wherever the body stopped (`while n.kind != ParRi: skip n`), so only the unconsumed remainder is walked and nothing is traversed twice. The virtualParRi mode keeps its O(1) rewind+jump finish. The body contract now matches `into` (stop at a child boundary); both current use sites already comply.

**`collectSyms`** — now takes `n: var Cursor`, so the recursion itself advances the cursor and the per-child `skip` is gone (the two outer call sites pass a `var` copy).

**`containsGenericParams`** — same treatment via a `var TypeCursor` helper; the public by-value signature stays (≈20 call sites across sem/sigmatch/semtypes/semcall unchanged). On a `true` early-out the helper abandons the cursor mid-subtree, matching the previous early-return behaviour.

Full tests/nimony green (the only failures are the 14 pre-existing `concepts/` nifler parse errors — those files are byte-identical to master, it's my local host-Nim being older than CI's devel); `nim check -d:virtualParRi` clean.

PR #2073, opened 2026-06-30
nimsem: drop dead `[position]` placeholder from type-mismatch error

savannt:
## Problem

The overload-resolution mismatch error was built with a literal `[position]` placeholder that is **never substituted**:

```nim
errorMsg = "Type mismatch at [position]\n"
```

`[position]` appears nowhere else in `src/` or `lib/` — nothing ever fills it in, so it prints verbatim in every type-mismatch diagnostic (one of the most common compiler errors):

```
foo.nim(11, 12) Error: Type mismatch at [position]
result +=? x
[2] expected: float64 but got: T (declared in foo.nim(6, 1))
```

The argument position is already reported per candidate as the `[N]` prefix (from `m.error.pos` via `addErrorMsg`), and the source location is already in the `Error:` line — so `at [position]` was pure dead noise.

## Fix

Drop the placeholder, and lowercase to `type mismatch` for consistency with the sibling messages produced by the same proc (`ambiguous call`, `undeclared field`, `undeclared identifier`):

```
foo.nim(11, 12) Error: type mismatch
result +=? x
[2] expected: float64 but got: T (declared in foo.nim(6, 1))
```

Affected `.msgs` goldens regenerated (header line only).

## Verification

`errmsgs` 29/29, `generics` 25/25, `calls` 10/10, `overload` 9/9 — all green.

### DIFF
--- a/src/nimony/semcall.nim
+++ b/src/nimony/semcall.nim
@@ -1032,7 +1032,7 @@ proc resolveOverloads(c: var SemContext; dest: var TokenBuf; it: var Item; cs: v
         errorMsg.add " for type "
         errorMsg.add typeToString(cs.args[0].typ)
     elif m.len > 0:
-      errorMsg = "Type mismatch at [position]\n"
+      errorMsg = "type mismatch\n"
       errorMsg.add asNimCode erroredN
       for i in 0..<m.len:
         errorMsg.add "\n"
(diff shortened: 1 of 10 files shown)


Araq:
Read again the original error message.

```
foo.nim(11, 12) Error: Type mismatch at [position]
result +=? x
[2] expected: float64 but got: T (declared in foo.nim(6, 1))
```

It explains to you that the [2] is the position.

PR #2075, opened 2026-07-01
sem: compile-time range subtype checking

savannt:
## What

Nimony currently treats `range[0..10]` exactly like its base type `int` with **no checking at all** (documented at `doc/differences.md`). This PR adds compile-time enforcement for statically-known values.

| Case | Before | After |
|---|---|---|
| `var x: range[0..10] = 20` | silently compiled | `Error: value out of range: 20 notin 0..10` |
| `(range[0..10])(50)` | silently compiled | rejected with the same message |
| `range[2..5]` → `range[0..10]` | **rejected** (exact-tree match) | accepted (proper subtyping) |
| `range[0..20]` → `range[0..10]` | silently compiled | `value out of range: 0..20 notin 0..10` |
| in-range values | ok | ok |

Covers `var`/`let`/`const` initialization and explicit conversions.

## How

Confined to semantic analysis — **no ABI or codegen changes**:

- `sigmatch.nim`: a new `ValueOutOfRange` match-error kind with a precise diagnostic, reusing the existing `firstOrd`/`lastOrd`/`checkIntLitRange` helpers (which already understand `rangetype`). The `of RangetypeT:` case in `singleArgImpl` now does real subset checking instead of exact-tree `linearMatch`, and rejects out-of-range integer literals.
- `sem.nim`: routes the relevant type-mismatch reports to surface the range diagnostic (taking care that the object-downcast fallback probe in `semConvArg` doesn't swallow it).

Degenerate/synthetic ranges (e.g. the empty `range[0.. -1]`) and non-statically-known bounds are guarded and left untouched.

## Tests

- `tests/nimony/errmsgs/trangecheck.{nim,msgs}` — the five rejection cases.
- extended `tests/nimony/sysbasics/tranges.nim` — subtyping + boundary values compile and run.

Regression-clean locally: sysbasics 61/61, errmsgs 28/28, generics 25/25, enums 10/10, const 8/8, rtchecks 1/1, and overload/calls/object/converter/iter/opt/misc/casestmt/arc all pass.

## Scoped out (follow-ups)

- **Runtime** bounds enforcement for values only known at run time (`var x: range[0..10] = someInt`). Needs a new compilerproc + hexer lowering and a representation decision (overload `hconv` vs. a new tag; panic vs. `RangeDefect`) — happy to do it in a follow-up once the direction is agreed.
- **Float ranges** are separately broken (bounds are stored `IntLit`-only, so `range[0.0..1.0]` can't even be constructed). Distinct fix.

Updates `doc/differences.md` accordingly.

### DIFF
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -1704,15 +1704,25 @@ proc singleArgImpl(m: var Match; f: var Cursor; arg: CallArg) =
         # handled in linearMatch
         linearMatch m, f, a
     of RangetypeT:
-      # for now acts the same as base type
+      # A `range[lo..hi]` matches structurally on its *base type* only; whether
+      # a value actually fits the bounds (and whether one range is a subset of
+      # another) is a proof obligation discharged later by the contracts engine
+      # (see `checkRangeAssign` in contracts_njvl.nim), not a type-match failure
+      # here. This deliberately accepts legal narrowings such as
+      # `range[2..5]` -> `range[0..10]` that an exact-tree match would reject.
       var a = skipModifier(arg.typ)
       if a.typeKind == RangetypeT:
-        linearMatch m, f, a
+        var fb = f
+        var ab = a
+        inc fb # -> formal base type
+        inc ab # -> arg base type
+        linearMatch m, fb, ab # base types must be compatible
+        skip f # consume the whole formal range type
       else:
         inc f # skip to base type
         linearMatch m, f, a
-        skip f
-        skip f
+        skip f # lo bound
+        skip f # hi bound
         expectParRi m, f
     of ArrayT:
       var a = skipModifier(arg.typ)
--- a/doc/differences.md
+++ b/doc/differences.md
@@ -53,7 +53,14 @@ The first pre-release of Nimony misses lots of things but might be useful for yo
 Nimony currently lacks these features:
 
 - An exception handling system that is compatible with Nim's. Instead Nimony uses its own based on an `ErrorCode` enum.
-- Nimony does not check `range` subtypes and treats a `range` type like its underlying base type. For example `range[0..10]` is treated as `int`.
+- Nimony checks `range` subtypes at compile time by delegating to its
+  contracts engine: a value bound to a `range[lo..hi]` carries the proof
+  obligation `lo <= value <= hi`, and whatever the engine cannot prove in range
+  is rejected — no runtime check is ever emitted, so there are no new dynamic
+  failure modes. A narrower range is a proper subtype of a wider one
+  (`range[2..5]` fits `range[0..10]`), an out-of-range literal is rejected
+  (`var x: range[0..10] = 20`), and a value proven in range by a preceding guard
+  (`if a in 0..10: (var x: range[0..10] = a)`) is accepted.
 - Nim's effect system:
   - `tags` annotations are ignored. Tag mismatches should produce warnings in Nim 3 with the option to turn these warnings into errors.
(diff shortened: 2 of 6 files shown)


src/nimony/sem.nim
```diff
@@ -765,7 +769,10 @@ proc semConvArg(c: var SemContext; dest: var TokenBuf; destType: Cursor; arg: It
     var m = createMatch(addr c)
     typematch m, destBase, matchArg
     if m.err:
-      c.typeMismatch dest, info, arg.typ, destType
+      if m.isValueOutOfRange:
+        buildErr c, dest, info, getErrorMsg(m)
+      else:
+        c.typeMismatch dest, info, arg.typ, destType
```
Araq: Copy&pasted 3 times already, make a helper proc.

src/nimony/sigmatch.nim
```diff
@@ -48,10 +48,12 @@ type
     TooFewArguments
     NameNotFound
     ParamAlreadyGiven
+    ValueOutOfRange
 
   MatchError* = object
     info: PackedLineInfo
-    #msg: string
+    msg: string ## pre-rendered message for errors (e.g. ValueOutOfRange) whose
+                ## text is cheaper to build at the error site than in getErrorMsg
```
Araq: I don't care about "cheaper to build", I care about "lazy on demand to build", because that matters in overload resolution where errors typically mean "different proc gets selected".

Araq:
We can do much better and delegate the task to contracts+inferle -- what the engine cannot prove, we reject. Zero runtime costs, no new dynamic failure modes!

savannt:
Done — reworked to delegate range checking to the contracts + inferle engine, exactly as you suggested. No runtime checks are ever emitted.

A value bound to a `range[lo..hi]` now carries the obligation `lo <= value <= hi`, discharged in `contracts_njvl.nim` at every binding site (local init, store, checked conversion):

- A range-typed value is proven in range **structurally** from its own declared type (`range[2..5]` ⊆ `range[0..10]`), which is robust across control-flow joins where flow-derived facts get intersected away.
- Otherwise the obligation goes to `inferle`, so a value proven in range by a preceding guard is accepted:
  ```nim
  proc f(a: int) =
    if a >= 0:
      if a <= 10:
        var x: range[0..10] = a   # accepted — proven by the guards
  ```
- What the engine cannot prove is rejected (`var x: range[0..10] = 20`, or narrowing a wider `range[0..20]` parameter to `range[0..10]`).

`sigmatch.nim` now only matches a `range` formal on its **base type** (so range types are never rejected outright and legal narrowings pass); the `ValueOutOfRange` match-error machinery is gone. `sem.nim` is reverted to master — which also resolves the "copy&pasted 3 times" comment, since that plumbing no longer exists.

Verified locally: errmsgs 32/32, sysbasics 61/61, borrows/notnil/rtchecks/typeconversions/generics/enums/const/overload/calls/object/converter/iter/opt/misc/casestmt/arc all green.

savannt:
Rebased onto current master, so the range logic now lives in the Final IR contracts engine (`contracts_fir.nim`, `analyzeContractsFinalIr`) rather than the old `contracts_njvl.nim` that #2093 removed. No behavioural change from my previous comment — the checks moved with the engine and everything still verifies (errmsgs incl. `trangecheck`, sysbasics, borrows/notnil, generics/enums/const/overload/calls/object/converter/iter/opt/misc/casestmt/arc/exceptions/mut/scopes/when all green). PR is a single commit and merges cleanly.

Araq:
In a follow-up PR, add more tests for this feature.

PR #2076, opened 2026-07-01
Implement the RangeCheck runtime mode for range conversions

savannt:
## What

`RangeCheck` is a declared, default-on check mode — it sits next to `BoundCheck` in `DefaultSettings` and is threaded all the way into `checkFlags` (`"br"`) — but **nothing ever consumed it**. As a result a conversion to an ordinal `range[a..b]` type silently truncated an out-of-range value at runtime instead of aborting:

```nim
proc get(v: int): int = v
echo range[0..10](get(20))   # before: prints 20;  after: aborts
```

This PR implements the runtime side, closely mirroring the existing array bound check.

## How

- **`lib/std/system/panics.nim`** — add `nimIRcheck` / `raiseRangeError3`, analogous to `nimIcheckAB` / `raiseIndexError3`. Unlike the bound check, the value keeps its magnitude (no `- a` rebasing), since a range value *is* its ordinal.
- **`src/hexer/desugar.nim`** — `trRangeConv` rewrites `(conv|hconv (rangetype base lo hi) x)` into `(conv … (call nimIRcheck x lo hi))` when the `RangeCheck` mode is active. This is done in `desugar` (before `xelim`), for the same reason the bound check is: the check call can then be hoisted into a temp and inlined, instead of staying buried in the conversion expression.
- **`src/nimony/nimony.nim`** — add a `--rangechecks:on|off` switch mirroring `--boundchecks`. `-d:danger` already clears all check modes, so it turns range checks off with the rest.

The error message matches the compile-time diagnostic and the bound-check style:

```
value out of range: 20 notin 0..10
```

## Scope

This covers **explicit** `range[a..b](x)` conversions, which sem already emits as `conv` nodes (with the range type preserved). It works for `int`, `char`, and negative/`char` ordinal ranges.

**Implicit** narrowings — `var x: range[0..10] = i`, parameter passing, assignment — currently store the value bare (the `# for now acts the same as base type` path in `sigmatch`), so there is no conversion node to check. Making sem emit a conversion there is the deferred "ranges act as their base type" work; I explored it and it has non-trivial ripple effects (it changes the type of `low`/`high` results), so it belongs in its own change. The good news is the desugar machinery here already matches **any** conv/hconv-to-`rangetype`, so implicit narrowings get checked for free the moment sem starts emitting those nodes.

## Tests

`tests/nimony/rtchecks/trangeconv` exercises both the in-range pass-through and the out-of-range abort. Full `hastur nimony` is green except for a pre-existing set of `concepts/*` tests that fail at the `nifler` parse stage (`concept … of` syntax), unrelated to this change.

### DIFF
--- a/lib/std/system/panics.nim
+++ b/lib/std/system/panics.nim
@@ -105,6 +105,27 @@ proc nimUcheckB(i, b: uint): uint {.inline, exportc: "nimUcheckB".} =
   if result > b:
     raiseIndexError3(i, 0, b)
 
+proc raiseRangeError3[T: HasWriteErr](i, a, b: T) {.noinline.} =
+  writeErr "value out of range: "
+  writeErr i
+  writeErr " notin "
+  writeErr a
+  writeErr ".."
+  writeErr b
+  writeErr "\n"
+  die 1'i32
+
+proc nimIRcheck(i, a, b: int): int {.inline, exportc: "nimIRcheck".} =
+  ## Range check for a conversion to an ordinal `range[a..b]`: returns `i`
+  ## unchanged when `a <= i <= b`, otherwise terminates. Unlike the array
+  ## bound check (`nimIcheckAB`), the value keeps its magnitude (no `- a`
+  ## rebasing), since a range value *is* its ordinal.
+  if i >= a and i <= b:
+    result = i
+  else:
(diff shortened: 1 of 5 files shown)


savannt:

PR #2078, opened 2026-07-01
make libc-free the default

Araq:


Araq:
> Happy to open this as a PR against nativenif if you'd like.

Of course!

savannt:
Opened it as nim-lang/nativenif#79 (branch `savannt:fix-native-plugin-mov-typecheck`). Same 3 edits, verified 161/161 on nativenif's suite + 23/24 here; `tparfor_race` left out as the separate arkham-codegen SIGBUS.

PR #2080, opened 2026-07-01
nimsem: on-demand signature resolution of toplevel let/var for `when` (#1974)

### LINKED ISSUE #1974: Phased semchecking limitation when resolving `when` statements

savannt:
Split out of #2070 per your request to keep the nifcore / structured-traversal refactoring independent of the phase de-rigidification work.

## What

Fixes #1974: a signature-phase `when` that inspects a toplevel `let`/`var`'s type currently fails with `undeclared identifier`, because toplevel locals get no signature until the body phase. This resolves *just that one decl's signature on demand*, only when a `when` condition actually references it.

- Phase 2 still defers a toplevel let/var to the body phase (output stays byte-identical for modules that don't use the pattern), but records `name -> declCursor` in `deferredLocals`.
- When a phase-2 `when` condition's `semIdentImpl` hits an undeclared name that is in `deferredLocals`, `resolveDeferredLocal` sems just that decl's signature into a throwaway buffer (value carried **verbatim**, so no macro expansion / forward-ref hazard), registers it in scope + `prog.mem`, and retries.
- The minted `SymId` is stashed in `onDemandResolved`; the body phase's `handleSymDef` reuses it, so the decl is neither redeclared nor shifts the global name counter.

Verified: modules without such a `when` are byte-identical (nosystem canary). Works for branch-uses-`a`, use-after-`when`, forward-ref values, multi-var conditions. Inferred / compound / template-injected explicit types degrade to a clean "undeclared" rather than crashing.

The last commit also activates the previously-inert `loadSymWithPhase` / `ensurePhase` keystone (relocated out of `semmain` so the on-demand consumers can reach it) and routes `tryPromoteTemplateBody` through it as the first "drive symbol to phase N" consumer — the general shape for unifying the three ad-hoc on-demand mechanisms.

## Open question — the eligibility heuristic

You flagged `localHasAtomExplicitType`: `var x: T` with `T` a generic param is an atom, so the atom-vs-compound *syntactic* test is the wrong discriminator. Agreed. The real question isn't "is the type an atom" but "can this decl's signature be resolved standalone in phase 2 with no body side effects?" I'd like your steer on the right predicate — e.g. drive resolution purely from the undeclared-id point (`loadSymWithPhase` in a mint-if-absent mode) and let a failed standalone resolution fall back to *defer*, instead of gating on syntax up front. That would drop the atom heuristic entirely.

Toward #2064 (bullet B: phase de-rigidification).


src/nimony/decls.nim
```diff
@@ -139,6 +139,22 @@ proc asLocal*(c: Cursor; mode = SkipInclBody): Local =
 proc asTypevar*(c: Cursor): Local {.inline.} =
   result = asLocal(c)
 
+proc localHasAtomExplicitType*(n: Cursor): bool =
+  ## Peek whether a `(let|var ...)` decl carries an explicit type annotation
+  ## that is a single atom (a bare Ident/Symbol — e.g. `int`, `string`, a
+  ## non-generic nominal or alias), as opposed to a `DotToken` (inferred) or
+  ## a compound type expression (a `ParLe`, e.g. a generic instance `Foo[T]`
+  ## or `seq[int]`). Atom types resolve without triggering generic
+  ## instantiation, so they are safe to resolve in the signature phase;
```
Araq: A single generic `T` is also an atom (Symbol) and needs to be instantiated. This heuristic is bonkers, use some other solution. One based on the discussed `linearScan` primitive seems to fit.

savannt:
Agreed the atom test is wrong — dropping it. Before I rebuild the gate, I want to match your intent rather than guess: by a `linearScan`-based solution, do you mean walking the decl's type subtree and only resolving it early when every symbol it references is already available in the signature phase — so any generic / `var x: T` defers, while a fully-concrete `seq[int]` resolves? Or a different use of `linearScan`?

Araq:
> so any generic / var x: T defers, while a fully-concrete seq[int] resolves

Yes, that is what I mean, but if test cases disagree with my intuition we need to take another look at this idea.

savannt:
Done — dropped the record-time syntactic gate entirely (`localHasAtomExplicitType` is gone). Every deferred toplevel `let`/`var` is now recorded, and `resolveDeferredLocal` decides at *resolve* time via a linear scan of the decl's type (`typeSymsAvailable`): it resolves early only when every name the type references is already available in the signature phase — each `Ident` declared in the current scopes/imports, each (template-injected) `Symbol` loading to a non-typevar decl. So a generic `T` (typevar) or any unavailable name defers and the `when` reference degrades to a clean "undeclared identifier", while `seq[int]` resolves.

Matches your intuition on every case I tried — `seq[int]`, aliases (`MyInt = int`), user objects, concrete generic instances (`Box[int]`), and forward-declared-later types all resolve; `var x: T` defers. Added `when/twhenlocalcompound` and `errmsgs/twhenlocalundeclared` to lock in both directions. Green locally across when/generics/templates/macros/const/consteval/basics/modules/types/errmsgs.

It's the atom-visiting analog of the `linearScan` primitive from the #2064 refactor (that one stops at each `ParLe`; this also needs the leaf `Ident`/`Symbol` tokens) — I left a note to fold `typeSymsAvailable` onto `linearScan` once it lands on master.

PR #2083, opened 2026-07-01
std/bitops: fixed-width unsigned bit manipulation

savannt:
Adds `lib/std/bitops.nim`, a Nimony port of the common bit-manipulation surface, over the fixed-width unsigned integers `uint8`, `uint16`, `uint32` and `uint64`:

* `bitand` / `bitor` / `bitxor` / `bitnot`
* `countSetBits` / `popcount` / `parityBits`
* `firstSetBit` / `leadingZeroBits` / `trailingZeroBits`
* `rotateLeftBits` / `rotateRightBits`
* `setBit` / `clearBit` / `flipBit` / `testBit`

`countSetBits`/`popcount` reuse system's `countBits32`/`countBits64`. The leading/trailing zero scans are simple width-loops, so everything lowers cleanly to the C, LLVM and JS backends (no compiler intrinsics required). Behaviour at `0` is **defined**, not undefined: `firstSetBit(0) == 0`, and `leadingZeroBits(0) == trailingZeroBits(0) == bit width`. Rotate amounts are taken modulo the bit width, so any amount is well defined (and shift-by-width UB is avoided).

### Why per-width overloads rather than one `SomeInteger` generic

Nim 2's `std/bitops` is generic over `SomeInteger`. That isn't currently expressible in Nimony:

* the bitwise operators (`shl`/`shr`/`and`/`or`/`xor`/`not`) resolve against **concrete** integer types — `shl`/`shr` are generic only in the shift *amount*, the value operand is concrete — so a body written against an abstract `T` fails to type-check (`expected uintN but got T`), for both generic `func`s and `[T]`-typed templates;
* `when sizeof(x)` on a generic type parameter is not usable (this is also why `system/countbits_impl.nim` hardcodes `const arch64 = true` instead of `when sizeof(int) == 8`).

So the API is provided as an overload set per unsigned width, which keeps every operation total and backend-neutral. Signed and `int`/`uint` overloads can be layered on later (same-size `cast[uintN]` gives the correct bit pattern).

### Tests

`tests/nimony/stdlib/tbitops.nim` exercises every function (values, edge cases at `0`, rotate wrap-around). Runs green via `nimony c -r`.

### DIFF
--- a/lib/std/bitops.nim
+++ b/lib/std/bitops.nim
@@ -1,0 +1,150 @@
+#
+#
+#            Nim's Runtime Library
+#        (c) Copyright 2024 Nim contributors
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## This module implements a series of low-level bit manipulation procedures
+## over the fixed-width unsigned integers `uint8`, `uint16`, `uint32` and
+## `uint64`.
+##
+## The procedures are generic and constrained by the `BitInteger` concept
+## below, which lists exactly the bitwise and shift operators the algorithms
+## rely on. The bodies are written against those operations only (plus a
+## runtime `sizeof` to recover the bit width), so they lower cleanly to C,
+## LLVM and JS alike, and any type providing the operator set is accepted.
+##
+## Bit indices are 0-based and, for the rotate procedures, the shift amount is
+## taken modulo the bit width so any amount is well defined.
+
+type
+  BitInteger* = concept ## Fixed-width unsigned integers usable with the bit
+                        ## manipulation procedures below. Satisfied by `uint8`,
+                        ## `uint16`, `uint32` and `uint64`.
+    func `and`(x, y: Self): Self
+    func `or`(x, y: Self): Self
+    func `xor`(x, y: Self): Self
+    func `not`(x: Self): Self
+    func `shl`(x: Self, y: int): Self
+    func `shr`(x: Self, y: int): Self
+    func `-`(x, y: Self): Self
+    func `==`(x, y: Self): bool
+
+template bitWidth[T: BitInteger](x: T): int = sizeof(x) * 8
+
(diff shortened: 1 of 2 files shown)


Araq:
Make it generic, use a concept for the generic constraints.

savannt:
Done — reworked to a single generic implementation. I added a small `BitInteger` concept that lists exactly the bitwise/shift operators the algorithms use, and every proc is now `[T: BitInteger]`:

```nim
type
  BitInteger* = concept
    func `and`(x, y: Self): Self
    func `or`(x, y: Self): Self
    func `xor`(x, y: Self): Self
    func `not`(x: Self): Self
    func `shl`(x: Self, y: int): Self
    func `shr`(x: Self, y: int): Self
    func `-`(x, y: Self): Self
    func `==`(x, y: Self): bool
```

With the concept carrying the operators, the bodies check fine against the abstract `T` (this was the piece I was missing before — thanks). Bit width is recovered at runtime with `sizeof(x) * 8` rather than `when sizeof`, and the population count now uses Kernighan's loop instead of the width-specific `countBits32/64`, so nothing is per-width anymore and it still lowers cleanly to C/LLVM/JS. `uint8/uint16/uint32/uint64` satisfy the concept; the existing `tbitops.nim` (all four widths) passes unchanged.

lib/std/bitops.nim
```diff
@@ -0,0 +1,150 @@
...
+#            Nim's Runtime Library
+#        (c) Copyright 2024 Nim contributors
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## This module implements a series of low-level bit manipulation procedures
+## over the fixed-width unsigned integers `uint8`, `uint16`, `uint32` and
+## `uint64`.
```
Araq:
```suggestion
## `uint64` -- or any type that satifies the `BitInteger` concept. 
```

PR #2084, opened 2026-07-01
std/lists: add singly and doubly linked lists

savannt:
Adds `lib/std/lists.nim`, a Nimony port of `std/lists` providing `SinglyLinkedList[T]` and `DoublyLinkedList[T]` and the standard operations:

* `initSinglyLinkedList` / `initDoublyLinkedList`, `newSinglyLinkedNode` / `newDoublyLinkedNode`
* `prepend` / `append` / `add` (node and value overloads; node overloads take `sink`, transferring ownership)
* `items` / `nodes` iterators
* `find` / `contains` (constrained to an `Equatable` concept) and `$` (constrained to `Stringable`) — the concept-per-operator pattern `std/options` uses
* `remove`

### Value semantics + non-recursive destructors

The lists have **value semantics** like `seq`/`string`: `=dup`/`=copy` deep-copy the chain, so copies are independent. Each list uniquely owns its nodes through the `head -> next -> …` chain; `tail` (and the doubly linked list's `prev` links) are non-owning `{.cursor.}` back-references, so the structure has **no reference cycle**.

Each list carries a custom, **non-recursive** `=destroy` (`.nodestroy`) that also frees the contained elements: it tears the chain down in a loop (moving each successor out before destroying its predecessor), so a multi-million-node list is reclaimed without overflowing the stack. `remove` lifts the target node out of the owning chain before splicing the gap shut, so it is never freed mid-operation.

### Rings deferred (for the right reason this time)

The ring variants are still omitted, but not because of `{.cursor.}` — a ring is a genuine **ownership cycle** (the last node's `next` points back at the head), and Nimony reclaims memory by reference counting **without a cycle collector**, so a ring would leak. Reclaiming it needs a different design (e.g. an owning spine + cursor closure) and is left for a follow-up.

### Tests

`tests/nimony/stdlib/tlists.nim` covers, for both list types: append/prepend ordering, `find`/`contains`, `remove` of head/middle/tail and down-to-empty (checking `tail`/`prev` stay valid so a later `append` still works), the iterators, deep-copy independence of `=dup` and `=copy`, and a `string`-element list exercising `$`. Runs green via `nimony c -r`. Separately verified that a 2,000,000-node list is destroyed without a stack overflow and that element destructors run.

### DIFF
--- a/lib/std/lists.nim
+++ b/lib/std/lists.nim
@@ -7,16 +7,31 @@
 #    distribution, for details about the copyright.
 #
 
-## This module implements a singly linked list and its operations.
+## This module implements singly and doubly linked lists and their operations.
 ##
-## This is the Nimony port. It currently provides `SinglyLinkedList[T]` only.
-## A singly linked list holds no reference cycles, so it needs no ownership
-## annotations; the doubly linked and ring variants of Nim 2's `std/lists`
-## depend on `{.cursor.}` (issue #1518) to break the `prev` back-reference
-## cycle and are therefore left for a follow-up.
+## This is the Nimony port. It provides `SinglyLinkedList[T]` and
+## `DoublyLinkedList[T]`.
 ##
-## The list stores `ref` nodes, so copying a `SinglyLinkedList` value shares the
-## underlying nodes (the same aliasing behaviour as Nim's `lists`).
+## Ownership and copying
+## ----------------------
+##
+## The lists have **value semantics**, like `seq` and `string`: copying a list
+## deep-copies its nodes, so the two values are independent. Each list uniquely
+## owns its nodes through the `head` -> `next` -> ... chain; the `tail` pointer
+## (and, in the doubly linked list, every `prev` pointer) is a non-owning
+## `{.cursor.}` back-reference, so the structure has no reference cycle and is
+## reclaimed by plain reference counting.
+##
+## Because of the unique-ownership invariant each list carries a custom,
+## **non-recursive** `=destroy` hook: freeing a very long list walks the chain
+## in a loop rather than recursing node by node, so it cannot overflow the
+## stack. The hook frees the contained elements too.
+##
+## The ring variants of Nim 2's `std/lists` are deliberately omitted: a ring is
+## a genuine ownership cycle (its last node's `next` refers back to the head),
+## and Nimony reclaims memory by reference counting without a cycle collector,
+## so a ring would leak. That needs a different design and is left for a
+## follow-up.
 
 type
   SinglyLinkedNode*[T] = ref object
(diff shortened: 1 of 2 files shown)


Araq:
> The doubly linked and ring variants of Nim 2's std/lists rely on {.cursor.} to break the prev back-reference cycle. Nimony doesn't have field {.cursor.} yet 

Not true, we do have that. We don't have the cursor-optimization yet which is a totally different thing.

Araq:
Also, while you are at it, give these lists a custom destructor (use `.nodestroy`, do not forget to free the elements too) so that it is non-recursive.

savannt:
Thanks — fixed both points in f6aeb0be:

1. **`{.cursor.}` is available.** You're right, my note conflated the field pragma with the cursor *optimization*. So this now also adds **`DoublyLinkedList[T]`**, with `prev` (and every list's `tail`) as `{.cursor.}` back-references — no cycle, reclaimed by plain refcounting.

2. **Custom non-recursive `=destroy`** (`.nodestroy`) for both lists, freeing the elements too. It walks `head -> next` in a loop, moving each successor out before destroying its predecessor, so the per-node destructor never recurses into `next`. Verified a 2,000,000-node list is torn down without a stack overflow, and that element destructors run.

Because a non-recursive teardown needs unique node ownership, the lists now have value semantics like `seq`/`string` (`=dup`/`=copy` deep-copy). `remove` lifts the node out of the owning chain before splicing the gap shut so it can't be freed mid-operation.

I left the **ring** variants out, but the real blocker isn't `{.cursor.}` — a ring is a genuine ownership cycle (last `next` -> head) and there's no cycle collector, so it would leak. Happy to do rings as a follow-up with an owning-spine + cursor-closure design if you'd like.

One heads-up worth flagging: dereferencing a generic node *inside* a `=destroy`/`=dup`/`=copy` hook currently trips a C-codegen bug (`… has no member named 'next_0'`), and `{.cursor.}` locals inside an inlined `iterator` crash `hexer`'s iterinliner (`n.kind in {Symbol, SymbolDef}`). I worked around both here (hooks delegate to ordinary procs; iterators use plain locals), but can open separate issues with minimal repros if useful.

Araq:
> I worked around both here (hooks delegate to ordinary procs; iterators use plain locals), but can open separate issues with minimal repros if useful.

Do not work around, fix the compiler bugs and once they are fixed, we will continue to work on this module.

lib/std/lists.nim
```diff
@@ -0,0 +1,351 @@
...
+## This module implements singly and doubly linked lists and their operations.
+##
+## This is the Nimony port. It provides `SinglyLinkedList[T]` and
+## `DoublyLinkedList[T]`.
+##
+## Ownership and copying
+## ----------------------
+##
+## The lists have **value semantics**, like `seq` and `string`: copying a list
+## deep-copies its nodes, so the two values are independent. Each list uniquely
```
Araq, line 19: Yeah... I didn't ask for that. Give the list nodes reference semantics.

savannt:

PR #2085, opened 2026-07-01
std/editdistance: add Levenshtein editDistance

savannt:
Adds `lib/std/editdistance.nim` with `editDistance(a, b: string): int` — the Levenshtein edit distance (minimum single-character insertions, deletions or substitutions to turn `a` into `b`), computed with a space-optimized single-row DP (`O(len a * len b)` time, `O(len b)` space).

### Scope

`editDistance` compares by `char` (i.e. by byte), so it corresponds to Nim 2's `editDistanceAscii`; for pure ASCII the two agree. A rune-aware variant over `std/unicode` (matching Nim 2's `editDistance`) is left for a follow-up.

### Tests

`tests/nimony/stdlib/teditdistance.nim` covers empty/identity inputs, single insert/delete/substitute edits, the classic vectors (`kitten`/`sitting` = 3, `flaw`/`lawn` = 2, `gumbo`/`gambol` = 2, `book`/`back` = 2, `sunday`/`saturday` = 3), symmetry, and repeated characters. Runs green via `nimony c -r`.

### DIFF
--- a/lib/std/editdistance.nim
+++ b/lib/std/editdistance.nim
@@ -1,0 +1,185 @@
+#
+#
+#            Nim's Runtime Library
+#        (c) Copyright 2024 Nim contributors
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## This module implements the
+## `Levenshtein edit distance <https://en.wikipedia.org/wiki/Levenshtein_distance>`_:
+## the minimum number of single-character insertions, deletions or substitutions
+## needed to turn one string into another.
+##
+## `editDistance` compares by Unicode `Rune`, matching Nim 2's `editDistance`.
+
+import std/unicode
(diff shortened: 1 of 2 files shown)


Araq:
> A rune-aware variant over std/unicode (matching Nim 2's editDistance) is left for a follow-up.

But why? Copy, paste&edit Nim's stdlib editDistance, how hard can it be.

savannt:
Done — reworked to be rune-aware, ported from Nim's stdlib `editDistance` (the common prefix/suffix stripping + triangle-skipping single-row DP). Verified against Nim's reference on the classic vectors plus a 500-pair random fuzz over multibyte input (café, Straße, Москва, 日本語, emoji, Greek) — all match. Dropped the "follow-up" note and added the multibyte assertions to the test.

PR #2086, opened 2026-07-01
std/sums: add numerically stable float summation

savannt:
Adds `lib/std/sums.nim` with two numerically stable floating-point summation routines over `openArray[float]`:

* `sumKbn` — Kahan-Babuska-Neumaier compensated summation (an improved Kahan sum that also handles the case where the running total is smaller in magnitude than the next term).
* `sumPairs` — pairwise (cascade) summation, whose error grows only logarithmically in the number of terms.

Both keep rounding error far below a naive left-to-right sum. `sumKbn` also recovers terms that would otherwise be lost when the running total dwarfs the next term — e.g. `sumKbn([1.0, 1e100, 1.0, -1e100]) == 2.0`, where a naive sum returns `0.0`.

### Scope

This is the Nimony port; it operates on `openArray[float]`. Generic (`SomeFloat`) and `float32` variants of Nim 2's `std/sums` are left for a follow-up. `sumPairs` recurses over index ranges (base case 128) rather than `toOpenArray` slices.

### Tests

`tests/nimony/stdlib/tsums.nim` covers basic sums, single-element and empty input, the compensated-cancellation case above, a 1000-element input that exercises the pairwise recursion past its base case, and exact negative/fractional cases. Runs green via `nimony c -r`.

### DIFF
--- a/lib/std/sums.nim
+++ b/lib/std/sums.nim
@@ -1,0 +1,55 @@
+#
+#
+#            Nim's Runtime Library
+#        (c) Copyright 2024 Nim contributors
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## This module implements numerically stable floating-point summation.
+##
+## Naively adding a sequence of floats accumulates rounding error that grows
+## with the number of terms. The routines here keep that error small:
+##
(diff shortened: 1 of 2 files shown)


planetis-m:
This is a separate package and was removed from stdlib, why add it back?

Araq:
Exactly.

PR #2087, opened 2026-07-01
std/md5: add MD5 message digest

savannt:
Adds `lib/std/md5.nim`, a Nimony port of the MD5 message-digest algorithm (RFC 1321):

* `type MD5Digest = array[16, uint8]`
* `toMD5(s: string): MD5Digest` — the raw 128-bit digest
* `` `$`(d: MD5Digest): string `` — lowercase hex rendering
* `getMD5(s: string): string` — hex digest of a string

> [!WARNING]
> MD5 is cryptographically broken (collisions are feasible) — use it only for checksums / non-security purposes.

### Scope

Exposes the one-shot API. The incremental `MD5Context` (`md5Init`/`md5Update`/`md5Final`) of Nim 2's `std/md5` is left for a follow-up. The implementation relies only on wrapping `uint32` arithmetic and an inline rotate (no `importc`), so it lowers cleanly to the C, LLVM and JS backends.

### Tests

`tests/nimony/stdlib/tmd5.nim` verifies the full RFC 1321 Appendix A.5 suite (`""`, `"a"`, `"abc"`, `"message digest"`, the alphabet, the alphanumeric string, the 80-digit string), plus multi-block inputs and the 56-byte padding edge case — all cross-checked against `md5sum`. Runs green via `nimony c -r`.

### DIFF
--- a/lib/std/md5.nim
+++ b/lib/std/md5.nim
@@ -1,0 +1,192 @@
+#
+#
+#            Nim's Runtime Library
+#        (c) Copyright 2024 Nim contributors
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## This module implements the `MD5 <https://en.wikipedia.org/wiki/MD5>`_ message
+## digest algorithm (RFC 1321).
+##
+## .. warning:: MD5 is considered broken for cryptographic use (collisions are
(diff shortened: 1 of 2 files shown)


Araq:
Do it properly and offer the incremental `MD5Context` (`md5Init`/`md5Update`/`md5Final`).

savannt:
Done

PR #2088, opened 2026-07-01
std/colors + std/rgb: terminal styling, and RGB values under a precise name

savannt:
This adds two small, composable "colors" modules — and, yes, proposes a naming swap. I'll make the case honestly and leave an easy escape hatch at the bottom.

## What's here

**`std/rgb`** — the RGB `Color` *value* type (`0xRRGGBB`): `rgb`, `extractRGB`, `$` → `#rrggbb`, `mix`, and named constants (`colRed`, `colOrange`, …). Pure data + colour math, no output device. This is exactly Nim 2's `std/colors` value type, unchanged in behaviour — just under a more precise name.

**`std/colors`** — fluent ANSI terminal styling in the spirit of npm `colors`:

```nim
echo "this is red and bold".red.bold
echo "warning".yellow
echo "selected".black.bgWhite
echo "brand".fg(rgb(255, 128, 0))   # truecolor, rendered from an std/rgb value
```

`"x".red.bold` is just `bold(red("x"))`; each styler emits its own reset so wrappers nest. Also includes `bg`/`fg` truecolor bridges to `std/rgb`, the raw `ansiFg`/`ansiBg`/`ansiSgr`/`ansiReset` generators, and `stripAnsi`/`visibleLen` for going back to plain text. It re-exports `std/rgb`, so the value type and its rendering compose from one import.

Both modules are covered by `trgb.nim` / `tcolors.nim` (exact escape-code assertions, chaining, truecolor, stripping) and run green under `nimony c -r`.

## The case for the swap (`colors` → terminal styling, values → `rgb`)

1. **Least surprise for the common case.** The overwhelmingly common reason a programmer types `import std/colors` is *"I want to colour my program's output."* Today that gives them an RGB value type that can't print anything — a papercut for exactly the newcomers the name attracts.

2. **Cross-ecosystem precedent.** In essentially every other ecosystem, the library called *colors* means terminal styling: npm `colors` and `chalk`, Python `colorama`/`termcolor`, Rust `colored`, Go `fatih/color`. Nim is the odd one out. Aligning removes a real cross-language surprise.

3. **`rgb` is the more precise name.** A 24-bit `Color` value + a bag of RGB constants is *literally* RGB; `std/rgb` says what it is. `colors` is the vaguer label, and it's better spent on the thing users reach for.

4. **This is the one cheap moment to do it.** Nimony is a fresh re-port for Nim 3.0 and already, deliberately, diverges from Nim 2 across the stdlib (e.g. `OrderedTable` is currently an alias, exceptions are re-modelled, many modules are documented "Nimony port"). Renaming during the re-port is free; doing it after `std/colors` ships as the value type is a breaking change.

5. **Cleaner separation, and they still compose.** Colour *math* lives in `rgb`; *rendering* lives in `colors`; the truecolor `fg`/`bg` bridge ties them together. Two focused modules beat one overloaded name.

## The homage

Making `std/colors` the terminal-styling module is also a small tip of the hat to npm `colors` — the package whose name taught a generation of programmers that "colors" means styled terminal output. `rainbow` is included as a wink. The 2022 self-sabotage is, needless to say, **not** included. 🌈

## Escape hatch

If keeping `std/colors` as the RGB value type is non-negotiable, I'm happy to flip the names: ship the value type as `std/colors` (Nim 2-compatible) and the terminal styling as `std/termcolors`. The two modules and the bridge stand on their own either way — the only question on the table is which one gets the short name.

### DIFF
--- a/lib/std/options.nim
+++ b/lib/std/options.nim
@@ -1,0 +1,101 @@
+#
+#
+#            Nim's Runtime Library
+#        (c) Copyright 2024 Nim contributors
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## This module implements types which encapsulate an optional value.
+##
+## A value of type `Option[T]` either contains a value `x` (represented as
+## `some(x)`) or is empty (`none(T)`).
+##
+## This is the Nimony port. Since exceptions are not yet wired up, `get` on an
+## empty `Option` terminates the program (mirroring `assert`) rather than
+## raising `UnpackDefect`. Equality is the compiler-provided structural
+## comparison: an empty `Option` always stores the default value, so two empty
+## `Option`s compare equal and an empty one never equals a populated one.
+
+import std/syncio
+
+type
+  Option*[T] = object
+    ## An optional type that may or may not contain a value of type `T`.
+    val: T
+    has: bool
+
+
+func raiseEmptyOption() {.noinline, noreturn.} =
+  {.cast(noSideEffect).}: quit "Tried to access the value of an empty Option"
+
+func some*[T](val: sink T): Option[T] =
+  ## Returns an `Option` that contains `val`.
+  runnableExamples:
+    let a = some(5)
+    assert a.isSome
+  result = Option[T](val: val, has: true)
+
+func none*[T: HasDefault](): Option[T] =
+  ## Returns an empty `Option` (i.e. with no value).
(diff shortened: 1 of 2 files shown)


savannt:
Reworked per @Araq in #2091: `std/colors` stays the RGB value type and the fluent `"x".red.bold` styling is added directly to `std/terminal`, no new module.

PR #2090, opened 2026-07-02
Fix nil checks for objects fields and mutating functions

Gesee-y:
# What was the problem

Nimony's nil checks were limited to root object, so code like the following crashed because Nimony didn't know how to check nil for the fields:

```nim
type
  TT = ref object
    x: int
    next: nil TT

proc main() =
  var a: nil TT = TT()

  if a != nil:
    if a.next != nil:
      a.next.x += 1
```

And also, It failed with mutating procs, so this program segfault instead of safely failing during compilation due to `a` potentially being `nil`:

```nim
var a: nil TT = TT()
if a != nil:
  a.x += 1

  var b = move(a)
  a.x += 1
```

# What was done

In order to fix it functions relative to nil checks had to take paths in to account:

- Added `pathToVarId: Table[seq[SymId], VarId]` to `NjvlContext`
- Added `analysableRootPath` variant that returns a `VarId` for compound paths
- Updated nil-check code to use path-aware VarIds
- `UnknownV` now cascades invalidation to all sub-paths (e.g. invalidating `a` also invalidates `a.next`)

# Open problem

When doing a `move` on a `not nil` variable, that var goes to nil. Which make further use of it a segfault:

```nim
var a: TT = TT()
a.x += 1
var b = move(a)
a.x += 1 # segfault here, since `a` is not nil, the compiler assume there is no need to check it.
```

This problem predate this PR, as `not nil` vars are always assumed to be `not nil`, even after a `move`


Araq:
We moved to contracts_fir.nim.

savannt:
Since nil-check analysis has moved from `contracts_njvl.nim` to `contracts_fir.nim`, I ported this change onto the new file so it applies against current `master`. The two files are structurally identical, so it was mostly mechanical:

Branch: https://github.com/savannt/nimony/tree/fix-nil-fields-fir (`savannt:fix-nil-fields-fir`)

What carried over to `src/nimony/contracts_fir.nim` (+107/-25):
- `SymPath` + `pathToVarId: Table[SymPath, VarId]` on `NjvlContext`, minting synthetic VarIds for paths longer than one symbol
- `allowDeref` threaded through `extractBorrowPath`/`extractPath` so a mid-path deref (a ref field) still yields an analysable path
- `translateCond`/`wantNotNil` derive & prove not-nil facts over compound paths via `getPathVarId`/`analysablePath`
- `invalidatePathsWithPrefix` cascades invalidation to sub-paths on store and on `var`/`out` argument passing

I left out two things that looked accidental in the original diff: the 1915-line `src/nifgram/examples/test_grammar.nim` and the trailing-newline deletions in `nimony_tags.nim`/`lenientnils.md`. Both tests and the doc Note are kept.

Verified (built `nimsem`, no `-d:useMalloc` needed since #2103):
- `tnil_fields` prints `1` / `1`; `tnil_mutate` compiles & runs clean
- unguarded `a.next.x` is still correctly rejected ("cannot prove expression is not nil: a.next") — so checking is proven, not disabled
- existing `notnil` suite (`tnew_raises`, `tnilbintree`, `tnilref`) unchanged, no regressions

Feel free to cherry-pick the commit (`1c374c76`) or pull the branch — whatever's easiest to fold into this PR. Happy to open it as a separate PR crediting you if @Araq prefers.

Gesee-y:
Nice @savannt if the fixes have been ported then it's fine, I will start trying to improve generics and wrapping types (such as lent T). Feel free to mention this or me, I'm not against publicity :p

savannt:
Opened #2110 with this work rebased onto `contracts_fir.nim` (per the "we moved to contracts_fir.nim" note — the analysis moved off the now-deleted `contracts_njvl.nim` in the finalir switch #2093). The commit there is authored by @Gesee-y so the credit stays with you. Feel free to close this one in favor of #2110.

Gesee-y:
Closed in favor of [#2110](https://github.com/nim-lang/nimony/pull/2110)

PR #2091, opened 2026-07-02
std/terminal: fluent npm-`colors`-style ANSI string styling

savannt:
Reworks #2088 to follow @Araq's guidance on that PR:

> nah. std/colors it is and "ansi" styling is part of terminal.nim, as before.

So this drops the new-module approach entirely. Instead of adding `std/colors`/`std/rgb`, it adds the npm `colors`-style **fluent string styling directly to `std/terminal`**, alongside the existing enum/`File` API. The name `std/colors` is left free to mean the RGB `Color` value type, as before.

### Usage

```nim
import std/terminal

echo "this is red and bold".red.bold
echo "warning".yellow
echo "selected".black.onWhite
```

Each styler is a `func (s: string): string` that wraps `s` in the matching ANSI SGR sequence and emits its own reset code, so `"x".red.bold` is exactly `bold(red("x"))` and the wrappers nest correctly.

### What's added to `std/terminal`

- Foreground: `black red green yellow blue magenta cyan white gray/grey` + `brightRed … brightWhite`
- Background: `onBlack … onWhite` — named `on<Colour>` (not `bg<Colour>`) to avoid clashing with the existing `BackgroundColor` enum values, and because `"x".black.onWhite` reads naturally in a chain
- Styles: `bold dim italic underline inverse hidden strikethrough`
- Helpers: `stripAnsi` and `visibleLen` for recovering plain text / measuring visible width

The existing `setForegroundColor`, `styledWriteLine`, enums, etc. are untouched.

### Deferred

A truecolor bridge (`"x".fg(rgb(255,128,0))`) is intentionally left out until a `std/colors` RGB value type exists to hang it on — at which point `terminal` can import it for `fg`/`bg`.

### Tests

Adds `tests/nimony/stdlib/tterminal.nim` covering colours, styles, chaining/nesting, ANSI stripping, and interop with the existing enum API. Passes locally via `bin/nimony c -r`.

### DIFF
--- a/lib/std/terminal.nim
+++ b/lib/std/terminal.nim
@@ -154,3 +154,117 @@ template styledWriteLine*(f: File) {.varargs.} =
     styledEchoProcessArg(f, x)
   write f, '\n'
   resetAttributes(f)
+
+# ------------------------------------------------------------------------
+# Fluent string styling, in the spirit of the npm `colors` package.
+#
+# In addition to the `File`-oriented procs above, these `func`s return the
+# styled *string*: each wraps its argument in the matching ANSI SGR escape
+# sequence, so they compose through method-call syntax. `"err".red.bold` is
+# just `bold(red("err"))`, and because every styler emits its own reset code
+# the wrappers nest correctly:
+#
+#   ```nim
+#   echo "this is red and bold".red.bold
+#   echo "warning".yellow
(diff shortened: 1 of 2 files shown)


Araq:
Sorry but we don't innovate in terminal.nim for now.

PR #2093, opened 2026-07-02
sema: use Final IR for frontend init/definite-assignment checking (contracts_fir)

savannt:
Frontend-only slice of @Araq's #2013, extracted per the "land the frontend part first" direction. Issue: #1946.

## What this does
Switches the sema frontend's initialization / definite-assignment checking from the nj/vl-based analyzer to the Final IR analyzer:
- `semmain` now calls `analyzeContractsFinalIr` (was `analyzeContractsNjvl`).
- New Final IR + analysis: `src/njvl/finalir.nim` (goto-form IR), `src/njvl/tracker.nim` (generic definite-assignment tracker), `src/nimony/contracts_fir.nim` (the frontend contracts pass, renamed from `contracts_finalir.nim`).
- Removes the now-dead `contracts_njvl.nim`: after the switch its only export `analyzeContractsNjvl` has no callers, and `sem.nim`'s `import contracts_njvl` was unused (removing it builds clean).

## What this deliberately excludes
The CPS portion of #2013 (`WIP: use finalir for CPS too`, `CPS: use better NJ pass…`, `gah`). `src/hexer/coro_transform.nim`, `src/hexer/cps.nim` and `src/njvl/nj.nim` are byte-identical to master; the `njvl/nj.nim`+`vl.nim` VMs are untouched. The CPS additions to `finalir.nim` were additive (extra procs + `toFinalIr(flatten, noCx)` defaulted off), so they aren't needed here.

## Testing
- `hastur build all` green.
- `tests/nimony/track` 9/9 (init/definite-assignment + def/use tracking).
- Init-checker behavior: valid code compiles; missing `result`/local init → `cannot prove that … has been initialized`.

Commits 1–5 are @Araq's (cherry-picked unchanged); the final commit renames `contracts_finalir.nim` → `contracts_fir.nim` and drops `contracts_njvl.nim`.


Araq:
It crashes with OOM due to Nim's stdlib sets.nim's `intersection` doing bad memory management. Avoid that.

savannt:
Status + a design question for @Araq.

This is the **frontend-only** slice of #2013 (the init/definite-assignment swap `analyzeContractsNjvl` → `analyzeContractsFinalIr`), leaving the CPS work out. Two issues turned up and are fixed here:

1. `finalir.nim` was missing the #2081 fix ("allow an lvalue call as a location-taking builtin's operand") that `nj.nim` has → `[Bug] call must have been bound to a location` on `twasmovedlvalue.nim`. Ported it.
2. The init-set intersection I hand-rolled to avoid `std/sets.intersection` (whose result is sized from `data.len`, the never-shrunk slot capacity → OOM on large procs) used a size-hinted `initHashSet` overload that host Nim accepts but nimony's own `std/sets` doesn't → the self-host boot failed with a type mismatch. Dropped the hint.

The remaining CI failure is a **precision gap in the Final-IR not-nil analysis that I don't want to fix without your input**, since it's core facts-analysis design.

`contracts_fir.nim` (`bindKeyBoth`) collapses **all** facts to ∅ at every loop-exit label:
```nim
of ekLabel:
  bindLabel(c.tr, key.label)
  if hadExit: c.facts.clearJournaled()   # drops even loop-invariant facts
```
So a not-nil fact about a variable that is *never assigned in the loop* is lost after the loop. `contracts_njvl` kept it by `save`ing facts before the loop and `restore`ing after. Minimal repro — false positive under finalir, accepted under njvl:
```nim
type DIR = object
  bpos: int
proc readdir(dirp: nil ptr DIR): int =
  if dirp == nil: return -1
  while true:
    if dirp.bpos >= 10: return 0
    var i = 0
    while i < 255: inc i
    dirp.bpos = i          # Error: cannot prove expression is not nil: dirp
    return dirp.bpos
```
This is exactly what fails on `lib/std/posix/posix.nim`'s `readdir` (9 `arc` native tests: `cannot prove expression is not nil: dirp`).

A sound fix would drop facts only about variables *assigned* in the loop body (keeping loop-invariant ones), which means finalir needs to snapshot/merge facts at `jmp`/`lab` joins the way the `Tracker` already does for init-sets — or mirror njvl's `save`/`restore`. Since that's your facts-plumbing design, I'd rather defer than guess: **how do you want facts handled at loop/label joins?**

Marking this a draft until that's settled.

src/hexer/xelim.nim
```diff
@@ -484,20 +510,52 @@ proc trCondOr(c: var Context; dest: var TokenBuf; n: var Cursor; tar: var Target
...
-      # `mustUseLabel` (cfvar lowering) is NJ-only. In `LowerCasts` mode we
-      # still want the binding/hoisting path inside `trAnd`'s `isComplex`
-      # branch, but never the cfvar form.
-      if mustUseLabel:
+      # `mustUseLabel` (cfvar lowering) is NJ-only. In `LowerCasts` and
+      # `TowardsFinalIr` mode we still want the binding/hoisting path inside
+      # `trAnd`'s `isComplex` branch, but never the cfvar form.
+      if c.goal == TowardsFinalIr and condPassthroughSafe(n):
+        tar.t.copyTree n
+        skip n
```
Araq:
```suggestion
        tar.t.takeTree n
```

src/hexer/xelim.nim
```diff
@@ -484,20 +510,52 @@ proc trCondOr(c: var Context; dest: var TokenBuf; n: var Cursor; tar: var Target
...
+      elif mustUseLabel:
         trCondAnd c, dest, n, tar
       else:
         trAnd c, dest, n, tar
     of OrX:
-      if mustUseLabel:
+      if c.goal == TowardsFinalIr and condPassthroughSafe(n):
+        tar.t.copyTree n
+        skip n
```
Araq: Likewise.

Araq:
Well that's why we added the journal to inferle.

savannt:
Done — used the journal as you suggested. `traverseLoop` already `rollbackTo`s the pre-loop facts (so `dirp != nil` is restored); the only thing dropping them was `bindKeyBoth` calling `clearJournaled()` at the loop-exit label. I now record each loop's trailing `(lab loopExit)` and skip the collapse for it, so the journal-restored facts survive — general forward-jump joins still collapse, since finalir doesn't snapshot facts per `jmp`. Matches njvl's save/restore.

Also applied both `takeTree` suggestions.

Locally the `readdir` repro is fixed, the self-host boot is clean, and the negative dirs (errmsgs / object / casestmt / exceptions / notnil) + arc stay green. I can't run the full `native` stage here (no `../nativenif` checkout), so I'll leave this a draft until CI confirms that leg, then mark it ready.

savannt:
Thanks — that `nativenif` checkout did it. `bin/hastur native` now runs here: **57/57** (the 9 `readdir` cases pass), and CI is green on all four platforms (linux-amd64/i386, macos-arm64, windows-amd64) with the full suite + self-host boot + native.

Marking ready for review.

Summary of the four commits on top of your frontend commits:
- `finalir`: port #2081 (allow an lvalue call as a location-taking builtin's operand) so `=wasMoved(s[i])` doesn't hit `[Bug] call must have been bound to a location`.
- `contracts_fir`: hand-roll the init-set intersection (avoids `std/sets.intersection`'s `data.len`-based over-allocation → the OOM you flagged) and keep it self-host compatible.
- `contracts_fir`: keep loop-invariant facts across loop exits via the journal (skip `clearJournaled` for loop-exit labels), fixing the `not nil` false positives.
- `xelim`: your `takeTree` suggestions.

Issue #2089, opened 2026-07-02
Static parameters (value generics)


Araq:
# Static parameters (value generics) for Nimony

Status: design. This document records the design for compile-time *value*
parameters — the replacement for Nim 2's `static[T]` type modifier.

## Goal

Allow constant values to appear in a generic argument list,
e.g. `Matrix[3, 4, int]` or `Vector[N, T]`, and have the value threaded to
wherever it is used (an array length, a field default, arithmetic on lengths)
or dropped if unused.

We explicitly want to avoid the two Nim 2 problems:

1. A `static[T]` **type modifier** that the whole compiler has to be aware of
   and constantly *skip*. Every `sameType`, every sigmatch path, every hexer
   lowering learns to unwrap it. That pervasiveness is the disease.
2. The overload-resolution asymmetry where an integer literal `3` has type
   `int` yet matches a `static int` parameter *better* than it matches `int`.
   That surprise comes entirely from solving a value parameter out of the
   *value* of a runtime argument.

### 1. A value parameter is an ordinary value; its type is the plain type

The type of a value parameter `N` is `int`, full stop. There is no `static[int]`
*type* in the checked type system. `static[int]` is **declaration syntax** only:

```nim
type Matrix[M, N: static[int]; T] = object
  data: array[M * N, T]
```

`static[int]` desugars to a value parameter whose type slot holds plain `int`.
The element type must be spelled — bare `N: static` is rejected — because we now
*compute* with `N` (see arithmetic below) and declaration-time checking of
`N + M` needs to know both operands are `int` before any instantiation exists.
A pure value parameter has no other inference source for its element type.

Because the type slot is always `int`, `sameType`, type-directed sigmatch, and
hexer never see anything but `int`. The "static-ness" is not in the type, so
there is no wrapper to skip anywhere in the type algebra. That property is the
whole point; everything below is arranged to preserve it.

### 2. Static-ness lives in the NIF **tag**, not in a flag or a pragma

An out-of-band boolean flag has no NIF representation, so every pass would have
to reconstruct it — exactly the hacky pervasiveness we are avoiding. A `.static`
pragma would push a *structural* distinction into the soft attribute bag, where
dispatch is a pragma scan that any pass can silently forget.

Instead, introduce sibling tags:

| ordinary        | static variant        | meaning                                   |
|-----------------|-----------------------|-------------------------------------------|
| `typevar`       | `staticTypevar`       | a generic parameter that is a value       |
[...]

PR #2101, opened 2026-07-03
static params: support enum value parameters (#2089 follow-up)

savannt:
Follow-up to #2089 (value / `static` generic parameters). While looking at test coverage for the feature, enum element types were the first gap: `staticTypevar` only bound scalar element types, so both

```nim
proc name[C: static[Color]](): int = ord(C)
echo name[green]()                 # was: Type mismatch

type Holder[C: static[Color]] = object
  x: int
var h: Holder[green]               # was: "must be a constant expression"
```

failed.

### Root cause
An enum field is the *canonical typed value* of its enum type — `annotateOrdinal` maps an ordinal back to exactly that field symbol. The matcher/sem folded an enum argument to its **bare ordinal**, which drops the enum type; the stored `int` value then fails re-checking as `got int64 but wanted <Enum>` when the instance is re-instantiated. The fix is to bind the enum field **verbatim** instead of folding it.

### Changes
- **semtypes** (`isStaticValue`): accept an enum-field symbol as a canonical static value, so type instantiation (`var h: Holder[green]`) binds it directly.
- **sigmatch** (`staticValueToBind`): bind an enum-field symbol verbatim; additionally fold a `const` symbol to the value it aliases and accept that (covers `const d = south; foo[d]()`). Integer `const` folding and the scalar cases are unchanged.

### Tests
`tests/nimony/generics/tstaticenum` — direct call, unification through an object type, mixed enum + int parameters, a holey enum (explicit ordinal values), a `const` alias, and an enum ordinal driving type-level array arithmetic. The existing `tstaticparams` output is unchanged.

### Scope
This covers **enum** value parameters. Aggregate value parameters (`object` / `array` / `tuple` / `openArray`) still fail and are left as a follow-up — they need a canonical aggregate-constant form plus structural equality, which is a larger design piece.


src/nimony/sigmatch.nim
```diff
@@ -547,9 +564,20 @@ proc staticValueToBind(m: var Match; elemType: Cursor; a: Cursor): Cursor =
...
+    elif isEnumFieldSym(a) and isOrdinalType(elemType, allowEnumWithHoles = true):
+      # an enum field is already the canonical typed value of its enum type;
+      # bind it verbatim (folding to the bare ordinal would lose the type)
+      result = a
+    else:
+      # a `const`: bind the literal it aliases, accepted exactly as if that
+      # literal had been written in the argument position
+      let folded = constSymbolValue(a)
+      if folded != default(Cursor):
+        result = staticValueToBind(m, elemType, folded)
```
Araq: Use the expreval engine already, but under a new mode that never shells out to `executeExpr`

savannt: Done in c1341909. Added a `noExecute` mode to `EvalContext` (`initEvalContext(c, noExecute = true)`): every `executeExpr` shell-out site now yields a local `cannot evaluate` result instead, so the engine stays fully in-process during overload resolution. `staticValueToBind`'s `const` branch now calls `expreval.eval` in that mode and re-types the folded value with the existing `annotateConstantType`; the hand-rolled `constSymbolValue` is gone. `annotateConstantType` maps the folded ordinal back to its enum field, so an enum-valued `const` (`const c = blue; viaConst[c]()`) still binds the field rather than a bare ordinal. `tstaticenum`, `tests/nimony/generics` (28/28) and `tests/nimony/overload` (11/11) pass.

PR #2103, opened 2026-07-03
nifcursors: never request a zero- / sub-FreeCell allocation in TokenBuf

savannt:
## Problem

`createTokenBuf(0)` — used in several njvl contract-analysis paths (`nj.nim` 842/944/950) — and the slow COW branch of `prepareMutation` allocated `sizeof(PackedToken) * cap` bytes directly. `PackedToken` is 8 bytes, so `cap == 0` is `alloc(0)` and `cap == 1` is `alloc(8)` — both below the default allocator's `sizeof(FreeCell)` (16 B on 64-bit) minimum.

`alloc` with a sub-`FreeCell` size is undefined for Nim's default (non-`useMalloc`) allocator: the request is rounded through the small-block freelist and corrupts the shared heap. Under `-d:useSysAssert` this fires:

```
[SYSASSERT] rawAlloc: requested size too small
```

In release it corrupts silently — benign on the current CI platforms (hence green CI) but a hard `SIGSEGV` in `rawAlloc` / `addToSharedFreeList` on some platforms (reproduced deterministically on WSL2 during `.requires` contract analysis). Building the compiler with `-d:useMalloc` only masks it, because `malloc(0)` returns a valid pointer.

## Fix

Route both raw-`alloc` sites through a small `allocStorage` helper that floors the request at 8 slots — the same minimum the grow path already uses, so an empty buffer's first append needs no extra realloc. `cap` semantics are unchanged (`createTokenBuf(0)` still reports `cap == 0`); only the physical backing size is floored. No behavioural change for non-empty buffers.

## Verification

On WSL2 without `-d:useMalloc` (default allocator, `-d:useSysAssert` on):
- the previously-crashing `.requires` contract-analysis path now compiles and runs;
- the full std-lib test suite passes **39/39** with zero crashes and zero regressions.

### DIFF
--- a/src/lib/nifcursors.nim
+++ b/src/lib/nifcursors.nim
@@ -446,8 +446,21 @@ else:
     else:
       if dest.data != nil: dealloc(dest.data)
 
+template allocStorage(cap: int): Storage =
+  ## Allocate backing storage for `cap` tokens, never requesting a zero- or
+  ## sub-`FreeCell` block. `alloc(0)` (cap 0) — and any request below the
+  ## allocator's `sizeof(FreeCell)` minimum, e.g. one 8-byte `PackedToken` —
+  ## is undefined for Nim's default (non-`useMalloc`) allocator: it rounds the
+  ## size through the small-block freelist and corrupts the heap. That fires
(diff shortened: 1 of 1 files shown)


Araq:
Amazing!!!

PR #2107, opened 2026-07-04
std/posix/io_uring

blackmius:


blackmius:
need tests

blackmius:
blocked by https://github.com/nim-lang/nimony/issues/2112

lib/std/posix/posix_other.nim
```diff
@@ -25,4 +45,35 @@ else:
...
+      msg_iov*: ptr IOVec    ## Scatter/gather array.
+      msg_iovlen*: cint      ## Members in msg_iov.
+      msg_control*: pointer  ## Ancillary data; see below.
+      msg_controllen*: SockLen ## Ancillary data buffer len.
+      msg_flags*: cint ## Flags on received message.
 
+
+when defined(linux):
+  var
+    MAP_POPULATE* {.importc, header: "<sys/mman.h>".}: cint
```
Araq, line 75: Not good enough, we want to be libc-free.

Araq: But can be done in a follow-up PR.

Issue #2104, opened 2026-07-03
Current limitations in `static[T]`


gmpreussner:
I'm gonna use this ticket to track known issues with the new `static[T]` feature, so we can work through them. The feature was originally added via https://github.com/nim-lang/nimony/pull/2092. The original design spec is in https://github.com/nim-lang/nimony/issues/2089

* * *

1. Only simple built-in numeric types (`int`, `float`, `string`, etc.) appear to work for `T`. However, `T` should support all types, including user defined types, without special casing in the compiler. Specific cases I have tested so far:
  - ~~Enumeration types do not work~~
  - `openArray[T]` does notwork

2. Overload resolution does not fold `static[T]` expressions. As per the design spec, we do not want to recover parameters automaticaly, i.e. by factoring or algebraic normalization across `static[int]` expressions. However, we should fold expressions when parameters are known to the compiler:
  - Example:
    ```
    proc foo[R, C: static[int]; T](x: array[R * C, T]) = discard
    foo[3, 2](...)
    ```
    `foo` should resolve if called with an `array[6, T]` argument

PR #2108, opened 2026-07-04
sem: support non-builtin `static[T]` values and fold known `static[int]` in overloading

savannt:
Works through the two limitations tracked in #2104, on top of the `static[T]` feature from #2092.

### 1. `static[T]` for non-builtin types
`static[T]` was restricted to primitive literal shapes, so user-defined types could not be used as value parameters. `isStaticValue` is now recursive and accepts enum fields, `const` symbols and typed aggregate constructors (array/set/tuple/object). A static value's type is matched against the declared element type, which also makes `static[openArray[T]]` bound to an array constructor work.

```nim
type Channel = enum red, green, blue
proc paint[C: static[Channel]](): int = ...
echo paint[green]()                 # ok

proc countValues[XS: static[openArray[int]]](): int = XS.len
echo countValues[[1, 2, 3, 4]]()    # ok
```

### 2. Fold known `static[int]` expressions in overload resolution
As per the design spec we don't recover parameters by algebra, but we now fold expressions when the parameters are already known, so array-length matching succeeds:

```nim
proc takesFlat[R, C: static[int]; T](x: array[R * C, T]): int = x.len
var flat: array[6, int]
echo takesFlat[3, 2, int](flat)     # resolves: R*C == 6
```

`matchArrayType` folds already-bound `static[int]` params (via `foldValueExpr`) before falling back to `lengthOrd`.

### Cleanup
The primitive-only `isStaticValue` in `semtypes` is removed in favour of the shared recognizer exported from `sigmatch`.

### Tests
Adds regressions to `tests/nimony/generics/tstaticparams` for enum `static[T]`, `static[openArray[int]]`, and `R * C` overload matching.

### Verification
- `nim c -r src/hastur build nimony`
- `tests/nimony/generics/tstaticparams` passes
- `tests/nimony/generics` → 27/27
- `tests/nimony/overload` → 11/11

Refs #2104, #2092, #2089.


src/nimony/sigmatch.nim
```diff
@@ -486,6 +486,124 @@ proc isStaticTypevar(s: SymId): bool =
   assert res.status == LacksNothing
   result = res.decl.symKind == StaticTypevarY
 
+proc sameStaticSymbol(a, b: SymId): bool =
```
Araq: Move these and the other helpers that do not use the `Match` object to a new helper module called `staticmatches.nim`

savannt:
Follow-up commit `00784de`: applied the same approach Araq asked for on #2101 (use the `expreval` engine under a mode that never shells out to `executeExpr`) here too.

It also fixes a real gap: a `const` bound to a value parameter was kept as a bare symbol, so `foldValueExpr` could not resolve it and length matching failed — e.g. `takesFlat[rows, 2, int]` with `const rows = 3` against `array[6, T]` reported `array[, T]`. `foldValueExpr` now resolves such leaf constants through `expreval.eval` in the new `noExecute` mode. `tests/nimony/generics` (27/27) and `tests/nimony/overload` (11/11) pass.

src/nimony/sigmatch.nim
savannt: Done in c73f705b. Moved `sameStaticSymbol`, `isStaticValue`, `staticValueType`, `staticOpenArrayElemType` and `staticValueTypeMatches` into a new `src/nimony/staticmatches.nim`; `sigmatch` and `semtypes` (via `sem`) now import it. The `Match`-bound bits (`evalStaticOrdinal`, `foldValueExpr`, `staticValueToBind`) stay in `sigmatch`. No behaviour change — tstaticparams, generics (27/27) and overload (11/11) still pass.

savannt:
You're right that most of this was subsumed — by #2101. I've **slimmed #2108 down to just the two parts that survive**, rebased onto current master (now 1 commit, +203/−7, mergeable).

**Subsumed by #2101 (dropped):**
- the `expreval.nim` `noExecute` change — now byte-identical to what #2101 merged (that's why the branch had gone `dirty`);
- **const value-param folding** — #2101's `foldStaticArg` already folds const/foldable symbols through the same engine;
- **enum-field static values** — #2101 put `isEnumFieldSym` into `isStaticValue`.

**Kept (not in master), both from the old `4afe0ea1`:**
1. **Folding an already-bound `static[int]` in array-length matching** — `foldValueExpr` gets a `Symbol` case that folds an inferred value typevar, and `matchArrayType` uses it (falling back to `lengthOrd`), so `proc foo[R, C: static[int]; T](x: array[R * C, T])` resolves for an `array[6, T]` once `R`, `C` are bound. Master's `matchArrayType` still uses `lengthOrd` only, so this doesn't resolve today.
2. **Aggregate / non-primitive `static[T]` values** — array/set/tuple/object constructor literals, plus a `static[openArray[T]]`/`varargs[T]` parameter satisfied by an array literal, with element-type matching (`staticValueType` / `staticOpenArrayElemType` / `staticValueTypeMatches`). Master's `isStaticValue` is still primitive-+-enum only.

The `staticmatches.nim` extraction you'd asked for is **not** in this slim version — #2101 landed those helpers inline, so extracting now would just churn them; I folded the two pieces above into the existing structure instead. Happy to do the module split as a separate follow-up if you still want it.

Verified locally (rebuilt `nimsem`): `tstaticparams` matches its `.output` (adds the aggregate-openArray value and `array[R * C, T]` matched with both literal and `const`-fed bounds), and a generics spot-check is 6/6 including #2101's `tstaticenum`.

src/nimony/sigmatch.nim
Araq: Not successfully "done", the code is still here, +146 lines added in total.

gmpreussner:
Hey @savannt, can you check if something like this works? I built this as a minimal example from various Nim features (compiles with Nim):

- using static[T] parameter to parameterize another parameter
- compile-time function evaluation on static expressions
- compile-time operation/iteration on static arguments

```
type
  Shape[N: static[int]] = object
    bounds: array[N, int]

template product*[N: static[int]](shape: Shape[N]): int =
  result = 1
  for extent in shape.bounds:                             # Error: type mismatch: got: void but wanted: int64
    result *= extent

type
  Box[N: static[int], B: static[Shape[N]], T] = object    # Error: not a type (N in Shape[N])
    data: array[product(B), T]                            # Error: undeclared identifier: `product`
```

If it doesn't work yet AND would require a lot of changes to make it work, we should circle back with Araq to confirm how it should be done. Thanks!

savannt:
Thanks for the sharp test case @gmpreussner — I ran it against this branch and it does **not** work yet. Breaking down why, because the failures are informative:

```
Shape[N: static[int]] value + array[N, int]            ✅ works today
[S: static[Shape2]] with a CONCRETE aggregate value    ✅ works (this PR)
proc foo[R, C: static[int]](x: array[R*C, T])          ✅ works (this PR)
B: static[Shape[N]]   (N parameterizing B's type)      ❌ "not a type"
product(B) as an array length                          ❌ depends on the above
```

The wall is the **dependent static type** `B: static[Shape[N]]` — a `static[T]` whose `T` is itself a generic instantiation depending on another (static) parameter. Sema rejects it with `Error: not a type` regardless of whether it's in a `type` section or a proc signature. Two reduced repros both fail the same way:

```nim
# in a type section
Box[N: static[int]; B: static[Shape[N]]; T] = object   # Error: not a type

# in a proc signature
proc foo[N: static[int]; S: static[Shape[N]]](x: int): int = x   # Error: not a type
```

What this PR added is narrower: it accepts **concrete** aggregate `static[T]` values (`[S: static[Shape2]]` where `Shape2` is fully concrete) and folds an already-bound `static[int]` inside an array length. It does not teach sema that one static param can parameterize the *type* of another, nor does it do compile-time evaluation of a function/loop over a static aggregate to produce an array length (`array[product(B), T]`).

(Two smaller, orthogonal notes on the snippet as written: `product` is a `template`, and templates don't inject a `result`, so that line errors independently; and `static[Shape[N]]` in a plain value-parameter position currently doesn't even parse — `undeclared identifier: static` — so aggregate `static[T]` value params only work via the `[S: static[...]]` generic-constraint form. Neither of those is the real blocker though.)

So: it **would require a fair amount of work** — dependent static types plus compile-time reduction of an aggregate `static` to an array length are both meaningful sema features, well beyond this PR's scope. Agree we should circle back with Araq on how he wants dependent `static[T]` (and static-driven array lengths) to be designed before anyone builds it. Happy to take a run at it once there's a direction.

savannt:
Pushed a follow-up commit (`548e382`) that makes **dependent static type parameters** work — i.e. gmpreussner's first two points now compile end-to-end:

- **a value param parameterizing another param's type** — `[N: static[int]; S: static[Shape[N]]]` — for matching, in type sections, everything;
- **compile-time operation/iteration on the static value** — `S.bounds[0]` and `for e in S.bounds: ...`.

```nim
type
  Shape[N: static[int]] = object
    bounds: array[N, int]

proc sumBounds[N: static[int]; S: static[Shape[N]]](): int =
  result = 0
  for e in S.bounds: result = result + e            # compile-time iteration
echo sumBounds[3, Shape[3](bounds: [10, 20, 30])]() # 60

type
  DepBox[N: static[int]; B: static[Shape[N]]; T] = object
    data: array[N, T]                               # dependent N reused as a length
```

Three small fixes (+53/−7), covered by an added block in `tstaticparams`:

1. `semStaticTypevarType` treated *any* `(at …)` element type as the `static[T]` sugar `(at static T)`, so a desugared `Shape[N]` → `(at Shape N)` had `Shape` skipped as if it were the `static` keyword and only `N` semchecked → *"not a type"*. Guarded with `isStaticSugarHead`.
2. `staticValueToBind` compared the element type structurally (`sameTrees`), which can't relate the dependent `Shape[N]` to the argument's concrete `Shape[2]`. When the element type is a generic `invoke`, it now unifies via `tryLinearMatch`, binding/equality-checking `N`.
3. The post-sem validator didn't list `staticTypevar` among the tags whose slot 3 is a type, so a dependent `(at Shape N)` slot was rejected as an expression (a plain `int` slot passed regardless, so only the dependent case tripped).

Negatives still reject correctly (wrong aggregate type; `N=3` vs `Shape[2]`), and a ~67-file generics/overload/sysbasics sweep shows no regressions.

@gmpreussner — the remaining third point (`array[product(B), T]`) needs `static[T]` as an *ordinary value parameter* (`func product(s: static[Shape[N]])`), which doesn't parse yet — only the `[S: static[…]]` generic-constraint form works — plus folding such a compile-time call into an array length. That's a bigger, separate piece (the deferred `staticParam` follow-up). I'm looking into it now; @Araq, I'll check the design with you before doing anything invasive there. (Side note: your `product` is a `template`, which in nimony/Nim doesn't inject a `result` — that error is independent of the static machinery.)

savannt:
@Araq — a design question before I go invasive on the remaining third of gmpreussner's example (`array[product(B), T]`, where `product` takes a `static` value). I've scoped it precisely; two independent pieces are needed and both touch design you'll have opinions on.

**(A) `static[T]` as an ordinary value parameter** — `func product(s: static[Shape[N]]): int`, `func dbl(n: static[int]): int`.
nifler renders the param type `static[int]` as `(at static int)`, which classifies as an `invoke`, so it hit `semInvoke` and tried to instantiate a type named `static` → "undeclared identifier: static". I have a small parse-layer fix (rewrite `(at static T)` → the `(static T)` modifier in the `InvokeT` branch), after which `func dbl(n: static[int]): int = n*2` compiles. The wall is the next layer: `singleArgImpl` has no `StaticT` case, so `dbl(5)` gives `BUG: unhandled type: static`. A `static[T]` param needs const-only args, bound as a value, and the proc **instantiated per value** (its result type `array[n, int]` depends on `n`).

The natural implementation is to **lower a `static[T]` param into an implicit `staticTypevar` generic parameter** — that reuses all the value-generic infra already here (#2092/#2101/#2108): value binding, `array[N*2, T]` folding, per-instance instantiation. But that means binding a generic parameter from a *value*-call argument (`dbl(5)`, not `dbl[5]()`), i.e. an implicit generic — which cuts against nimony's deliberate no-implicit-`typedesc`-generics stance.

**Q1:** do you want `static[T]` value params lowered to an implicit `staticTypevar` generic, or handled as a distinct static-param mechanism (param stays in the value signature, matched + instantiated by value)? Or should `static` value params simply be disallowed, requiring the explicit `[N: static[int]]` form (which already works, including the dependent `[S: static[Shape[N]]]` case from `548e382`)?

**(B) folding a *generic* proc call in an array-length position** — independent of (A).
`array[f(), T]` with a non-generic `f` already folds via the const-evaluator. But a generic call — even fully concrete `array[two[3](), T]` — fails with "invalid array index type", because `evalOrdinal`/`evalExpr` won't instantiate + evaluate a generic proc; and `array[two[N](), T]` isn't deferred and re-folded at instantiation either.

**Q2:** should the array-length const-evaluator instantiate and fold generic proc calls (and defer `two[N]()` to instantiation time), or do you consider a proc call in an array length out of scope for now?

Happy to implement whichever direction you prefer — I just don't want to build the implicit-generic lowering or the generic-call const-eval and have it be the wrong shape. Everything above is reproduced with minimal cases if useful.

gmpreussner:
> needs static[T] as an ordinary value paramete

You're right, I think Araq left this out of the original spec to keep it simple for the first iteration. I wouldn't be surprised if it requires further thought. There may be good arguments against supporting it.

I would recommend that we ignore this / leave it out for now. It can be a future update, if desired.

gmpreussner:
One other edge case we do need to consider now is the interaction of `static[T]` with the new `range` subtype implementation. The following may already work with your changes - we should add it to the unit tests:

```
type
  Box[N: static[int], T] = object
    data: array[N, T]

func item[N: static[int], T](box: Box[N, T], index: range[0..N-1]): T =
  box.data[index]
```

savannt:
Thanks @gmpreussner — agreed on both counts.

**`static[T]` value params:** left out for now, per your call. (I had a working parse-layer prototype but the matching/per-value-instantiation half is the invasive part, so I've dropped it from this PR — it can be a future update.)

**`static[T]` × `range` interaction:** good catch — it did *not* work yet, and now does (commit `5fcc19d`). `addRangeValues` was evaluating each range bound eagerly, so `range[0..N-1]` reported *"could not evaluate as ordinal"*. A bound that still mentions a value parameter is now kept symbolic and folded at instantiation, mirroring how `semArrayType` already defers a symbolic array length. Your example runs:

```nim
type
  Box[N: static[int], T] = object
    data: array[N, T]

func item[N: static[int], T](box: Box[N, T], index: range[0..N-1]): T =
  box.data[index]

var b: Box[3, int]
b.data = [10, 20, 30]
echo item(b, 1)   # 20
```

And, importantly, once folded the range-checking engine (#2075) discharges the obligations against the concrete range — so a value that can't fit is still rejected:

```nim
proc bad[N: static[int]](): int =
  var x: range[0..N-1] = N   # bad[3] => x: range[0..2] = 3
  int(x)
discard bad[3]()             # Error: value out of range: 3 notin 0..2
```

Added to `tstaticparams` as a unit test; the errmsgs range goldens (`trangecheck`, `tbadrange`) and a ~80-file sweep are unchanged.

gmpreussner:
@savannt we still gotta move the helper functions into a new module `staticmatches.nim`

Araq:
> You're right, I think Araq left this out of the original spec to keep it simple for the first iteration. I wouldn't be surprised if it requires further thought. There may be good arguments against supporting it.

Should be part of `semcompat.nim` which emulates Nim's implicit generics feature already. As for the other question of allowing values depending on other values... I'm still thinking about it.

Araq:
Also address my review, there is still stuff in sigmatch.nim that doesn't use Match and should be moved elsewhere!

src/nimony/sigmatch.nim
savannt: Done for real this time (`4ec111c`): moved `isStaticTypevar`, `sameStaticSymbol`, `isStaticValue`, `staticValueType`, `staticOpenArrayElemType` and `staticValueTypeMatches` into a new `src/nimony/staticmatches.nim`. sigmatch is −133 lines and only the `Match`-bound helpers (`foldValueExpr`, `staticValueToBind`, `bindStaticTypevar`, `matchesConstraint`, ...) remain. No behaviour change — `tstaticparams`, generics and overload still pass. (Last time I had folded them back inline during the slim; this actually extracts them.)

savannt:
Addressed the review — the non-`Match` static helpers now live in `staticmatches.nim` (`4ec111c`), so sigmatch only keeps the `Match`-bound logic. Thanks @gmpreussner for the nudge.

On `static[T]` value params: understood — when it's picked up it should go through `semcompat.nim`'s implicit-generics emulation rather than a bespoke path. I've left it out of this PR entirely (reverted the parse-layer prototype). Happy to take it on as a follow-up via `semcompat` whenever you've settled the "values depending on other values" question.

This PR is now: dependent static type params (`static[Shape[N]]`), the `static[T]`×`range[0..N-1]` fix, and the `staticmatches.nim` extraction — all with tests and no regressions.

PR #2111, opened 2026-07-04
hexer: fix two bugs blocking custom `ref`-container lifetime hooks

savannt:
Two independent `hexer` bugs that made writing custom lifetime hooks for linked-list-style `ref` containers impossible. Each is a distinct root cause with its own regression test.

### 1. `{.cursor.}` local inside an inlined iterator crashed the compiler

`replaceSymbol` (iterator inlining) renames an iterator body's locals. Its `VarS, LetS, CursorS, PatternvarS` branch assumed the tag is always a declaration and read the next token as the decl's name — but `CursorS` is *also* the tag of the `{.cursor.}` pragma `(cursor)`, which has no operand. Inlining an iterator with a `{.cursor.}` local recursed into `(pragmas (cursor))` and asserted `n.kind in {Symbol, SymbolDef}` on the `)`.

Fix: only rename when a `SymbolDef` actually follows the tag.

```nim
iterator walk(head: Node): int =
  var it {.cursor.} = head   # crashed the hexer before this PR
  while it != nil:
    yield it.value
    it = it.next
```

### 2. Ref field reads in `.nodestroy` bodies missed the payload hop

A boxed `ref` is a cell `{ rc; data }`, so a field read through it must reach the `data` payload: `(dot (deref p) field)` → `(dot (dot (deref p) d) field)`. The duplifier's full path does this in `trDeref`, but `.nodestroy` bodies go through `trOnlyEssentials`, which copied the dot verbatim. The generated C then dereferenced the ref *cell* and named a non-existent member (`… has no member named 'next_0'`), so any custom `=destroy`/`=copy`/`=dup` reading a field of a `ref` node failed to compile.

Fix: give `trOnlyEssentials` a `DotX` case that splices the payload hop in for a user-field read through a boxed ref. Lifter-emitted hooks already spell out the cell's own `d`/`r` fields, so those (matched by the reserved `DataField`/`RcField` names) and non-ref derefs are left untouched.

```nim
proc `=destroy`(x: List) {.nodestroy.} =   # failed to compile before this PR
  var it = x.head
  while it != nil:
    var nxt = move it.next
    `=destroy`(it); `=wasMoved`(it)
    it = move nxt
```

### Testing

New regression tests `tests/nimony/iter/titer_cursor_local.nim` and `tests/nimony/arc/tref_field_in_hook.nim` (the latter frees a 1000-node chain non-recursively and checks every element destructor ran). Locally green: `arc` 41/41, `iter` 11/11, `object` 18/18, `generics` 28/28. (`tests/hexer/setup.nim` fails on both this branch and master — it compiles `src/install.nim`, unrelated to these passes.)

### Not fixed here

Inline object *construction* (`newobj`) in a `.nodestroy` body — e.g. a `=dup` that allocates nodes directly — still needs the full ownership lowering that `trOnlyEssentials` deliberately skips; naively routing it through `trNewobj` compiles but double-frees. That is a deeper essentials-mode ownership question, left for a follow-up; hooks can construct via an ordinary helper proc in the meantime.

Context: unblocks the custom non-recursive list destructors requested in #2084.

### DIFF
--- a/src/hexer/duplifier.nim
+++ b/src/hexer/duplifier.nim
@@ -691,13 +691,52 @@ proc trOnlyEssentials(c: var Context; n: var Cursor)
         # generic statement: copy the head and recurse into the children
         copyInto c.dest, n:
           while n.hasMore: trOnlyEssentials c, n
-    of ErrX, SufX, AtX, DerefX, DotX, PatX, ParX, AddrX, NilX,
+    of DotX:
+      # Reading a *user* field through a boxed `ref` must reach the object
+      # payload, i.e. `(dot (deref p) field)` has to become
+      # `(dot (dot (deref p) d) field)`, exactly as `trDeref` does on the full
+      # path. Copying it verbatim — as this branch used to — left the payload
+      # hop off, so field reads in `.nodestroy` bodies dereferenced the ref
+      # cell itself and produced C accessing a non-existent member.
+      # Lifter-emitted hooks already spell out the cell's own `d`/`r` fields,
+      # so those (and non-ref derefs) are left untouched.
+      var probe = n
+      inc probe # -> object operand
+      var doHop = false
+      if probe.exprKind in {DerefX, HderefX}:
+        var operand = probe
+        inc operand # -> deref'd pointer
+        var typ = getType(c.typeCache, operand, {SkipAliases})
+        if typ.kind == ParLe and typ.typeKind == SinkT:
+          inc typ
+        if not cursorIsNil(typ) and typ.typeKind == RefT:
+          var field = probe
+          skip field # object operand -> field name
+          if field.kind == Symbol and
+             field.symId != pool.syms.getOrIncl(DataField) and
+             field.symId != pool.syms.getOrIncl(RcField):
+            doHop = true
+      if doHop:
+        let info = n.info
+        c.dest.addParLe DotX, info # outer dot: (… .field)
+        c.dest.addParLe DotX, info # inner dot: ((deref p) .d)
+        inc n # into the DotX, at the (deref …) operand
+        trOnlyEssentials c, n
+        c.dest.add symToken(pool.syms.getOrIncl(DataField), info)
+        c.dest.addIntLit(0, info) # inheritance
+        c.dest.addParRi() # close inner dot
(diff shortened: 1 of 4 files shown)


src/hexer/iterinliner.nim
```diff
@@ -434,12 +434,15 @@ proc replaceSymbol(e: var EContext; dest: var TokenBuf; c: var Cursor; relations
...
       inc c
-      let oldName = c.symId
-      let newName = pool.syms.getOrIncl("`lf." & $e.instId)
-      inc e.instId
-      relations[oldName] = newName
-      dest.add symdefToken(newName, c.info)
-      inc c
+      # `CursorS` is also the tag of the `{.cursor.}` *pragma* `(cursor)`, which
+      # carries no symbol; only rename when a real declaration name follows.
+      if c.kind == SymbolDef:
```
Araq: Ugly hack, instead ensure that pragmas do not recurse into here willy-nilly.

src/hexer/duplifier.nim
```diff
@@ -691,13 +691,52 @@ proc trOnlyEssentials(c: var Context; n: var Cursor)
         # generic statement: copy the head and recurse into the children
         copyInto c.dest, n:
           while n.hasMore: trOnlyEssentials c, n
-    of ErrX, SufX, AtX, DerefX, DotX, PatX, ParX, AddrX, NilX,
+    of DotX:
```
Araq, line 709: Looks copied&pasted...

savannt:
Both addressed in 6e31a07:

- **`iterinliner.nim` (ugly hack):** dropped the `kind == SymbolDef` guard and instead stopped `replaceSymbol` from recursing into pragma lists — a new `of PragmasS:` branch copies them verbatim (`dest.takeTree c`). Pragmas declare no locals to rename, and never descending into `(pragmas …)` means we never reach the `(cursor)` pragma whose tag collides with the `cursor` decl kind, so the decl-name read is unconditional again.
- **`duplifier.nim` (copied & pasted):** factored the shared ref-detection and payload-hop emission out of `trDeref` and the new `trOnlyEssentials` `DotX` case into `derefsBoxedRef` / `addDataFieldHop`, so the two paths use the same code and can't drift.

Locally green after the change: `arc` 41/41, `iter` 11/11, `object` 18/18, `generics` 28/28.

PR #2109, opened 2026-07-04
nifcore: add span/firstSon/isLastSon/takeTree/linearScan navigation primitives

savannt:
Adds the subtree-navigation helpers the frontend relies on (through `nifcursors` today) to `nifcore`, implemented **natively on the bounded-cursor / tag+jump model** rather than as ParLe/ParRi shims — groundwork toward porting nimsem to nifcore.

- `span` — alias of `subtreeWidth` (head + body token count; the frontend spells the subtree token-count `span`)
- `firstSon` — `inc` into the `TagLit` body
- `isLastSon` — `skip; not hasMore` (bounded `rem` replaces the ParRi sentinel)
- `takeTree` — `addSubtree; skip` (reuses the existing bulk subtree copy)
- `linearScan` — all-depth document-order `TagLit` walk, counting via `subtreeWidth` instead of ParLe/ParRi depth; supports early `break`

`tests/nimony/nifcore/tnifcore` is extended to exercise all five.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2113, opened 2026-07-04
nifcore: add navigation-primitive edge-case tests (#2109 follow-up)

savannt:
Follow-up to #2109 as requested ("In a follow-up PR, add more tests for this feature.").

Adds `tests/nimony/nifcore/tnavprims.nim`, a focused regression test for the five navigation primitives, covering the edge cases the `tnifcore.nim` smoke test doesn't:

- **span** — an atom is a single token; a subtree's span equals its head width plus the span of every child.
- **firstSon** — tag-first and atom-first children, and that it equals a plain `inc` of a copy.
- **isLastSon** — sole child, middle atom, trailing atom, and a subtree child.
- **takeTree** — the same-pool bulk-copy path draining every child in sequence, and the cross-pool re-interning path (fresh literal pool, shared tag pool) where pooled + inline strings and ints all survive the copy.
- **linearScan** — all-depth document order, the root itself excluded, atoms skipped, an early `break` leaving the cursor on the matching node, and a subtree with no nested tags yielding nothing.

**Stacked on #2109** — the first commit here is #2109's primitives (master doesn't have them yet, so the test can't compile against master alone). Merge #2109 first; this then rebases down to just the test commit.

Verified: the test's logic passes compiled under host Nim against `nifcore.nim`; CI covers the `nimony c -r` path.


savannt:
Closing — I mis-filed this. The follow-up-tests request was on #2075 (compile-time range checking), not #2109. Reopening the correct follow-up against that feature instead. Apologies for the noise.

PR #2114, opened 2026-07-05
Side effect-free floating point stringification and helpers

gmpreussner:


lib/std/system/formatfloat.nim
```diff
@@ -1558,12 +1563,14 @@ proc addFloat*(result: var string; x: float32) =
...
   for i in 0 ..< n: result.add buffer[i]
 
-proc `$`*(x: float): string =
+func `$`*(x: float): string =
   ## Outplace `addFloat` for `float`.
-  result = ""
-  result.addFloat(x)
+  var buffer = default(array[65, char])
+  let n = toChars(buffer, x, true)
+  floatCharsToString(n, buffer)
```
Araq: Making it a `func` is fine but why can't it use `addFloat` anymore?

gmpreussner: `addFloat` takes a `var` parameter, which implies side effects. We'd have to cast it.

gmpreussner: Ugh... it looks like we can make it a `func`. I could swear the compiler was choking on it recently. I was wondering why it wouldn't allow pass-through parameters. Maybe I mixed it up with something else. I'll push an update

PR #2117, opened 2026-07-05
Fix: let variable should be immutable

Gesee-y:
# What was the problem

`let` variable still allowed their fields to be mutated, which doesn't align with Nim 2, so it was fixed:

```nim
type
  MyObject = object
    field: int

let a = MyObject()
a.field = 1 # Should throw an error but doesn't
```

Same with `ref` types, even though Nim 2 allowed that behavior on `ref`s, which create inconsistencies

# What was done

- Checking if a variable is `let` skip field, and directly goes to the root
- Added `{.feature: "lenientlets".}` which make the behavior of `let` match the one of Nim 2 for `ref` types
- `let` variables can't be use in functions as `var` argument

# How it looks

## Let with object types

```nim
type
  MyObject = object
    field: int

let a = MyObject()
a.field = 1 # Will throw an error
```

## Let with ref types

```nim
type
  MyObject = ref object
    field: int

let a = MyObject()
a.field = 1 # Will throw an error, but in Nim 2 this would be valid
```


Araq:
Quod licet Iovi, non licet bovi. Give me the pure bugfix please, not another feature flag.

Gesee-y:
@Araq Well the issue is that fixing let for `object` also fix it for `ref`.
So I don't think I can just fix `object` and not ref as it would requires to explicitly tell the nimony to have that behavior for refs (which the feature flag already does)

lib/std/system/stringimpl.nim
```diff
@@ -167,13 +167,13 @@ func ensureUniqueLong(s: var string; oldLen, newLen: int) =
     copyMem(inlinePtrV(s), addr s.more.data[0], min(oldLen, AlwaysAvail))
   else:
     let newCap = if newLen > cap: max(newLen, ssResize(cap)) else: cap
-    let p = cast[ptr LongString](alloc(LongStringDataOffset + newCap))
+    var p = cast[ptr LongString](alloc(LongStringDataOffset + newCap))
```
Araq: Undo these changes.

src/nimony/contracts_fir.nim
```diff
@@ -1003,6 +1080,15 @@ proc analyseCallArgs(c: var NjvlContext; n: var Cursor) =
...
     if isMut:
       var inner = n
+
+      # Make sure the variable is `var`, else it error
+      let path = extractPath(c, inner, allowDeref=true)
+      if path.path.len > 0:
+        let rootSym = path.path[0]
+        let x = getLocalInfo(c.typeCache, rootSym)
+        if (x.kind in {LetY, GletY, TletY} and not c.lenientLets) and isInitialized(c, rootSym):
+          buildErr c, n.info, "Can't pass `let` variable as var argument"
```
Araq: But derefs.nim already performs this check...

src/nimony/semmain.nim
```diff
@@ -461,6 +461,8 @@ proc semcheckCore(c: var SemContext; dest: var TokenBuf; n0: Cursor) =
       reorderInnerGenericInstances(c, afterSem)
     var finalBuf = beginRead afterSem
     dest = injectDerefs(finalBuf, c.typeHooks, c.classes, c.thisModuleSuffix, c.g.config.bits)
+
+
```
Araq:
```suggestion
```

tests/nimony/threads/tpool1.nim
```diff
@@ -26,7 +26,7 @@ proc main =
   initPool()
 
   for i in 0 ..< NumTasks:
-    let frame = cast[ptr IncFrame](alloc(sizeof(IncFrame)))
+    var frame = cast[ptr IncFrame](alloc(sizeof(IncFrame)))
```
Araq:
```suggestion
    let frame = cast[ptr IncFrame](alloc(sizeof(IncFrame)))
```

src/nimony/contracts_fir.nim
Gesee-y: Yes but not for paths, it was still valid to pass a field coming from a `let` to a `var` argument

lib/std/system/stringimpl.nim
Gesee-y: Well if I do so nimony will complains (as `let` var are no more supposed to be modified after declaration)

Araq: How so, it's a `ptr LongString`, you can use pointers to write things.

Gesee-y: Yeah that's  what I was trying to say. The fix with `let` automatically fixed everything else.
So the type doesn't matter, nimony only check the variable.
So in order to tell it **not** to block mutation on a `let` var of a given type, it should be explicit in the code

Araq: It is not much of a fix if it breaks previously working and valid code. I don't know how else I can explain it to you that I want a mere bugfix, not some random changes that implement a spec that only exists in your head...

Gesee-y: OK, I will just put lenientLets to true, and it will fallback to Nim 2 behavior

Araq:
Fixed it my way: https://github.com/nim-lang/nimony/pull/2120

PR #2118, opened 2026-07-05
WebAssembly backend (lengwasm) — additive, sharing the JS backend's layout core

savannt:
# A WebAssembly backend (`lengwasm`) — additive, sharing the JS backend's layout core

This adds a fourth Leng codegen — after C (`codegen`), LLVM (`llvmcodegen`) and JS (`jscodegen`, #2043) — that emits real `.wasm` bytecode. It is **additive and user-selectable**, not a rewrite of anything.

**Stacked on #2043.** It reuses that PR's `jslayout` (the C-ABI byte-offset/`AccessKind` computer), so it needs #2043 first. Reviewing against master will show #2043's commits too; the WASM-only delta is 13 files (`git diff js-backend..wasm-backend`).

## Why it's additive: WASM linear memory *is* the linear-memory model

Under the Typed-Array direction (#2043), Nim-native data lives as bytes in one byte-addressable buffer and a pointer is an integer offset — which is exactly what WASM linear memory is (a single resizable `ArrayBuffer`, `i32.load`/`store` at byte offsets). So the layout is 100% shared; only the emitted *instruction* differs:

| | JS backend | WASM backend |
|---|---|---|
| layout | `jslayout` byte offsets | **`jslayout`, unchanged** |
| field load `(dot p f)` | `mem.getI32(p+off)` (JS text) | `i32.load` at `p+off` (bytecode) |
| output | a `jsnif` text tree → `.js` | a `WasmModule` → `.wasm` |

The split mirrors `llvmcodegen`/`llvmserializer`: `wasmcodegen` translates Leng → a `WasmModule`; `wasmenc` (Leng-agnostic — LEB128/sections/opcodes) encodes it and is independently unit-testable.

## Scope of this first slice

Pure-compute + linear-memory scalars: `i32`/`i64`/`f32`/`f64` params/locals, arithmetic (**WASM ints wrap natively — the 32-bit-wrapping fix the JS backend needed is free here**), comparisons, bit ops, int/float conversions; structured control flow (`if`/`while`/`break`) → `block`/`loop`/`if` with relative `br` depths (the `.c.nif` is already structured — no raw `goto` — so no relooper); direct calls + recursion; and object-field / array-element / deref load+store at `jslayout` offsets.

**Entry model:** a WASM module is instantiated by a host that calls its exports, so it emits a function per `exportc` proc and exports it, rather than the C `main(argc,argv,envp)` shim (argv wiring / cross-module `ini` are C-runtime conventions). Module-level globals, `string`/`seq` heap and `echo` (which need a start function and the ported allocator over an imported `WebAssembly.Memory`) are the next increment, on this same core. Unsupported nodes emit `unreachable` (the module still validates) and are reported on stderr.

## Tests

`tests/wasmbackend/` is a hastur custom runner (like `tests/jsbackend/`). Each `t*.nim` is compiled to a real `.wasm`, which **Node's WebAssembly engine validates before executing**, and a sibling `t*.js` driver instantiates it, calls its exports, and prints; stdout is diffed against a golden.

- **`tcompute`** — arithmetic, `if`/`while`, `break` (via `jmp`→`block`+`br`), recursion (`fib`), `int64` (`i64.mul`), `float64`.
- **`tmemory`** — a `Point{x,y,z:int32}` laid out in the shared `WebAssembly.Memory`; `dot`/`manhattan` read fields and `setY` writes one, each an `iN.load`/`store` at the **`jslayout`** byte offset — proving the layout core is shared.

Both green (`hastur tests/wasmbackend`). `node` required; the dir carries `hastur.mode = skip` so this WIP suite stays out of the default sweep.

## Direction question

This is deliberately additive and user-selectable — it doesn't touch the C/LLVM/JS paths. Posting it to get your read on whether a WASM target belongs in-tree (sharing the layout core, near-free to maintain alongside) vs. as a separate repo consuming `.c.nif`. Happy to adjust scope or hold pending your view on the #2043 direction.


savannt:
**Update — `echo "hello world"` now runs end-to-end as WASM.**

Since the initial post the backend has grown from "compute + host-provided pointers" to running a real program, all on the same reused `jslayout` core. Five commits, `tests/wasmbackend` at 5/5 green under Node's `WebAssembly` engine (which validates every module before executing it):

| commit | what |
|---|---|
| aggregate construction | objects/arrays built into a bump-allocated region (`$hp`, the WASM analog of `allocFixed`); `oconstr`/`aconstr`, field stores, value-semantic copy (`memory.copy`) |
| multi-module linking | a call to a foreign *defined* proc is pulled in via `getDeclOrNil` and emitted as real WASM (transitively); only `importc` primitives stay host imports. + call-boundary scalar coercion |
| module globals | module-level vars/consts (this module's and foreign ones like `stdout`) as fixed memory slots |
| whole-program + data segments | emit the C `main` + its whole closure; string-literal consts laid out as WASM data segments |

**`echo "hello world"` → `main` → `ini` (init chain) → `write(stdout, str)` → `fwrite` (host), string materialised from a data segment.** The only host imports are the C stdio primitives (`fwrite`/`fputc`) — the same seam `runtime.js` fills for the JS backend. `techo` is in the suite; multi-`echo` works too.

The point that keeps proving out: because WASM linear memory *is* the JS backend's linear-memory model, `jslayout` is reused verbatim — the bounds check `nimIcheckB` links in from `system` as real WASM, the string's `LongString` header + `data` flexarray are laid out at the same byte offsets the runtime reads back, etc. Nothing in this stack is WASM-specific beyond the instruction selection.

**Next:** `echo` of an integer (int→string; one unhandled node in that path), then heap `string`/`seq` growth over the ported allocator.

Still a draft, and the direction question stands — is a WASM target something you'd want in-tree (sharing this core), or as a separate repo consuming `.c.nif`? Happy to adjust scope either way.

savannt:
Relocated to **[nim-lang/jantak](https://github.com/nim-lang/jantak)** — closing.

Per Araq's decision, the WebAssembly backend now lives in nim-lang/jantak, imported in **nim-lang/jantak#2** (alongside the JavaScript backend this was stacked on, #2043). The one idea it demonstrated carries over intact: WASM linear memory *is* the JS backend's `ArrayBuffer` model, so both backends drive one shared `jslayout` and differ only in instruction selection.

`echo` runs end-to-end there (5/5 suite green under Node's WASM engine), and the plugins build against a clean nimony `master` — nothing from this PR needs to stay here. Continuing in jantak.

savannt:
I've closed this and withdrawn the branch. This code is not being contributed here — it lives in my own repository under my own license. Please do not reopen or publish it. Thanks.

PR #2119, opened 2026-07-05
Added range iterators for counting down

gmpreussner:
This change introduces two new `range` iterators:

```
iterator `>..`*[T: Ordinal](a, b: T): T     # count down in the range `(b, a)`, for `a` >= `b`
iterator `>..<`*[T: Ordinal](a, b: T): T    # count down in the range `[b, a]`, for `a` >= `b`
```

My rationale was as follows:
- There are too many combinations of up/down/inclusive/exclusive iterators. Making them unambiguous, short, and intuitive is probably not possible. Most of the combinations are also not really used in practice. Therefore, I added only the two most common use cases, i.e. `[N..0]` and `[N-1..0]`
- the `>` on the left of `..` indicates that we are counting from the higher bound (on the left) to the lower bound (on the right)
- for symmetry with the two existing iterators, `>..` is down-count version of `..`, and `>..<` is the down-count version of `..<`

Example:
- `n >.. 0` counts from `n` down to `0`
- `n >..< 0` counts from `n-1` down to `0`

I'm open to suggestions on the naming of these iterators. Specifically, I interprete `>` as "counting down" and `<` as "not including the higher bound". I understand that this is not inline with the existing iterators where `<` means "not including the bound to the right".

## Update 2026/07/13

Only the `>..` and `countup` iterators are in this change now. `>..<` has poor readability, and other bound combinations can be covered by `countdown` and `countup`

### DIFF
--- a/lib/std/system/iterators.nim
+++ b/lib/std/system/iterators.nim
@@ -20,14 +20,6 @@ iterator `>..`*[T: Ordinal](a, b: T): T {.inline.} =
     if i == b: break
     dec i
 
-iterator `>..<`*[T: Ordinal](a, b: T): T {.inline.} =
-  ## Count down in the range `[b, a)`.
-  var i = pred(a)
-  while i >= b:
-    yield i
-    if i == b: break
-    dec i
-
 iterator countdown*[T, V: Ordinal](a, b: T; step: V = T(1)): T {.inline.} =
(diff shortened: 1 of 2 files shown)


Araq:
Only `>..` is easy to read and not covered by `countdown` already.

PR #2121, opened 2026-07-06
checks if `tools/gen_tags` was executed

demotomohiro:
When you changed `doc/tags.md` but forgot to run `tools/gen_tags.nim` and sent the pull request, `checkGenTags` proc in `tools/tester.nim` detects that files in `src/models/*.nim` are not updated.


Araq:
but we face the opposite problem: Large language models and its reviewers not noticing the Nim code was updated with new tags but tags.md did not!

demotomohiro:
I updated the error message.

All files in `src/models` has
```nim
# Generated by tools/gen_tags.nim from doc/tags.md. DO NOT EDIT!\n
```
at the first line.
Large language models can ignore the message and change files in `src/models` without updating `doc/tags.md`?
Even if some LLM or contributers don't read the message and change files in `src/models` instead of `doc/tags.md`, `checkGenTags` prints error for such a case.

Should `checkGenTags` proc in this PR checks if a PR from contributer changed `doc/tags.md` or `src/model/*.nim` using git command and prints an better error message?

PR #2124, opened 2026-07-07
defining `>=`, `>` or `!=` operator generates compile error

demotomohiro:
Adds the same error check code as https://github.com/nim-lang/Nim/pull/25787


gmpreussner:
@demotomohiro is there still something left to do here? It looks like some commits were already merged.

Araq:
Will do it my way.

PR #2128, opened 2026-07-09
Fold const aliases in static generic type arguments

gmpreussner:
Re-type const symbols via annotateConstantType in semStaticInvokeArg so type instantiations using a const canonicalize to the same instance as the value they name.


src/nimony/semtypes.nim
```diff
@@ -527,8 +534,14 @@ proc semStaticInvokeArg(c: var SemContext; dest: var TokenBuf; n: var Cursor;
     endRead(dest)
   else:
     var value2 = value
-    var folded = evalExpr(c, value2)
-    let f = beginRead(folded)
+    let constAlias = isConstSym(value)
+    var folded = evalExpr(c, value2, if constAlias: elemType else: default(Cursor))
```
Araq: This should all be done by `evalExpr` to begin with. But maybe it does need the `elemType`.

src/nimony/expreval.nim
```diff
@@ -994,13 +994,27 @@ proc eval*(c: var EvalContext; n: var Cursor): Cursor =
...
+  result = res.status == LacksNothing and res.decl.symKind == ConstY
+
 proc evalExpr*(c: var SemContext, n: var Cursor;
-               expectedType: TypeCursor = default(Cursor)): TokenBuf =
+               expectedType: TypeCursor = default(Cursor);
+               retypeConstAlias = false): TokenBuf =
+  ## Evaluate a constant expression. When `retypeConstAlias` is set and `n` is
+  ## a `const` symbol, the folded value is re-typed via `annotateConstantType`
+  ## so it canonicalizes to the typed static value the name aliases.
+  let retypeConst = retypeConstAlias and not cursorIsNil(expectedType) and isConstSym(n)
```
Araq:
```suggestion
  let retypeConst = retypeConstAlias and not cursorIsNil(expectedType)
```

gmpreussner:
Superseeded by https://github.com/nim-lang/nimony/pull/2164

PR #2131, opened 2026-07-10
Fixed generic proc name / module name clash

### LINKED ISSUE #2130: Generic proc with same name as module tries to instantiate module symbol instead of proc
## Minimal Repro

_deps/foo.nim_
```
proc foo*[N: static[int]; T](elems: array[N, T]): array[N, T] = elems
```

_tmodprocnameclash.nim_
```
import deps/foo

let xs = foo[3, int]([1, 2, 3])  # Error: cannot call expression of type auto
```

## Observed Behavior

If a module imports another module `foo` containing a generic proc `foo[N, T]`, Nimony currently rewrites `foo[N, T]` as a subscript on the module symbol and then fails the surrounding call with "cannot call expression of type auto".

This error is emitted from _semcall.nim_ when overload resolution reaches a call whose callee:
- is not an overload set (`OchoiceX` / `CchoiceX`),
- is not an identifier left for later resolution,
- does not have a routine/proc type,
- and is not a symbol that can be re-looked up as overloads.

So Nimony thinks that we are trying to call something whose type is still `auto`, not a callable routine.

In Nimony’s `semCall` handling of `@` (`AtX`) nodes, the callee is sem-checked with `{KeepMagics, AllowUndeclared}` but without `AllowOverloads`. Lookup for bare `foo` can therefore hit the module symbol first, not the constructor overloads.

Then two things go wrong:
1. **Not recognized as generic instantiation**:  Nimony only sets `treatAsGenericInst` if the resolved symbol is a generic routine, or if `atHasTypeArgs` sees a type as the first argument. Here the first argument is `N`, a `static[int]` value generic parameter, not a type. So `atHasTypeArgs` returns false.
2. **Falls back to subscripting**: Because it is not treated as generic instantiation, Nimony calls `semBuiltinSubscript`, which rewrites `foo[N,  T]` as a `[]` call on whatever matrix resolved to. With the module symbol, that produces an expression of type `auto`.

The outer call then tries to invoke that `auto` value

##  Expected Behavior

`foo[N, T](...)` must instantiate the generic proc `foo`, not treat bare `foo` as the imported module symbol when `import deps/foo` brings both a module and a proc named `foo` into scope.

## Workaround

Alias the import and qualify proc access to avoid the collision:
```
import deps/foo as bar

let xs = bar.foo[3, int]([1, 2, 3])
```

gmpreussner:
Fixes https://github.com/nim-lang/nimony/issues/2130

The first commit fixes the general name clash.
The second commit fixes overloads of the affected proc.


gmpreussner:
@Araq We can probably do better here; would love your guidance.

Araq:
When a symchoice is generated for a caller `fn` expression, do not add modules as candidates of the sym choice. Much easier.

src/nimony/sembasics.nim
```diff
@@ -19,22 +19,31 @@ import ".." / gear2 / modnames
...
       for foreignName in stylesOfIface(iface[], realName, ignoreStyle):
         let candidates = addr iface[].getOrQuit(foreignName)
-        inc result, candidates[].len
         for defId in candidates[]:
+          if option == FindOverloads:
+            let res = tryLoadSym(defId)
+            if res.status == LacksNothing and res.decl.symKind == ModuleY:
+              continue
+          inc result
           dest.add symToken(defId, info)
```
Araq:
```suggestion
          if option == FindOverloads:
            let res = tryLoadSym(defId)
            if res.status == LacksNothing and not isRoutine(res.decl.symKind):
              continue
          inc result
dest.add symToken(defId, info)
```

src/nimony/sembasics.nim
```diff
@@ -130,6 +135,8 @@ proc rawBuildSymChoice(c: var SemContext; dest: var TokenBuf; identifier: StrId;
     var nonOverloadable = 0
     for k in stylesOfScope(it, identifier, ignoreStyle):
       for sym in it.tab.getOrDefault(k):
+        if option == FindOverloads and sym.kind == ModuleY:
```
Araq:
```suggestion
        if option == FindOverloads and not isRoutine(sym.kind):
```

PR #2144, opened 2026-07-12
Fix various generic iterator related semcheck crashes

### LINKED ISSUE #2143: Proc calls inside `do`-block closures crash compiler when enclosing generic is constrained by concept
## Minimal Repro
```nim
type
  MagnitudeOps* = concept
    proc max(a, b: Self): Self

  RealScalar* = concept of MagnitudeOps
    proc `+`(a, b: Self): Self
    proc zero(_: typedesc[Self]): Self

proc fold*[N: static[int], T, A](values: array[N, T], initial: A, combine: proc(acc: A, value: T): A {.closure.}): A =
  result = initial
  for value in values:
    result = combine(result, value)

func test1*[N: static[int], T: RealScalar](values: array[N, T]): T =
  fold(values, T.zero) do (acc, value: T) -> T:
    if acc <= value: value else: acc                           # this is OK - no proc call

func test2*[N: static[int], T: RealScalar](values: array[N, T]): T =
  fold(values, T.zero) do (acc, value: T) -> T:
    max(acc, value)                                            # derefs.nim(595) `fnType.isParamsTag` [AssertionDefect]
```

## Problem Description

Semcheck is producing call trees whose callee type could not be walked as a routine signature.

Two things combined:

1. Concept dispatch for `max(acc, value)` inside a generic `do`-block causes sem to synthesize a helper proc (using `if`/`<=`) and to leave overload sets as `(ochoice ...)` nodes.
2. `skipProcTypeToParams` only handled proctype nodes (`ProctypeT`, etc.), not routine declarations (`proc@`, `func@`, …). For those, `typeKind` is `NoType`, so the helper returned the cursor unchanged.

Downstream code then expects `(params ...)` and hits:
```
derefs.nim(595) `fnType.isParamsTag` [AssertionDefect]`
```

gmpreussner:
Fixes https://github.com/nim-lang/nimony/issues/2143

Routine decl cursors were not handled by `skipProcTypeToParams`, so derefs failed on synthesized helpers and unresolved overload sets. Resolve `ochoice` callee types and guard `hasRtti` against type variables.

## Changes:

1. **decls.nim**: treat routine decls via `symKind`:
```
elif kind in RoutineTypes or sk in RoutineKinds:
```
(and the same check in `skipProcTypeToParams`)

2. **typenav.nim**: when a callee is still an `(ochoice ...)`, use the first symbol’s routine type instead of `(auto)`.

3. **typeprops.nim**: `hasRtti` returns `false` for typevars (avoids a follow-on crash in hexer when sem generates generic helper procs).

## Regression Tests

- `tests/nimony/concepts/tconceptdocall.nim`

### DIFF
--- a/src/hexer/intramodinliner.nim
+++ b/src/hexer/intramodinliner.nim
@@ -396,6 +396,7 @@ proc slotRootOf(c: Cursor): SymId =
   ## it lets a param that is only ever *written through* still count as
   ## value-stable (its pointer value never changes), so its argument can be
   ## substituted instead of copied.
+  result = SymId(0)
   var n = c
   while true:
     case n.kind
--- a/src/nimony/decls.nim
+++ b/src/nimony/decls.nim
@@ -74,10 +74,11 @@ proc skipToParams*(c: var Cursor) =
   ## Proctype/itertype layout: `(<tag> <NilTag> (params...) RetType <Pragmas>)`.
   ## Proc-decl layout: `(proc Name ExportMarker Pattern Typevars (params...) ...)`.
   let kind = c.typeKind
+  let sk = c.symKind
   inc c # skip ParLe
   if kind in {ProctypeT, ItertypeT}:
     skip c # nilability tag
-  elif kind in RoutineTypes:
+  elif kind in RoutineTypes or sk in RoutineKinds:
     skip c # name
     skip c # export marker
     skip c # pattern
@@ -100,7 +101,7 @@ proc skipRoutineDeclPrefix*(n: var Cursor; parentKind: TypeKind) =
 proc skipProcTypeToParams*(t: Cursor): Cursor =
   ## Pure version: returns a cursor advanced past the prefix slots.
   result = t
-  if result.typeKind in RoutineTypes:
+  if result.typeKind in RoutineTypes or result.symKind in RoutineKinds:
     skipToParams result
 
 const
--- a/src/nimony/decls.nim
+++ b/src/nimony/decls.nim
@@ -174,7 +174,8 @@ type
     body*: Cursor
 
 proc isGeneric*(r: Routine): bool {.inline.} =
-  r.typevars.substructureKind == TypevarsU
+  if not isRoutine(r.kind): return false
+  result = r.typevars.substructureKind == TypevarsU
 
 proc takeRoutine*(c: var Cursor; mode: SkipMode): Routine =
   let kind = symKind c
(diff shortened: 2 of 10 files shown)


gmpreussner:
Changing this to draft because there are still cases where the compiler crashes

gmpreussner:
When calling a `.closure.` proc parameter (`predicate(value)`) inside a `for`/`if` body, during overload resolution, semcheck did:
```
isGeneric(getProcDecl(finalFn.sym))
```
For closure proc calls, `finalFn` is not always a routine symbol. `getProcDecl` then returns a mostly-empty `Routine` with an invalid `typevars` cursor, and `isGeneric` crashes on it.

## Additional Changes

- **decls.nim**: guard `isGeneric(Routine)`
- **semcall.nim**: guard instantiation path
- **intramodinliner.nim**: Fixes `Error: Cannot prove that 'result' is initialized. [ProveInit]` during hexer compilation. @Araq I don't understand why this is needed since all code paths return a value. Latest Nim devel would not compile hexer without it though. Semantically, it looks fine to me, but the analyzer seems to treat the default return value as possibly falling through.

## Additional Regression Tests
- `tests/nimony/iter/tfilterclosure.nim`
- `tests/nimony/concepts/tconcepttypevarhook.nim`

Araq:
> typenav.nim: when a callee is still an (ochoice ...), use the first symbol’s routine type instead of (auto).

I don't see how this would be sound.

gmpreussner:
> I don't see how this would be sound.

What would the correct behavior be?

Are my unit tests OK? I want to make sure I'm not solving the wrong problems.

I'm seeing a couple other crashes when combining generics and iterators/closures. It all seems a bit fragile at the moment. I'm gonna explore these a bit further. At the very least I'd like to get some useful regression tests out of it.

Araq:
> What would the correct behavior be?

Well typenav shouldn't guess the type for a choice, returning `auto` (for don't-know) was fine. The caller shouldn't burden typenav with symchoices if it needs a real type.

> Are my unit tests OK? I want to make sure I'm not solving the wrong problems.

They all look good to me.

Araq:
Same issue, why is derefs.nim fed sym-choices? It's sem's job to resolve them.

gmpreussner:
I'm gonna need a bit more constructive guidance for a proper fix. I understand from your comment that derefs should not receive any symbol choices - those should already be resolved by sem. I suspect that the call should be resolved when the match is concept-bound instead of preserving the sym choice, but I'm not clear on where exactly that should happen.

gmpreussner:
Superseded by https://github.com/nim-lang/nimony/pull/2178

PR #2157, opened 2026-07-17
fixes #2156; creates `nimcache_static` in the current user cache dir

### LINKED ISSUE #2156: nimony crashes when building a project if the stdlib is in a read only folder
This line sets the cache_static to point to a folder depending on the stdlib, which makes compilation fail when that folder is read-only. 

https://github.com/nim-lang/nimony/blob/d228d3d3262746db798b49899666f355c59943c6/src/nimony/deps.nim#L658

demotomohiro:
fixes https://github.com/nim-lang/nimony/issues/2156


demotomohiro:
It seems ci failed on windows because https://github.com/nim-lang/nimony/issues/2159

PR #2165, opened 2026-07-20
Fixed bootstrap log spam due to POSIX/glibc signature mismatch for execve

### LINKED ISSUE #2148: hastur: conflicting types for built-in function `execve`
When running `hastur boot` or `hastur all` in `master` on Linux, I recently started seeing multiple warnings of the following form:

```
warning: conflicting types for built-in function ‘execve’; expected ‘int(const char *, char * const*, char * const*)’ [-Wbuiltin-declaration-mismatch]
  516 | int execve(NC8* path_5, char** argv_0, char** env_0);
      |     ^~~~~~
nimcache/boot/nimsem/nim849cil1/oswd7dmm.c:582:12: warning: conflicting types for built-in function ‘execve’; expected ‘int(const char *, char * const*, char * const*)’ [-Wbuiltin-declaration-mismatch]
  582 | extern int execve(NC8* path_5, char** argv_0, char** env_0);
      |            ^~~~~~
nimcache/boot/nimsem/nim849cil1/ospa68oph1.c:767:12: warning: conflicting types for built-in function ‘execve’; expected ‘int(const char *, char * const*, char * const*)’ [-Wbuiltin-declaration-mismatch]
  767 | extern int execve(NC8* path_5, char** argv_0, char** env_0);
      |            ^~~~~~
```

gmpreussner:
Fixes https://github.com/nim-lang/nimony/issues/2148


lib/std/posix/posix.nim
```diff
@@ -451,27 +451,29 @@ when defined(posix):
...
+  # declaration matches glibc's builtin (`const char*`, `char *const*`).
   type CChar* {.importc: "char", nodecl.} = int8
   type CCharArray* = nil ptr UncheckedArray[nil ptr CChar]
 
   when defined(nimNativeIo):
+    {.passC: "-Ilib/std/posix".}
     proc pipe*(a: ptr cint): cint {.importc: "pipe", sideEffect.}
     proc dup2*(oldfd, newfd: cint): cint {.importc: "dup2", sideEffect.}
     proc fork*(): Pid {.importc: "fork", sideEffect.}
-    proc execve*(path: cstring; argv, env: CCharArray): cint {.importc: "execve", sideEffect.}
```
Araq, line 464: Just mark it as `nodecl` instead.

gmpreussner: That breaks `hastur boot` with:

```
error: implicit declaration of function ‘execve’ [-Wimplicit-function-declaration]
  584 |     int X60Qx_11 = execve(file_0, argv_1, envp_0);
      |      
```

gmpreussner:
Some more details:

Under `-d:nimNativeIo` (the default for `hastur boot`), `execve` was bound with a bare `{.importc.}` and no header. Leng then emitted:

```c
extern int execve(NC8* path_5, char** argv_0, char** env_0);
```
GCC 14+ compares that against its built-in execve prototype:
```c
int execve(const char *, char * const *, char * const *);
```
That mismatch triggers `-Wbuiltin-declaration-mismatch`. Every module that imported `execve` (`posix`, `os`, `osproc`, …) got its own conflicting `extern` declaration, so the warning appeared multiple times.

Araq:
> GCC 14+ compares that against its built-in execve prototype

That is not GCC's job unless the prototype also sneaked into by an include file.

gmpreussner:
I had another look because this is indeed weird.

This is actually a Nimony binding problem, not a Leng codegen bug; GCC also does what it's supposed to do. In GCC 14 and newer, there is stricter function signature checking. It specifically knows about `execve` and other functions.

It comes down to a `char**` versus `char* const*` mismatch. If I change the function signature it works without side-stepping the prototype derivation from Nimony types.

I will push an update for your review.

Araq:
>  It specifically knows about execve and other functions.

If true, that is terrible and one more reason to leave C compilers behind as they don't understand ABIs anymore... :-/

Araq:
__

PR #2178, opened 2026-07-23
Propagating generic type constraints to nested procs and closures

### LINKED ISSUE #2143: Proc calls inside `do`-block closures crash compiler when enclosing generic is constrained by concept

### LINKED ISSUE #2175: Concept constraints are not propagated into closures, nested procs, and do-calls
## Classic Nim
Concept bounds propagate indirectly, through the outer generic context:

1. **The outer type parameter stays in scope.** `T` in a nested proc or `do`-closure is the same generic symbol as in the outer signature, not a fresh unconstrained type.
2. **Generic bodies use deferred checking.** Inside a generic proc (including nested procs/closures semchecked as part of that body), unresolved calls are not required to resolve immediately. If overload resolution fails while `inGenericContext > 0`, the compiler keeps the call as a generic AST node and re-checks it at instantiation time.
3. **Concept dispatch runs at instantiation.** When you call a generic function with a specific type, `T` becomes that type and concept matching finds the required overload for that concrete type.

There is an official test for concept information surviving into closures (_tests/concepts/tconceptinclosure.nim_) though it focuses on field access and concept type variables rather than operator dispatch.

## What is not propagated
A nested proc cannot freely introduce its own generic parameters with concept bounds unless types can be inferred from context (see `errGenericLambdaNotAllowed` in the compiler). The usual pattern is: reuse the outer `T`, don't redeclare it.

Also, classic Nim does not literally re-annotate inner parameters as v: T: SomeConstraint. The bound lives on the outer generic param symbol; inner code just uses `T`.

## Nimony (failure case)
Nimony is different. In closures/nested procs it loses the concept constraint for overload resolution, so generic proc calls with specific types fall through on an effectively unconstrained `T` and may fail to compile.

### LINKED ISSUE #2176: Passing non-closure routines to `.closure` parameters hits compiler bug
During generic instantiation, sigmatch wraps non-closure routines passed to `{.closure.}` parameters in `(toClosure ...)`. That node is then reaching `semExpr` with no handler, triggering:

```
[Bug] frontend should not encounter `toClosure`
```

### LINKED ISSUE #2177: Usage of closures inside iterators can trigger compiler bug
`treProcType` treats only `ProctypeT` as the compact 4-field layout (`nilability, params, retType, pragmas`).

`ItertypeT` uses the same layout but is processed with the 8-field routine path (`BodyPos`), walking past the end of the tree and hitting:

```
[Bug] unexpected ')' inside
```

## Example
```nim
func map*[A, T](iter: iterator(): T, f: proc(value: T): A {.closure.}): iterator(): A =
  result = iterator(): A =
    for value in iter():
      yield f(value)
```

gmpreussner:
Please see the attached unit tests for covered use cases.

Fixes https://github.com/nim-lang/nimony/issues/2175
Fixes https://github.com/nim-lang/nimony/issues/2176
Fixes https://github.com/nim-lang/nimony/issues/2177
Fixes https://github.com/nim-lang/nimony/issues/2143


src/nimony/semtypes.nim
```diff
@@ -936,12 +936,19 @@ proc semLocalTypeImpl*(c: var SemContext; dest: var TokenBuf; n: var Cursor;
...
+      if resolved != s.name:
+        dest.shrink start
+        dest.add symToken(resolved, info)
+        s = fetchSym(c, resolved)
     semTypeSym c, dest, s, info, start, context
   of Symbol:
     let start = dest.len
-    let s = fetchSym(c, n.symId)
-    dest.add n
+    let symId = resolveNestedTypevar(c, n.symId)
```
Araq: Well it's a Symbol, it was already looked up... We don't want to patch it again!

src/nimony/semtypes.nim
```diff
@@ -936,12 +936,19 @@ proc semLocalTypeImpl*(c: var SemContext; dest: var TokenBuf; n: var Cursor;
   case n.kind
   of Ident:
     let start = dest.len
-    let s = semIdent(c, dest, n, {})
+    var s = semIdent(c, dest, n, {})
+    if s.kind == TypevarY:
```
Araq: Silly special cases are to be avoided. What is it with the scoping rules that make this necessary?

src/nimony/semdecls.nim
```diff
@@ -1008,6 +1031,8 @@ proc semProcImpl(c: var SemContext; dest: var TokenBuf; it: var Item; kind: SymK
       buildErr c, dest, it.n.info, "TR pattern not implemented"
       skip it.n
     c.routine = createSemRoutine(kind, c.routine)
+    if c.routine.parent != nil and c.routine.parent.genericParams.len > 0:
+      c.routine.genericParams = c.routine.parent.genericParams
```
Araq: Yeah... I'm not sure about that. Get the symbol lookups right instead.

src/nimony/semdecls.nim
```diff
@@ -544,6 +551,22 @@ proc semGenericParams(c: var SemContext; dest: var TokenBuf; n: var Cursor) =
...
+    let origRes = tryLoadSym(origin)
+    if origRes.status == LacksNothing and isRoutine(origRes.decl.symKind):
+      var tv = asRoutine(origRes.decl).typevars
+      if tv.substructureKind == TypevarsU:
+        tv = sub(tv)
+        while tv.hasMore:
+          if isTypevarLike(tv.symKind):
+            var tvs = tv
+            inc tvs
+            c.routine.genericParams.add tvs.symId
```
Araq: Why?

Araq:
You effectively added a secondary scoping layer. But we already have one.

src/nimony/semdata.nim
```diff
@@ -30,6 +30,16 @@ type
 proc createSemRoutine*(kind: SymKind; parent: SemRoutine): SemRoutine =
   result = SemRoutine(kind: kind, parent: parent, resId: SymId(0))
 
+proc inGenericDefinitionContext*(r: SemRoutine): bool =
```
Araq: Instead keep a counter for this in SemContext.

src/nimony/sigmatch.nim
```diff
@@ -1261,7 +1262,7 @@ proc procTypeMatch(m: var Match; f, a: var Cursor) =
     m.error ClosureMismatch, f, a
   elif fcc.usesPassive != acc.usesPassive:
     m.error PassiveMismatch, f, a
-  if not m.err and fcc.usesClosure and not acc.usesClosure:
+  if not m.err and fcc.usesClosure and not acc.usesClosure and m.argExpr.exprKind != ToClosureX:
```
Araq: Meh, instead flatten it away in sema later.

src/nimony/sem.nim
```diff
@@ -512,20 +536,31 @@ proc instantiateExprIntoBuf(c: var SemContext; buf: var TokenBuf; it: var Item;
   it.n = cursorAt(buf, start)
 
 proc fetchSym*(c: var SemContext; s: SymId): Sym =
-  # yyy find a better solution
+  ## Resolve `s` to a scoped or imported symbol. Nifler assigns distinct
+  ## symIds per occurrence of the same generic parameter name; when the exact
```
Araq: But Nifler doesn't do anything like that, it doesn't create syms, sem does!

src/nimony/sem.nim
```diff
@@ -344,6 +344,26 @@ type
     params: ptr Table[SymId, Cursor]
     instSuffix: string
 
+proc typevarBasename(s: SymId): StrId =
+  var name = pool.syms[s]
+  extractBasename(name)
+  result = pool.strings.getOrIncl(name)
+
+proc findOuterTypevarInSubs(s: SymId; sc: SubsContext): SymId =
```
Araq: Still the wrong idea afaict. There is a scope stack used for symbol lookups. More shouldn't be necessary to look up outer generic parameters.

gmpreussner:
I dont know if this getting better or worse; may have to leave this one up to you.

gmpreussner:
Superseded by https://github.com/nim-lang/nimony/pull/2186

PR #2179, opened 2026-07-23
fixes #2151; recompile modules when used plugins are changed

### LINKED ISSUE #2151: Nimony doesn't recompile a file when only the module plugin changed
With a file that uses a module plugin, when modifying the plugin and recompiling, nimony doesn't recompile ot and just throw the old result.

The only way to make it recompile is to delete nimcache which retrigger the whole compilation process making iteration a bit painful.

demotomohiro:
Fixes https://github.com/nim-lang/nimony/issues/2151
Nifler scans `plugin` pragmas and generates a list of used plugins in `*.p.deps.nif` file.
Nimony read the files and generates `*.build.nif` file so that modules uses plugins are recompiled when the plugins code are changed.

### DIFF
--- a/doc/tags.md
+++ b/doc/tags.md
@@ -221,7 +221,7 @@
 | `(borrow)` | NimonyPragma | `borrow` pragma |
 | `(noSideEffect)` | NimonyPragma | `noSideEffect` pragma |
 | `(nodestroy)` | NimonyPragma | `nodestroy` pragma |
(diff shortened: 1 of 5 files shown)


Araq:
Why does this have to be Nifler's job?

demotomohiro:
Created a new PR that doesn't use Nifler: https://github.com/nim-lang/nimony/pull/2226

PR #2182, opened 2026-07-24
Fix lazy template promotion binding wrong homonymous typevars.

### LINKED ISSUE #2181: Templates are order-dependent in certain combinations of constrained parameters
This is an edge case bug. It is only triggered with the exact combination of conditions listed below:

```nim
# Minimal nimony repro for bogus "got: T but wanted: T" on concept-bound templates.
#
# Trigger conditions:
#   1. Two templates before `two` reference `T.two`
#   2. They use different concept constraints on the same generic T
#   3. The first constraint must be a plain subconcept; the second must add ops
#   4. `two` is defined with `T.identity + T.identity` (typedesc field call)

type
  Add = concept
    func `+`(a, b: Self): Self
    func identity(_: typedesc[Self]): Self

  Plain = concept of Add

  Extra = concept of Add
    func `*`(a, b: Self): Self

template x*[T: Plain](_: typedesc[T]): T = T.two     # Error: type mismatch: got: T but wanted: T
template y*[T: Extra](_: typedesc[T]): T = T.two
template two*[T: Add](_: typedesc[T]): T = T.identity + T.identity
```

This error can be avoided by reordering the templates:
```nim
template two*[T: Add](_: typedesc[T]): T = T.identity + T.identity
template x*[T: Plain](_: typedesc[T]): T = T.two     # OK
template y*[T: Extra](_: typedesc[T]): T = T.two    # OK
```

gmpreussner:
When a template forward-references another generic template, lazy body promotion must register the callee's typevars on scope before `semTemplBody` runs; otherwise `buildSymChoice` picks the caller's `T` and produces bogus "T vs T" mismatches at expansion sites.

## Root cause
When template `x` forward-referenced `T.two`, nimony lazily promoted template `two`'s body via `semTemplBody`. That pass runs `buildSymChoice` against the live `SemContext` scope, which still contained `x`'s `T`. Because only formal **params** (not **typevars**) were registered on scope during promotion, `two`'s body bound the wrong `T` SymId into the cached AST.

Later, when `y` expanded `two`, the body and return type used different `T` symbols, both printing as `T`, hence:
```
Error: type mismatch: got: T but wanted: T
```

## Fix (`templates.nim`)
Pulled out the `addParamsToScope` block in `tryPromoteTemplateBody` into its own function `addTemplFormalsToScope` to register **both typevars and params** on the parameter scope before `semTemplBody`, matching what phase 3 does via `semGenericParams` / `semParams`.

Promotion now calls it for `typevarsAt` and `paramsAt` in the parameter scope (body scope stays separate).

Fixes https://github.com/nim-lang/nimony/issues/2181

### DIFF
--- a/src/nimony/templates.nim
+++ b/src/nimony/templates.nim
@@ -132,6 +132,24 @@ proc expandPlugin(c: var SemContext; dest: var TokenBuf; temp: Routine, args: Cu
       else:
         skip p
 
+proc addTemplFormalsToScope(c: var SemContext; buf: TokenBuf; at: int) =
+  ## Register published generic/type/formal params on `c.currentScope` so
+  ## `semTemplBody`'s `buildSymChoice` resolves to this template's symbols,
+  ## not homonymous typevars from an enclosing caller during lazy promotion.
+  var p = readonlyCursorAt(buf, at)
+  if p.substructureKind in {ParamsU, TypevarsU}:
+    p.into:
+      while p.hasMore:
+        let param = asLocal(p)
+        if param.name.kind == SymbolDef:
+          var nameStr = pool.syms[param.name.symId]
+          extractBasename(nameStr)
+          if nameStr.len > 0:
+            let s = Sym(kind: param.kind, name: param.name.symId, pos: 0)
+            addOverloadable(c.currentScope,
+                            pool.strings.getOrIncl(nameStr), s)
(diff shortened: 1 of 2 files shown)


Araq:
Now part of https://github.com/nim-lang/nimony/pull/2185

PR #2183, opened 2026-07-24
Io uring backend for ioring

blackmius:
restructured ioring - moved backend strategy out of threadpool and ioring, replaced fdslots with arena slots


lib/std/parfor.nim
```diff
@@ -56,14 +56,16 @@ var poolState: int
...
+
 proc ensureParPool*() =
   ## Start the worker pool the first time a parallel `for` runs. Idempotent and
   ## thread-safe: concurrent first-callers race through a CAS, the loser spins
   ## until the winner has finished `initPool`.
   if atomicLoad(poolState, moAcquire) == 2: return
   var expected = 0
   if atomicCompareExchange(poolState, expected, 1):
-    initPool()
+    gPool = createPool()
```
blackmius: looks it must be one global pool, but seems worker loops should learn how to add continuous work maybe something like idle functionality in uvloop

lib/std/ioring/core/backends.nim
```diff
@@ -0,0 +1,83 @@
...
+# processFd() to dispatch I/O for all slots on a given fd.
+# Subclasses override registerEvent/reArmEvent/fetchEvents.
+
+import ./types
+import ./slots
+
+type
+  PollBackend* = ref object of Backend
+    pollFd*: cint
+    arena*: int
```
blackmius: deepseek does not know how to use pointers

lib/std/ioring/core/types.nim
```diff
@@ -0,0 +1,39 @@
...
+    buf*: nil pointer
+    len*: int
+    cont*: Continuation
+    res*: int
+    acceptAddr*: Sockaddr_storage
+    acceptLen*: SockLen
+
+  Backend* = ref object of RootObj
+    completeFn*: proc(slotIdx: int; res: int; env: int) {.nimcall.}
+    completeEnv*: int
```
blackmius: and here, even more looks like backend should has a backreference to a ring

lib/std/ioring.nim
```diff
@@ -6,30 +6,179 @@
...
+# --- lifecycle ---
+
+proc initIoRing*(): IoRing =
+  new result
+  result.slots.init()
+  result.nextSeq = 1
+
+  when hasIouring:
+    import ./ioring/backends/iouring
+    result.backend = initIoUringBackend(cast[int](addr result.slots))
```
blackmius: i would move it to platform.nim under initBackend function

lib/std/parfor.nim
```diff
@@ -56,14 +56,16 @@ var poolState: int
   ## 0 = uninitialised, 1 = initialising, 2 = ready. Accessed atomically so the
```
blackmius: if it uses global pool this check should be shared between all modules which use threadpool

lib/std/ioring/backends/epoll.nim
```diff
@@ -0,0 +1,90 @@
...
+
+type EpollBackend* = ref object of PollBackend
+
+when defined(linux):
+  {.feature: "lenientnils".}
+
+  import std / assertions
+
+  type
+    EpollData {.importc: "epoll_data_t", header: "<sys/epoll.h>".} = object
```
blackmius: we have epoll already in std/posix/epoll

lib/std/ioring/backends/kqueue.nim
```diff
@@ -0,0 +1,89 @@
...
+import ../core/slots
+import ../core/backends
+
+type KqueueBackend* = ref object of PollBackend
+
+when defined(macosx) or defined(freebsd) or defined(netbsd) or defined(openbsd) or defined(dragonfly):
+  import std / assertions
+
+  type
+    Kevent {.importc: "struct kevent", header: "<sys/event.h>".} = object
```
blackmius: move to std/posix/kqueue

lib/std/ioring/backends/kqueue.nim
```diff
@@ -0,0 +1,89 @@
...
+# registerEvent/reArmEvent use kevent with EV_ONESHOT.
+# poll drives kevent → processFd per event.
+
+import ../core/types
+import ../core/slots
+import ../core/backends
+
+type KqueueBackend* = ref object of PollBackend
+
+when defined(macosx) or defined(freebsd) or defined(netbsd) or defined(openbsd) or defined(dragonfly):
```
blackmius: obsolete when check backend is only imported in platform.nim whoch already checked the platform

lib/std/ioring/core/backends.nim
```diff
@@ -0,0 +1,83 @@
...
+  proc posixWrite(fd: cint; buf: nil pointer; count: int): int {.importc: "write", header: "<unistd.h>".}
+  proc posixAccept(s: cint; `addr`: ptr SockAddr; addrlen: ptr SockLen): cint {.
+    importc: "accept", header: "<sys/socket.h>".}
+
+  proc processFd*(b: PollBackend; fd: cint; firedEvents: int) =
+    let a = cast[ptr SlotArena](b.arena)
+    # collect all pending slot indices for this fd
+    var pending = newSeq[int]()
+    for j in 0..<MaxOps:
+      if a.slots[j].inUse and a.slots[j].fd == fd:
```
blackmius: meh havent think it will traverse all slots for parallel operations, maybe pollBackends should still use fdslots but now it will be a Table[fd, seq[slotIdx]] for a faster lookup, or limit operations to one per fd per type write/read in these backends,  not sure need to double check how asyncdispatch handles it

lib/std/threadpool.nim
```diff
@@ -364,18 +151,21 @@ proc initPool*() =
       # workers read garbage thread indices (valgrind: uninitialised value in
       # tryBulkDequeue/workerLoop, origin the reused frame). No storage, no
       # lifetime, no race.
-      create workers[i], workerLoop, cast[pointer](i)
+      var arg: tuple[p: Pool, threadIdx: int] = (p, i)
```
Redict: maybe it's better to store args in something like `args: array[WorkerCount, tuple[...]]`

lib/std/ioring/backends/kqueue.nim
```diff
@@ -0,0 +1,89 @@
...
+              eventlist: ptr Kevent; nevents: cint; timeout: ptr Timespec): cint {.
+    importc, header: "<sys/event.h>".}
+
+  proc close_fd(fd: cint): cint {.importc: "close", header: "<unistd.h>".}
+
+  const
+    EVFILT_READ = -1.cshort
+    EVFILT_WRITE = -2.cshort
+    EV_ADD = 0x0001.cushort
+    EV_ONESHOT = 0x0020.cushort
```
Redict: previously was `EV_ONESHOT* = cushort(0x0010)`. what changed here?

lib/std/threadpool.nim
blackmius: the idea is that pool is not global anymore thus must hold pointer to pool instance + thread_id

lib/std/ioring/backends/kqueue.nim
```diff
@@ -0,0 +1,89 @@
...
+import ../core/backends
+
+type KqueueBackend* = ref object of PollBackend
+
+when defined(macosx) or defined(freebsd) or defined(netbsd) or defined(openbsd) or defined(dragonfly):
+  import std / assertions
+
+  type
+    Kevent {.importc: "struct kevent", header: "<sys/event.h>".} = object
+      ident: uintptr
```
Redict: i think it's `csize_t`, not `uintptr`

lib/std/ioring/backends/iouring.nim
```diff
@@ -0,0 +1,82 @@
...
+  try:
+    discard localQueue.submit()
+  except:
+    discard
+  let waitNr = if timeoutMs > 0: 1'u else: 0'u
+  const batchSize = 64
+  var cqes = newSeq[Cqe](batchSize)
+  var n: int = 0
+  try:
+    n = localQueue.copyCqes(cqes, waitNr)
```
Redict: with `waitNr=1` it will enter kernel with `min_complete=1` and no timeout, so it will hang in join (idle worker will sit in `io_uring_enter` forever). need a test for that i guess.

lib/std/threadpool.nim
```diff
@@ -129,34 +38,34 @@ type
...
-  gIoFd: cint
-  stopFlag: bool  # accessed atomically
+method poll*(p: Pool; timeoutMs: int): bool {.base.} =
+  ## Poll the shared I/O instance once and dispatch every ready handler on the
+  ## calling thread. `timeoutMs` is how long to block waiting for an event (`0`
+  ## = non-blocking peek). Returns true if at least one handler fired. Safe to
+  ## call from any thread (oneshot semantics: each event goes to exactly one
+  ## caller) — used both by the worker loop and by `parWait` so a thread blocked
+  ## on a join can still advance the I/O its parked chunks are waiting on.
+  result = false
```
Redict, line 46: this will busy-spin all idle workers at 100% CPU, because we ignore timeout

lib/std/ioring/core/backends.nim
```diff
@@ -0,0 +1,83 @@
...
+  proc posixRead(fd: cint; buf: nil pointer; count: int): int {.importc: "read", header: "<unistd.h>".}
+  proc posixWrite(fd: cint; buf: nil pointer; count: int): int {.importc: "write", header: "<unistd.h>".}
+  proc posixAccept(s: cint; `addr`: ptr SockAddr; addrlen: ptr SockLen): cint {.
+    importc: "accept", header: "<sys/socket.h>".}
+
+  proc processFd*(b: PollBackend; fd: cint; firedEvents: int) =
+    let a = cast[ptr SlotArena](b.arena)
+    # collect all pending slot indices for this fd
+    var pending = newSeq[int]()
+    for j in 0..<MaxOps:
```
Redict: 2*O(MaxOps) scans per event? previous implementation was O(1) fd-indexed lookup. regression?

lib/std/threadpool.nim
Redict:
```
--- a/lib/std/threadpool.nim
+++ b/lib/std/threadpool.nim
@@
+when not defined(windows):
+  proc usleepMicroseconds(usec: cuint): cint {.importc: "usleep", header: "<unistd.h>".}
+
 method poll*(p: Pool; timeoutMs: int): bool {.base.} =
   ## Poll the shared I/O instance once and dispatch every ready handler on the
   ## calling thread. `timeoutMs` is how long to block waiting for an event (`0`
   ## = non-blocking peek). Returns true if at least one handler fired. Safe to
   ## call from any thread (oneshot semantics: each event goes to exactly one
   ## caller) — used both by the worker loop and by `parWait` so a thread blocked
   ## on a join can still advance the I/O its parked chunks are waiting on.
   result = false
+  if timeoutMs > 0:
+    when defined(windows):
+      sleep(timeoutMs.uint32)
+    else:
+      discard usleepMicroseconds(cuint(timeoutMs) * 1000'u32)
```

lib/std/ioring/backends/iouring.nim
blackmius: this is fine, if there timeout was passed that means no work to do and we can wait, the poll backends should retry calls until event returned, io_uring just blocks until at least one submission is done

lib/std/threadpool.nim
blackmius: indeed

lib/std/ioring/backends/iouring.nim
blackmius: actually same idea can be applies for epoll_wait passing there -1 for timeout if we have nothing to do in the queue

lib/std/ioring/backends/epoll.nim
```diff
@@ -0,0 +1,58 @@
...
+
+method reArmEvent*(b: EpollBackend; fd: cint; slotIdx: int; mask: int) =
+  var ev {.noinit.}: EpollEvent
+  ev.events = EPOLLONESHOT
+  if (mask and EvRead) != 0:
+    ev.events = ev.events or EPOLLIN
+  if (mask and EvWrite) != 0:
+    ev.events = ev.events or EPOLLOUT
+  ev.data.`ptr` = cast[pointer](uint(slotIdx))
+  discard epoll_ctl(b.pollFd, EPOLL_CTL_MOD, fd, addr ev)
```
Redict: did this one compile after switching to `std/posix/epoll`? i get:

```
lib/std/ioring/backends/epoll.nim(39, 20) Error: Type mismatch at [position]
epoll_ctl(b.pollFd, EPOLL_CTL_MOD, fd, addrev)
[2] expected: 32 but got: 64 (declared in lib/std/posix/epoll.nim(68, 1))
```

looks like `std/posix/epoll` declares the opcodes as plain ints (`EPOLL_CTL_ADD* = 1`) while its `epoll_ctl` takes `op: cint`? same thing for `EPOLLIN`/`EPOLLONESHOT` against `ev.events: uint32` above. did those need `.cint`/`.uint32`?

lib/std/ioring/core/backends.nim
```diff
@@ -0,0 +1,83 @@
...
+    mask = mask or EvWrite
+  b.registerEvent(op.fd, slotIdx, mask)
+
+when defined(posix):
+  import std / assertions
+  from std/posix/posix import SockAddr, SockLen
+
+  proc posixRead(fd: cint; buf: nil pointer; count: int): int {.importc: "read", header: "<unistd.h>".}
+  proc posixWrite(fd: cint; buf: nil pointer; count: int): int {.importc: "write", header: "<unistd.h>".}
+  proc posixAccept(s: cint; `addr`: ptr SockAddr; addrlen: ptr SockLen): cint {.
```
Redict: after fixing the `epoll_ctl` type errors above i hit a second one here, at the C level:

```
error: incompatible type for argument 2 of 'accept'
note: expected '__SOCKADDR_ARG' but argument is of type 'SockAddr_0_... *'
```

nimony compiles with `-d:nimNativeIo`, and in that branch `posix_other.nim` defines `SockAddr` as a plain nim object rather than `importc: "struct sockaddr"` — so it's not libc's struct anymore. on master `ioring.nim` declared `SockAddr`/`Sockaddr_storage` locally *with* `importc`, which is why it worked there.

is `std/posix/posix` the right source for these here, or should the backend keep its own `importc`ed decls?

lib/std/ioring/backends/epoll.nim
```diff
@@ -0,0 +1,58 @@
...
+
+method registerEvent*(b: EpollBackend; fd: cint; slotIdx: int; mask: int) =
+  var ev {.noinit.}: EpollEvent
+  ev.events = EPOLLONESHOT
+  if (mask and EvRead) != 0:
+    ev.events = ev.events or EPOLLIN
+  if (mask and EvWrite) != 0:
+    ev.events = ev.events or EPOLLOUT
+  ev.data.`ptr` = cast[pointer](uint(slotIdx))
+  discard epoll_ctl(b.pollFd, EPOLL_CTL_ADD, fd, addr ev)
```
Redict: `registerEvent` always uses `EPOLL_CTL_ADD`, but with `EPOLLONESHOT` the fd stays registered after it fires (just disarmed), so wouldn't the second `submit` on the same fd get `EEXIST` here — which is discarded — and never re-arm?

master tracked this with a `registered: bool` per fd and used ADD vs MOD accordingly. is that flag coming back somewhere, or is the plan that `submit` goes through `reArmEvent` when the fd is already known?

lib/std/ioring/core/slots.nim
```diff
@@ -0,0 +1,26 @@
...
+  slots*: array[MaxOps, OpContext]
+  freelist*: seq[int]
+
+proc init*(a: var SlotArena) =
+  a.freelist = newSeq[int]()
+  for i in 0..<MaxOps:
+    a.freelist.add(i)
+
+proc allocSlot*(a: var SlotArena): int =
+  result = a.freelist.pop()
```
Redict: is `freelist` meant to be touched from more than one thread? `allocSlot` runs on whoever calls `submitRead`/`submitWrite`, and `freeSlot` runs from `completeCb` on a worker — that's a plain `seq` being `pop`/`add`ed concurrently.

also `pop` is `requires: (s.len > 0)`, so what should happen when all 8192 slots are in use? right now it looks like nothing checks it.

lib/std/ioring/core/backends.nim
```diff
@@ -0,0 +1,83 @@
...
+    importc: "accept", header: "<sys/socket.h>".}
+
+  proc processFd*(b: PollBackend; fd: cint; firedEvents: int) =
+    let a = cast[ptr SlotArena](b.arena)
```
Redict: `firedEvents` is taken but never read — so a write-readiness wakeup will also run a pending `opRead` on that fd, no? master masked with `EvRead`/`EvWrite` before doing the read/write. intentional for now, or leftover?

lib/std/ioring.nim
```diff
@@ -38,17 +187,19 @@ when defined(posix):
   else:
     const O_NONBLOCK* = 0x0004.cint
 
+  proc setNonBlocking*(fd: cint) =
+    var flags = fcntl(fd, F_GETFL)
+    discard fcntl(fd, F_SETFL, flags or O_NONBLOCK)
+
+  proc closeFd*(fd: cint) =
```
Redict, line 132: master's `closeFd` also unregistered the fd from the poller and cleared the slot; this one only closes. if an op is still pending on that fd, its slot stays `inUse` forever — is that leak intended to be handled elsewhere, or should this walk the arena too?

lib/std/posix/io_uring.nim
```diff
@@ -532,26 +532,22 @@ proc newRing(fd: FileHandle; offset: ptr SqringOffsets; size: uint32): SqRing {.
   #   raise ERROR_io_uring_not_initializes
 
 # XXX: is it ok that =destroy can raise error?
-proc `=destroy`(queue: Queue) {.raises, tags: [].} =
+proc `=destroy`(queue: Queue) {.tags: [].} =
```
Redict: the `=sink`/`=copy` hooks got dropped here — `=copy` was `{.error: "Queue can has only one owner".}`. what makes `Queue` safe to copy now? `=destroy` still closes the fd and unmaps the rings, so a copy would do that twice.

was that removal needed for something specific, or did it just come along with the `raises` change?

Redict (commented):
built the branch at `d082062f` locally and tried to actually run the ioring paths. worker-arg fix in `bce3b3e4` looks good — verified it, no more crash.

main thing: i can't get the epoll backend to compile at all now. a file whose entire content is `import std/ioring/backends/epoll` fails, first on `epoll_ctl` types, then (after casting past that) on `accept` at the C level. same file against master's `std/ioring` compiles and runs fine, so it seems to be new. details in the two inline comments.

is `std/ioring` built by anything in the test suite? if nothing imports it, breakage like this wouldn't show up in CI — maybe worth a compile-only smoke test.

rest are questions on things i wasn't sure were intentional.

lib/std/threadpool.nim
Redict: still present at `d082062f` fwiw — measured on the built branch, pool created and left completely idle for 500ms wall:

```
User time (seconds): 4.06
Percent of CPU this job got: 736%
Elapsed (wall clock) time: 0:00.55
Voluntary context switches: 9
```

9 voluntary ctx switches over the whole run, so the workers never sleep. with the `usleep` patch above it's 8% CPU at the same wall time.

note this hits `parfor`'s `gPool` too, since that's a plain `createPool()`.

lib/std/ioring/backends/iouring.nim
Redict: makes sense for the I/O side. but what wakes the worker for shutdown? `workerLoop` only re-checks `stopFlag` after `poll` returns, so if it's parked in `io_uring_enter` with no timeout and no ops are pending, wouldn't `pool.shutdown()` block in `join()`?

same question for `epoll_wait(-1)` if that gets applied. is there something planned to nudge the workers (eventfd / self-pipe), or should the timeout just stay bounded?

lib/std/threadpool.nim
Redict: the heap-alloc + `dealloc` in the worker works, confirmed on the built branch. before `bce3b3e4` a pool created inside a helper proc segfaulted every run; now it's fine. thanks.

lib/std/ioring/core/backends.nim
blackmius: I believe `std/posix/posix` should be one source of truth

lib/std/ioring/backends/epoll.nim
blackmius: try 70abfab

lib/std/posix/io_uring.nim
blackmius: it is not ok that `=destroy` can raise error, we can't control when it adds, we can't avoid error so it must be safe

lib/std/ioring/backends/epoll.nim
Redict: yep, `70abfab` fixes it — rebuilt, zero nim-level errors now. the `'u32`/`'i32` suffixes on the consts did it. sorry for the noise, my comment was against `d082062f`.

kqueue `EV_ONESHOT`/`ident` in the new `std/posix/kqueue` look right too.

lib/std/ioring/core/backends.nim
Redict: agree on one source of truth. but this one still fails on `70abfab` (rebuilt just now — the nim errors are gone, this is the only one left):

```
bac7wdgch.c:631:46: error: incompatible type for argument 2 of 'accept'
  int X60Qii_9 = accept(fd_4, ((SockAddr_0_pososrh1q1*)(&(*X60Qii_5).acceptAddr_0)), ...
note: expected '__SOCKADDR_ARG' but argument is of type 'SockAddr_0_pososrh1q1 *'
```

the issue is which branch of `posix_other.nim` we land in: nimony builds with `-d:nimNativeIo`, and there `SockAddr`/`Sockaddr_storage` are plain nim objects, so passing them to libc's `accept` doesn't typecheck. the `else` branch has the `importc`ed ones.

should `Sockaddr_storage` in the `nimNativeIo` branch get a real layout (it's currently an empty object, so `sizeof` is 0 too)? or should the poll backends not call libc `accept` in that config?

lib/std/posix/io_uring.nim
Redict: yeah, no argument on making `=destroy` not raise — that part makes sense.

i meant the two hooks below it that went away in the same diff: `=sink` (copyMem, "avoid unmapping uring object after moving") and `=copy` (`{.error: "Queue can has only one owner".}`). those were separate from the raises change — were they dropped on purpose, or just collateral?

Redict:
correction to my review above — it was written against `d082062f`, `70abfab` landed while i was posting.

rechecked on `70abfab`:
- epoll_ctl / EPOLLIN type errors — **fixed**, no nim-level errors left
- kqueue `EV_ONESHOT` (`0x0010`) and `ident` — **fixed** in the new `std/posix/kqueue`
- `accept` / `SockAddr` under `-d:nimNativeIo` — still there, it's the only thing blocking `import std/ioring/backends/epoll` now
- idle busy-spin in base `Pool.poll` — still there

the other comments (ADD vs MOD re-arm, slot freelist, `firedEvents`, `closeFd`) are unaffected by `70abfab`, still open as questions.

lib/std/ioring/core/slots.nim
blackmius: well, slots need lock, ill research for lockless queue here

lib/std/posix/io_uring.nim
blackmius: it turned out without =copy it was not possible to initialise queue and, surprisingly enough, assign it to pool ring. I didn't get the semantic used there and leave it as it, because it worked

Redict: the semantic those two hooks encoded: `=copy` deleted → `Queue` is move-only, and `=sink` copyMem'd the bits across without the source's `=destroy` unmapping anything. so `localQueue = newQueue(...)` was meant to be a *move* (rhs is a temporary), which is the case they were written for.

if that assignment only compiles once `=copy` is restored-as-default, that suggests the move isn't being detected and it silently became a copy — which is the thing `=copy {.error.}` existed to catch. with it gone, both the temporary and `localQueue` end up running `=destroy` on the same fd + mappings.

do you still have the error it gave? if nimony isn't picking a move for `x = newQueue(...)` that feels like a separate bug worth filing, rather than something to work around by dropping the hook.

(side note: i wanted to write a small double-teardown repro but `import std/posix/io_uring` on its own fails with `undeclared identifier: Sigset` at io_uring.nim:439 — reproduces on master too, so not from this PR, just noting it blocks testing that module standalone.)

blackmius: sigset missing is the same reason as above i think after merge nimony compiles with -s:nimNativeIo which lacks posix types

lib/std/ioring.nim
blackmius: well it must call backend specific close + yes clear it from slots,moreover it should close Operation not file descriptor

lib/std/ioring/backends/iouring.nim
blackmius: good catch, shutdown should close epoll and io_uring then they will return

lib/std/posix/io_uring.nim
Redict: yep, you're right — `Sigset` is declared in the `else:` of `when defined(nimNativeIo):` (posix.nim:460), so it genuinely doesn't exist in the native config. same shape as the `SockAddr` one. looks like it traces back to `1ac3e60c` "make libc-free the default" on master, which the merge picked up.

so `std/posix/posix` covers the libc build but not `nimNativeIo` for these socket/signal types. is filling in the native branch (real `Sockaddr_storage` layout, `Sigset`) the direction you'd want, or does ioring need to go through raw syscalls there instead of libc `accept`?

either way it's a master-side gap rather than something you introduced, so probably fine to not block this PR on it — but ioring can't compile until one of the two happens.

lib/std/ioring/backends/iouring.nim
Redict: closing the fds on shutdown works for waking them, yeah. one thing to watch: the workers call `p.poll(...)` in a loop and only re-check `stopFlag` after it returns, so if a worker is between iterations it can re-enter `poll` on the just-closed fd. setting `stopFlag` before closing (and having `poll` tolerate `EBADF`) should cover it.

happy to re-run my busy-spin/hang checks once that lands.

lib/std/threadpool.nim
blackmius: when again it should sleep only if there is nothing to do

Redict: agreed, and that's already the case at the call site: `workerLoop` passes `0` when `drainOnce` returned work and `1` only when it found nothing (threadpool.nim:145). so `timeoutMs > 0` *is* the "nothing to do" signal, and the sleep in the patch only runs on that path. a busy worker passes 0 and never sleeps.

the issue is the `{.base.}` `poll` itself: it's `result = false` and nothing else, so the 1ms never happens anywhere. subclasses that override it are fine - this only bites the plain `createPool()` path, which is what `parfor`'s `gPool` and `tpool1` use.

still reproduces at `70abfab`, 3 runs, 500ms of wall-clock idle:

```
idle FAIL: pool burned 3770ms CPU during 500ms idle
idle FAIL: pool burned 3773ms CPU during 500ms idle
idle FAIL: pool burned 3775ms CPU during 500ms idle
```

~7.5x wall, ie all 8 workers spinning. `/usr/bin/time -v` on the same run: 736% CPU, 9 voluntary context switches. with the usleep patch: `idle ok`, 8% CPU, same wall time.

usleep was just the minimal demonstration. if you'd rather the base method stay a no-op, the alternative is having `workerLoop` do the idle sleep itself - either way something has to block when `drainOnce` comes back empty.

blackmius: i assume when it uses io it should stay most of the time waiting for io op to complete in blocking syscall epoll_wait/io_uring_enter, when it used as threadpool and have work in queue it should drain it firstly and only then go to sleep

however i can imagine such scenario there waiting on epoll indefinitely long is wrong resource distribution, that only few threads do the work while others wait in syscalls

Redict: (our messages crossed, mine above has the numbers)

no disagreement on the io-backed model: drain first, then block in `epoll_wait`/`io_uring_enter`. that's what the backends do and it's fine.

the gap is the case with no io at all. base `Pool` has only stripes/workers/stopFlag, no poll fd, so there's no syscall to block in - `poll` returns `false` immediately and the loop spins. `createPool()` with pure compute tasks (`parfor`, `tpool1`) never constructs a backend, so it never reaches the blocking path you're describing.

on your second point, that's exactly why the call site passes `1` and not `-1`. a 1ms bounded wait keeps every worker coming back to re-check its stripe and steal, so no one parks indefinitely while work is queued. an indefinite wait would need a wakeup on submit (eventfd or similar) to avoid the imbalance you're describing.

so: backends block on io, base sleeps 1ms. only the second half is missing.

blackmius: nah timer-based poll is not good for p99 latencies, we should add wakeup event for case when we blocked by syscall but work occured

Redict: yeah, that's strictly better - eventfd in the poll set, block indefinitely, submit writes to it. no latency floor and no spinning.

one thing that falls out of it: the base `Pool` needs that eventfd too, otherwise it still has nothing to block on and keeps spinning. so it becomes the wakeup primitive for the no-io case rather than the 1ms sleep, which is a nicer answer than usleep anyway.

happy to re-run the idle measurement once it's in.

lib/std/threadpool.nim
```diff
@@ -170,192 +81,81 @@ proc submit*(t: Task; hint = 0) =
...
+  while not atomicLoad(p.stopFlag, moRelaxed):
     # 1. Bulk-drain tasks: own stripe first, then steal from others. Trampolines
     #    each continuation, re-submitting any that yield more work.
-    let busy = drainOnce(threadIdx)
+    let busy = p.drainOnce(threadIdx)
     # 2. Poll I/O — non-blocking when we just ran work, 1ms wait when idle.
-    discard poolPollIo(if busy: 0.cint else: 1.cint)
-
-# --- Lifecycle ---
+    let eventFired = p.poll(if busy: 0.cint else: 1.cint)
```
blackmius: reconsider timer-based poll for wakeup event

lib/std/threadpool.nim
```diff
@@ -170,192 +81,81 @@ proc submit*(t: Task; hint = 0) =
...
+    if not eventFired and not busy:
+      const timeoutMs = 1
+      when defined(windows):
+        sleep(timeoutMs.uint32)
+      else:
+        var ts = Timespec(tv_sec: (timeoutMs.float32/1000).Time)
```
Redict: `tv_sec` is `distinct clong`, so `(1.float32/1000).Time` truncates to 0 and `tv_nsec` never gets set. printing it gives `tv_sec=0 tv_nsec=0`, ie `nanosleep` returns immediately.

lib/std/threadpool.nim
blackmius: too much work for this pr I also want to experiment with offloading io work to dedicated cpus with io_uring sq poll like scylla db did it https://www.scylladb.com/2026/07/22/asymmetric-io_uring-backend-seastar/

lib/std/ioring/backends/iouring.nim
```diff
@@ -0,0 +1,109 @@
...
+  except:
+    discard
```
Araq: `quit "fatal: bug: copyCqes cannot fail"`

Or maybe something else along those lines.

lib/std/ioring/backends/iouring.nim
```diff
@@ -0,0 +1,109 @@
...
+      discard sqe.write(op.fd, cast[pointer](op.buf), op.len)
+  of opAccept:
+    discard sqe.accept(SocketHandle(op.fd), cast[ptr SockAddr](addr op.acceptAddr), addr op.acceptLen, 0)
+
+method poll*(b: IoUringBackend; timeoutMs: int): bool =
+  b.tryInitLocalQueue()
+  try:
+    discard localQueue.submit()
+  except:
+    discard
```
Araq: `quit "fatal: bug: submit cannot fail"`

Or maybe something else along those lines.

lib/std/ioring/backends/iouring.nim
```diff
@@ -0,0 +1,109 @@
...
+  try:
+    var sqe = localQueue.getSqe()
+    if sqe == nil:
+      discard localQueue.submit()
+      sqe = localQueue.getSqe()
+    if sqe != nil:
+      discard sqe.cancelFd(FileHandle(fd))
+      discard localQueue.submit()
+  except:
+    discard
```
Araq: Hmm, I think this is just something hacked together? :P

lib/std/ioring/backends/iouring.nim
```diff
@@ -0,0 +1,109 @@
...
+method close*(b: IoUringBackend) =
+  # Previously a no-op: the mapped SQ/CQ rings and the io_uring fd were never
+  # released, leaking both per shutdown. Overwriting the threadvar with a
+  # fresh (zero) `Queue` runs `=destroy` on the old value, which unmaps the
+  # rings and closes the fd (see `posix/io_uring.=destroy`).
+  if localQueue.params != nil:
+    try:
+      localQueue = newQueue(b.sqEntries)
+    except:
+      discard
```
Araq: Yeah, pretty sure you messed up with Nimony's .raises virality.

Araq:
This design based on `XBackend* = ref object of PollBackend` is in conflict with our relays-based stdlib design. Maybe yours is better but then I fail to see how.

blackmius:
> This design based on `XBackend* = ref object of PollBackend` is in conflict with our relays-based stdlib design. Maybe yours is better but then I fail to see how.

the difference in perfomance nearly zero because methods use vtables. this is essentially the same as manually setting function pointers inside objects, but machinery is deligated to compiler. so it just looks much cleaner, with no other practical difference, i tried it before in c32e030

Araq:
But it is not about performance. It is about code ceremony and information hiding and simpler APIs.

Araq:
The crux is `proc initEpollBackend*(ring: Ring): EpollBackend`, great now I have an EpollBackend, what am I gonna do with it? Store it in a global anyway. Am I then responsible for locking? Or a threadlocal? That's the stuff we don't want people to decide and it's the stuff that doesn't compose well. It's an event loop, things are better when there is only one of these beasts.

blackmius:
> what am I gonna do with it?

it shouldnt be public it is internals of ioring, as an user you dont have to do with it anything. as an user you have

```nim
ring = initIoRing()
ring.submitRead()
ring.submitEtc()
var comps: array[16, IoCompletion]
let n = ring.waitCompletions(comps)
# or pass continuation to submit
```

actualy you even does need a local ring in std it will be one global ioRing and passive procs that will call global default ioRing
```nim
var gRingState: int = 0
var gRing: Ring
proc initDefaultRing*(): Ring =
  if atomicLoad(gRingState, moAcquire) == 2: return gRing
  var expected = 0
  if atomicCompareExchange(gRingState, expected, 1):
    gRing = initIoRing()
    atomicStore(gRingState, 2, moRelease)
  else:
    while atomicLoad(gRingState, moAcquire) != 2:
      discard
  gRing
```

```nim
import ioring
# init the ring if has not initiated yet
initDefaultRing()

proc read(fd: cint, buf: pointer, size: int): int {.passive.} =
  var result = 0
  var cnt = delay()
  gRing.submitRead(fd, buf, size, cnt, result.addr)
  suspend()
  return result
```

so then user just runs `read` and wont think even about of initializing something


why not just make everything global? first, i am not an AI. i need structured code to understand it. second, having local copies provides testability, allowing us to create unit tests and isolate components from scratch. this makes testing much faster than recompiling the code

hackers who think they can write a backend that outperforms the standard library can simply monkeypatch the global gRing. they can replace it with their own custom backend inherited from the empty `Backend`. relays supply same functionality but via different api which more C grounded. instead of inheriting and overloading a method you assigning function pointer, however you then need to understand what backend was passed to your function and cast somehow to access local fields there are more manual work on every layer

Araq:
I don't understand it. Yes, you are not an AI, exactly, so we want simple things. Why not have an `initEcho()` call before you can use `echo`? Because it sucks, that's why.

blackmius:
> I don't understand it. Yes, you are not an AI, exactly, so we want simple things. Why not have an initEcho() call before you can use echo? Because it sucks, that's why.

i thought that was current approach in master, where you have to call initIoRing. i agree that it is better to hide it inside std libraries. you import and it is already initialised there

Araq:
Aha I see, you merely keep what already exists.

lib/std/ioring/backends/iouring.nim
blackmius: well, it actually can return errors from syscall https://man7.org/linux/man-pages/man2/io_uring_enter.2.html maybe for now it is easier to crash

lib/std/ioring/backends/iouring.nim
blackmius: dont get it, did you mean that there are too many try/except blocks? newQueue tries to call mmap and io_uring_setup syscalls which can fail, so it is a first what comes to mind when you dealing with somehting that can throw an error

blackmius:
there is a problem with io_uring that it have a threadlocal queue and we cant run a passive proc from a main thread becuase it will submit operation to main thread queue which nothing polls, we eiter need to run thread poll at the end, or run continutation from worker so it can then run ioring poll

```nim
import std/ioring
import std/threadpool
import std/syncio
import std/atomics

proc write(fd: cint, buf: pointer, size: int): int {.passive.} =
  result = 0
  var cnt = delay()
  discard gRing.submitWrite(fd, buf, size, cnt, result.addr)
  suspend()
  return result

var done = false

proc main() {.passive.} =
  var buf = "Hello world\n"
  var n = write(stdout.getFileHandle, buf.toCString, buf.len)
  echo "written=" & $n & " buf.len=" & $buf.len
  atomicStore(done, true, moRelaxed)

# would be cool just running main()
# and it works but it requires ro run workerLoop in main thread
defaultPool().submit(delay main())

while not atomicLoad(done, moAcquire):
  discard
```

lib/std/ioring/backends/iouring.nim
blackmius: i hope i understand you and added exception handling

lib/std/ioring/backends/iouring.nim
blackmius: well, yes, and it is works only if have luck be called inside same thread which submitted operation before

lib/std/ioring/backends/poll.nim
```diff
@@ -0,0 +1,85 @@
...
+      of opRead:
+        if (firedEvents and EvRead) == 0: continue
+        let r = posixRead(fd, s.buf, s.len)
+        complete(j, if r >= 0: r else: -1)
+      of opWrite:
+        if (firedEvents and EvWrite) == 0: continue
+        let r = posixWrite(fd, s.buf, s.len)
+        complete(j, if r >= 0: r else: -1)
+      of opAccept:
+        if (firedEvents and EvRead) == 0: continue
```
Araq: Use structured control flow.

lib/std/ioring/backends/epoll.nim
```diff
@@ -0,0 +1,115 @@
...
+
+# Header-free per the std/posix convention: real symbols are bare-importc'd
+# and constants are hand-written ABI transcriptions. EPERM/EBADF are 1/9 on
+# every Linux ABI (asm-generic, shared by amd64/arm64/i386; musl agrees).
+const
+  epollErrPerm = cint(1)   # EPERM
+  epollErrBadFd = cint(9)  # EBADF
+
+proc errnoLocation(): ptr cint {.importc: "__errno_location", sideEffect.}
+  ## glibc's and musl's address-returning errno accessor (same pattern as
```
Araq: Cool. But we use neither. No libc.

lib/std/ioring/backends/epoll.nim
```diff
@@ -0,0 +1,115 @@
...
+  ## std/posix). Read immediately after a failing epoll_ctl, no intervening
+  ## call, to classify the failure.
+
+proc reportResidualFailure(msg: string) =
+  ## A residual epoll_ctl failure (both the primary op and its fallback
+  ## failed on a live pollable fd) — report on stderr with the errno number.
```
Araq: No, we don't dumb shit on `stderr` in a standard library. Instead use a relay for logging or use exception handling.

blackmius: there is no relay for logging

oh, didnt expected to backport technical debt from merges
https://github.com/nim-lang/nimony/blob/b7e5691070024bfed80919cb51862be6fd02e3ab/lib/std/threadpool.nim#L238

blackmius: removed reportResidualFailure and errnoLocation

Araq: > there is no relay for logging

I know, but one could be added, if required.

blackmius:
the throughput leaves much to be desired

```
NOPs total: 3342767 in 10.000000046 sec
NOPs/sec: 334276.6984623272
```

```nim
import std/syncio
import std/monotimes
import std/atomics

import std/ioring
import std/threadpool

var running: bool
running.atomicStore(true, moRelaxed)
const maxTasks = 4
var counters: array[maxTasks, int]
var nextIdx: int = 0

proc counterProc() {.passive.} =
    let idx = nextIdx.atomicFetchAdd(1, moRelaxed)
    while running.atomicLoad(moRelaxed):
        let c = delay()
        discard submitNop(cont = c)
        suspend()
        inc counters[idx]

for i in 0 ..< maxTasks:
    submit(delay counterProc())

let t0 = getMonoTime()
let duration = 10*1_000_000_000'i64

while true:
    let elapsed = getMonoTime().ticks - t0.ticks
    if elapsed >= duration:
        break

running.atomicStore(false, moRelaxed)

var total = 0
for c in counters:
    total += c

let elapsed = getMonoTime().ticks - t0.ticks
echo "NOPs total: ", total, " in ", elapsed.float / 1_000_000_000.0, " sec"
echo "NOPs/sec: ", total.float / (elapsed.float / 1_000_000_000.0)

```

Araq:
> the throughput leaves much to be desired

Well maybe @tokyovigilante has a better solution or can share some insights, he uses .passive procs in production.

blackmius:
i run a benchmark on master line
| maxTasks | throughtput / s |
|--------|--------|
| 4 | 4 165 429, 2 460 148, 6 907 825 |
| 8 | 155 155, 141 045, 419 881 |
| 16 | 115 871, 257 956 , 27 150 |
| 32 | 6778, 10 085, 9860 | 

hmm, after setting WorkerCount to 3 because there is only 4 cpus on my linux vm
| maxTasks | throughtput / s |
|--------|--------|
| 32 | 14 887 398, 14 598 752, 14 882 826 |
| 64 | 14 668 994, 14 646 027, 14 098 987 | 
| 512 | 10 429 005, 10 626 075, 10 518 182 | 
| 2048 | segfault  | 

changed StripeCount to 32 and stripe StripeSize to 1024
| maxTasks | throughtput / s |
|--------|--------|
| 2048 | 9 630 120, 9 882 480, 10 624 258 |
| 8192 | 9 749 732, 10 389 812, 10 468 147 |
| 32768 | 10 187 190, 10 377 935, 10 694 564 |

tried megatest with StripeCount=256 and StripeSize=4096
| maxTasks | throughtput / s |
|--------|--------|
| 262144 | 10 202 315, 10 757 893, 10 830 326 |
| 524288 | 11 023 037, 10 965 474, 10 982 261 |
| 1048576 | 11 958 142, 11 672 414, 12 129 726 |

```
./bin/nimony c -d:release -r bench_threadpool.nim
```

```nim
import std/syncio
import std/monotimes
import std/atomics
import std/threadpool
initPool()
var running: bool
running.atomicStore(true, moRelaxed)
const maxTasks = 8
var counters: array[maxTasks, int]
var nextIdx: int = 0
proc counterProc() {.passive.} =
    let idx = nextIdx.atomicFetchAdd(1, moRelaxed)
    while running.atomicLoad(moRelaxed):
        submit(delay())
        suspend()
        inc counters[idx]
for i in 0 ..< maxTasks:
    submit(delay counterProc())
let t0 = getMonoTime()
let duration = 10*1_000_000_000'i64
while true:
    let elapsed = getMonoTime().ticks - t0.ticks
    if elapsed >= duration:
        break
running.atomicStore(false, moRelaxed)
var total = 0
for c in counters:
    total += c
let elapsed = getMonoTime().ticks - t0.ticks
echo "NOPs total: ", total, " in ", elapsed.float / 1_000_000_000.0, " sec"
echo "NOPs/sec: ", total.float / (elapsed.float / 1_000_000_000.0)
```

blackmius:
one thing it is clear from benches that we must folllow one thread per cpu rule
and somehow dynamicaly reallocate stripes

blackmius:
rerun benchmark on my branch with WorkerCount=3

maxTasks | throughtput / s
-- | --
262144 | 10 053 432, 8 617 563, 10 003 103
524288 | 10 311 561, 9 372 682, 10 343 429
1048576 | 10 660 413, 11 348 024, 11 458 594

and submitting noOp still slow
ok, here the problem with deffered queue and one global lock, need to be redesigned

lib/std/ioring/core/slots.nim
```diff
@@ -0,0 +1,116 @@
...
+import std/ticketlocks
+
+const MaxOps* = 8192
+
+type
+  SlotArena* = ref object
+    slots*: array[MaxOps, OpContext]
+    freelist: seq[int]
+    fdHeads: Table[cint, int]     ## fd -> head slot index of its op list
+    lock: TicketLock
```
blackmius, line 23: remove global lock here try something stripes and worksteal

lib/std/ioring/backends/iouring.nim
```diff
@@ -0,0 +1,171 @@
...
+  DeferredSize = 4096 ## Must be a power of 2.
+  DrainBatch   = 128  ## Max deferred entries drained per poll() call.
+
+type
+  DeferredEntry = object
+    slotIdx: int
+    fd: cint
+
+  DeferredQueue = object
+    lock: TicketLock
```
blackmius, line 30: and here

lib/std/threadpool.nim
blackmius: run cpuCount()-1 by default

blackmius, line 20: come up with dynamicaly changable stripe sizes

Araq:
My threadpool shows how to use lock-stripes, I bet you can adapt these ideas.

tokyovigilante:
> Well maybe @tokyovigilante has a better solution or can share some insights, he uses .passive procs in production.

Can take a look at these benchmarks, my testing has all been against my HTTP server which can do 50k requests/sec with an actual response (ie `GET /` -> `200 OK`) but thats with `epoll`. That's "fast enough" for me but can have a look. 

[EDIT] with 1000 connections, which is not bad I think.

Another data point is my abortive attempts to get `libdispatch` /GCD working with Nim 2 on Linux, it did and I could get ~1.5M blocks/sec via a serial queue, which was pretty neat, but I abandoned in favour of async then this as I could never get it to have the main queue on the main thread, which got too hard to manage with UI code and `dbus` etc.

blackmius:
> HTTP server which can do 50k requests/sec

io_uring should easily do 300k requests/sec by my benchmarks 3 years ago

> Another data point is my abortive attempts to get libdispatch /GCD working with Nim 2 on Linux, it did and I could get ~1.5M blocks/sec

and for this i have a benchmark of how many ops can io_uring handle at all

| queue entries | throughput ops / sec |
| - | - |
| 64 | 40 943 475 |
| 128 | 26 783 682 |
| 256 | 45 132 038 |
| 512 | 36 318 568 |
| 1024 | 33 441 647  |
| 2048 | 32 872 614  |
| 4096 | 32 159 553 |

so i just need to properly deliver ops submition

```nim
import std/posix/io_uring
import std/monotimes
import std/syncio

const repeat = 1_000_000

proc run(entries: int) =
  try:
    var q = newQueue(entries, {})
    var count = 0
    # run 1 million nops to check queue throughput
    let t0 = getMonoTime()
    var cqes = newSeq[Cqe](entries*2)
    while count < repeat:
        for i in 0..<entries:
          let sqe = q.getSqe()
          if sqe != nil:
            discard sqe.nop()
        q.submit()
        let cqesCount = q.copyCqes(cqes, entries.uint)
        count += cqesCount
    let elapsed = getMonoTime().ticks - t0.ticks
    var time = elapsed.float / 1_000_000_000.0
    var rps = repeat.float / time
    echo q.params.sqEntries, " ", rps
  except:
    discard
for entries in @[64, 128, 256, 512, 1024, 2048, 4096]:
  run(entries)
```

lib/std/threadpool.nim
```diff
@@ -415,45 +165,36 @@ proc workerLoop(arg: pointer) {.nimcall.} =
...
-      # workers read garbage thread indices (valgrind: uninitialised value in
-      # tryBulkDequeue/workerLoop, origin the reused frame). No storage, no
-      # lifetime, no race.
-      create workers[i], workerLoop, cast[pointer](i)
-    except:
+  if atomicLoad(poolState, moAcquire) == 2 or workerCount > 0: return
+  var expected = 0
+  if atomicCompareExchange(poolState, expected, 1):
+    workerCount = max(1, cpuinfo.countProcessors() - 1)
+    workers.setLen(workerCount)
```
blackmius, line 183: i dont understand why but statically allocated array of workers shows better throughput, now it is degraded from 14M -> 8.5M strange, i havent read the rawThread implementation maybe it somehow adds seq acessors in hot path with bound check but from threadpool.nim perspective it looks like it is accessed only on initialisation and shutdown

lib/std/ioring/core/slots.nim
```diff
@@ -0,0 +1,116 @@
...
+
+import ./types
+import std/tables
+import std/ticketlocks
+
+const MaxOps* = 8192
+
+type
+  SlotArena* = ref object
+    slots*: array[MaxOps, OpContext]
```
Araq, line 20: Seems wasteful.

lib/std/ioring/core/slots.nim
```diff
@@ -0,0 +1,116 @@
...
+    a.slots[idx] = OpContext(inUse: true, fd: fd, nextInFd: -1, prevInFd: -1)
+    let head = a.fdHeads.getOrDefault(fd, -1)
+    if head >= 0:
+      a.slots[head].prevInFd = idx
+    a.slots[idx].nextInFd = head
+    a.fdHeads[fd] = idx
+    result = true
+
+proc allocSlot*(a: SlotArena; fd: cint): int =
+  ## Convenience wrapper that raises on exhaustion. Prefer `tryAllocSlot`
```
Araq, line 56: It doesn't raise...

lib/std/ioring/core/slots.nim
```diff
@@ -0,0 +1,116 @@
...
+# held across a syscall or a user callback.
+
+import ./types
+import std/tables
+import std/ticketlocks
+
+const MaxOps* = 8192
+
+type
+  SlotArena* = ref object
```
Araq, line 19: Why is this a `ref`?

lib/std/ioring/core/slots.nim
blackmius: i think we should not share epoll between threads then we can reject global slotArena and locks. and we can dynamicaly allocate slots

blackmius:
| maxTasks | submit throughput/s | mainline submit | submitNop throughput/s |
| - | - | - | - |
| 4 | 95 241 526 | 17 522 677  | 8 777 240 |
| 16 | 127 058 760 | 44 392 835 | 14 536 195 |
| 32 | 8 981 746 | 14 932 373 | 17 226 215 |
| 64 | 8 528 955 | 13 758 668 | 11 710 481 |
| 128 | 8 415 316 | 14 060 888 | 10 255 108 |
| 512 | 8 542 743 | 14 583 024 | 9 053 481 |
| 2048 | 8 381 293 | 14 560 222 | 9 268 246 |
| 65536 | 8 805 410  | 13 039 172 | 7 914 645 |

blackmius:
bulkSize drasticaly changes throughput

<table>
<tr><th>bulk \ maxTasks</th><th>4</th><th>16</th><th>32</th><th>64</th><th>128</th><th>256</th><th>512</th><th>1024</th><th>2048</th></tr>
<tr><td><b>16</b></td><td>129_162_699</td><td><b>170_466_506</b></td><td>9_643_318</td><td>9_000_211</td><td>8_833_037</td><td>9_017_681</td><td>8_968_648</td><td>8_969_985</td><td>9_107_752</td></tr>
<tr><td><b>32</b></td><td>126_839_735</td><td>169_203_024</td><td><b>176_137_462</b></td><td>9_347_813</td><td>9_228_476</td><td>9_303_775</td><td>9_282_237</td><td>9_356_827</td><td>9_432_143</td></tr>
<tr><td><b>64</b></td><td>126_106_716</td><td>169_476_519</td><td><b>173_557_994</b></td><td>139_851_078</td><td>9_489_587</td><td>9_442_938</td><td>9_350_645</td><td>9_359_087</td><td>9_528_737</td></tr>
<tr><td><b>128</b></td><td>129_490_199</td><td>164_252_233</td><td><b>172_861_151</b></td><td>143_670_759</td><td>73_678_413</td><td>9_587_575</td><td>9_546_190</td><td>9_538_299</td><td>8_595_682</td></tr>
<tr><td><b>256</b></td><td>127_388_945</td><td>165_726_741</td><td><b>175_308_414</b></td><td>140_185_496</td><td>71_763_904</td><td>32_419_191</td><td>9_644_781</td><td>9_572_275</td><td>8_167_690</td></tr>
<tr><td><b>512</b></td><td>125_932_756</td><td>165_869_713</td><td><b>173_592_652</b></td><td>138_448_700</td><td>76_152_176</td><td>33_658_605</td><td>11_128_038</td><td>8_693_146</td><td>9_313_455</td></tr>
</table>


ok seems it works because it just fits on single worker and they wont share counter i think i miss cache-line padding on benchmark

UPD: benchmark counters alignment did not help

blackmius:
found the problem the submit always submits with hint=0 to stripe 0 but it should by default submit to running worker
```nim
proc counterProc() {.passive.} =
    while running.atomicLoad(moRelaxed):
        let c = delay()
        # discard submitNop(cont = c)
        submit(c)
        suspend()
        inc counters[threadIdx].value
```

with changing logic to somethieng like

```nim
proc submit*(c: Continuation; hint = -1) {.inline.} =
  ## Convenience: submit a bare continuation as a task.
  var hint = hint
  if hint == -1:
    hint = threadIdx
  submit(toTask(c), hint)
  ```
  
  the numbers start look stable
  
  <table>
<tr><th>bulk \ maxTasks</th><th>4</th><th>16</th><th>32</th><th>64</th><th>128</th><th>256</th><th>512</th><th>1024</th><th>2048</th></tr>
<tr><td><b>16</b></td><td>22_959_020</td><td>29_250_210</td><td>37_227_940</td><td>37_203_895</td><td><b>37_552_544</b></td><td>34_498_180</td><td>34_255_007</td><td>30_963_751</td><td>33_138_813</td></tr>
<tr><td><b>32</b></td><td>22_325_804</td><td>28_246_520</td><td>32_413_942</td><td>36_541_281</td><td>35_153_694</td><td><b>37_705_742</b></td><td>32_044_251</td><td>33_731_311</td><td>34_604_803</td></tr>
<tr><td><b>64</b></td><td>25_684_742</td><td>28_457_962</td><td>32_600_713</td><td>29_774_824</td><td>31_128_582</td><td>30_231_916</td><td>29_214_260</td><td>30_889_957</td><td><b>33_262_352</b></td></tr>
<tr><td><b>128</b></td><td>22_622_892</td><td>33_310_692</td><td><b>37_195_954</b></td><td>30_224_388</td><td>25_989_546</td><td>29_372_025</td><td>29_623_781</td><td>29_721_768</td><td>29_026_606</td></tr>
<tr><td><b>256</b></td><td>20_602_609</td><td>30_173_434</td><td>34_800_919</td><td>34_554_679</td><td>33_185_411</td><td>37_531_995</td><td>36_243_879</td><td><b>38_146_636</b></td><td>37_179_854</td></tr>
<tr><td><b>512</b></td><td>29_017_519</td><td>33_309_853</td><td>37_637_748</td><td>34_839_044</td><td>32_882_149</td><td>37_958_890</td><td>38_165_417</td><td><b>39_553_891</b></td><td>38_329_874</td></tr>
</table>

**Comment by blackmius** [comment]
ok, i made fastest multithread counter

<table>
<tr><th>bulk \ maxTasks</th><th>4</th><th>16</th><th>32</th><th>64</th><th>128</th><th>256</th><th>512</th><th>1024</th><th>2048</th></tr>
<tr><td><b>16</b></td><td>193_014_839</td><td>161_365_078</td><td>144_206_906</td><td>164_099_564</td><td>120_211_561</td><td><b>196_682_441</b></td><td>185_167_178</td><td>110_640_399</td><td>138_451_237</td></tr>
<tr><td><b>32</b></td><td>182_584_274</td><td><b>289_310_964</b></td><td>275_942_377</td><td>227_637_823</td><td>109_026_216</td><td>135_794_263</td><td>96_374_722</td><td>98_364_810</td><td>141_292_392</td></tr>
<tr><td><b>64</b></td><td>182_321_648</td><td>291_197_701</td><td>278_470_884</td><td>270_080_669</td><td><b>309_619_329</b></td><td>213_867_666</td><td>220_517_384</td><td>216_375_758</td><td>180_440_015</td></tr>
<tr><td><b>128</b></td><td>197_657_916</td><td>295_764_638</td><td>282_730_274</td><td>217_329_847</td><td>220_310_220</td><td><b>345_458_697</b></td><td>320_519_231</td><td>318_877_651</td><td>314_028_159</td></tr>
<tr><td><b>256</b></td><td>194_135_167</td><td>298_722_660</td><td>275_183_244</td><td>285_064_484</td><td>277_845_827</td><td>274_236_942</td><td><b>418_613_892</b></td><td>155_433_627</td><td>159_427_909</td></tr>
<tr><td><b>512</b></td><td>202_903_269</td><td>287_970_283</td><td>295_494_225</td><td>238_600_539</td><td>274_755_147</td><td>276_261_990</td><td><b>378_052_046</b></td><td>168_501_030</td><td>329_094_777</td></tr>
</table>

but what about latencies and what if one task will stuck in blocking syscall operation ie nanosleep it will block other staged tasks and other workers could not even steal them

**Comment by blackmius** [comment]
```nim
import std/monotimes
import std/syncio
import std/threadpool
import std/parfor
initPool()

proc dfs(depth, breadth: int): uint32 =
  if depth == 0:
    return 1
  var sums = newSeq[uint32](breadth)
  for i in `||`(0, breadth-1, workload = CpuBound):
    sums[i] = dfs(depth-1, breadth)
  result = 0
  for i in 0..<breadth:
    result += sums[i]

var
  depth = 9
  breadth = 9

let start = getMonoTime().ticks
let answer = dfs(depth, breadth)
let stop = getMonoTime().ticks
echo "answer: ", answer
echo "time (s) ", (stop-start).float32 / 1_000_000_000.0

for i, wm in workerMetrics.pairs:
  echo "WorkerMetrics id=", i, ", enqueueMiss=", wm.enqueueMiss, ", dequeueMiss=", wm.dequeueMiss,
    ", tasksSubmited=", wm.tasksSubmited, ", tasksHandled=", wm.tasksHandled,
    ", tasksStealed=", wm.tasksStealed
```

```
daniil@ubuntu24:~/projects/nimony$ ./bin/nimony c -r -d:danger -d:release bench_dfs.nim 
[ 62%] done
answer: 387420489
time (s) 2.8503258228302
WorkerMetrics id=0, enqueueMiss=0, dequeueMiss=307520, tasksSubmited=145254855, tasksHandled=145253143, tasksStealed=0
WorkerMetrics id=1, enqueueMiss=0, dequeueMiss=49459, tasksSubmited=0, tasksHandled=13914, tasksStealed=4326
WorkerMetrics id=2, enqueueMiss=0, dequeueMiss=98552, tasksSubmited=0, tasksHandled=15558, tasksStealed=2039
```

well it poorly balance workload

tokyovigilante:
Ran this against `hashi`, my CPS-based HTTP server and my curl multisocket integration, both consumers of std/ioring + std/threadpool) and read through the new backend plumbing. The architecture is cool, eg, per-worker rings, the deferred-SQE design etc, and closeFd cancelling is very helpful.

 Two API notes as well as some inline comments:
 
1. `registerFd`/`rearmFd`/`IoHandler` are gone from std/threadpool. I drive libcurl's multi-socket engine using these functions, and a completion-only model can't express "this fd became readable, ask curl what to do". Can we keep a readiness registration API (under `io_uring` it seems to map to multishot `POLL_ADD`)? Without it every poll-style integration needs a rewrite.
 
2. Small losses: `MaxFds` is gone (I size a `fd`-indexed connection table with it), `poolStopped` was renamed stopped, and std/ioring no longer re-exports initPool/shutdownPool.

With those restored I'm happy to test the branch properly, have a bunch of test suites plus a load-test instance, which would be ideal for direct comparison.

lib/std/ioring.nim
```diff
@@ -6,342 +6,150 @@
...
-    # flag set first the racer MODs; registerFd/rearmFd also fall back on
-    # EEXIST/ENOENT so either interleaving converges to an armed fd.
-    slot.registered = true
-    registerFd(fd, addr slot.handler, evMask)
+proc submitNop*(cont = Continuation(fn: nil, env: nil);
+                resPtr: nil ptr int = nil): SeqNum =
+  result = nextSeqNum()
+  var op = OpContext(kind: opNop, fd: -1, seqnum: result,
+    cont: cont, res: cast[int](resPtr))
+  discard gOpQueues[threadIdx].tryEnqueue(op)
```
tokyovigilante: A full stripe silently drops the op here. The continuation parks forever with no error, and at 8192 entries a burst can reach this. The task path solved the same problem with caller-runs; ops need equivalent backpressure (retry, caller-runs into a blocking submit, or at least a loud failure).

lib/std/ioring/backends/iouring.nim
```diff
@@ -0,0 +1,106 @@
...
+      discard sqe.read(op.fd, cast[pointer](op.buf), op.len)
+  of opWrite:
+    if op.buf != nil:
+      discard sqe.write(op.fd, cast[pointer](op.buf), op.len)
+  of opAccept:
+    discard sqe.accept(SocketHandle(op.fd), cast[ptr SockAddr](addr op.acceptAddr), addr op.acceptLen, 0)
+  of opNop:
+    discard sqe.nop()
+
+proc iouringPoll(timeoutMs: int): bool {.nimcall.} =
```
tokyovigilante: timeoutMs is never used: this poll can't block, unlike the epoll backend's epoll_wait(timeoutMs). Combined with waitCompletions' `while true: discard poll(0)` loop, an idle process spins at 100% per polling thread under this backend only. 

`io_uring_enter(GETEVENTS, min_complete = 1)` with a timespec (or an eventfd bridged into the worker's sleep) would restore the blocking behavior the interface implies.

lib/std/ioring/backends/iouring.nim
```diff
@@ -0,0 +1,106 @@
...
+  var n = gOpQueues[threadIdx].tryBulkDequeue(DrainBatch, buf)
+  if n > 0:
+    for i in 0..<n:
+      var sqe: nil ptr Sqe
+      try:
+        sqe = localQueues[threadIdx].getSqe()
+      except ErrorCode as e:
+        stderr.writeLine("ioring: failed to get sqe: " & $e)
+        break
+      if sqe == nil:
```
tokyovigilante: `tryBulkDequeue` has already removed up to DrainBatch ops at this point; breaking on SQ exhaustion (256 entries, reachable under load) loses `buf[i..n-1]`, neither submitted nor re-queued, so their continuations hang. The undrained remainder needs to go back on the queue (or the dequeue needs to be capped at free-SQE count).

lib/std/ioring/core/backend.nim
```diff
@@ -0,0 +1,58 @@
...
+
+proc complete*(slotIdx: int; res: int) =
+  let slot = addr gSlots[threadIdx].slots[slotIdx]
+  if slot.op.res != 0:
+    cast[ptr int](slot.op.res)[] = res
+  let cont = slot.op.cont
+  slot.op.cont = Continuation(fn: nil, env: nil)
+  gSlots[threadIdx].freeSlot(slotIdx)
+  if cont.fn != nil:
+    submit(cont, int(slot.op.fd))
```
tokyovigilante: freeSlot(slotIdx) on the line above zeroes the slot, so this reads fd 0 every time - the affinity hint is dead. Capture fd before freeing (same for anything else read from slot.op after this point).

lib/std/ioring/core/backend.nim
```diff
@@ -0,0 +1,58 @@
...
+  else:
+    gCqLock.acquire()
+    if gCqCount < CqSize:
```
tokyovigilante: When the CQ is full the completion is dropped with no trace. A poller waiting on it never learns the op finished. Even if overflow is considered operator error, it deserves a counter or a stderr line; silent loss here is very hard to debug from the outside.

lib/std/ioring/backends/iouring.nim
```diff
@@ -0,0 +1,106 @@
...
+        break
+      let idx = gSlots[threadIdx].allocSlot(buf[i])
+      sqe.userData = cast[pointer](uint(idx))
+      fillSqe(sqe, buf[i].addr)
+    try:
+      discard localQueues[threadIdx].submit()
+    except ErrorCode as e:
+      quit "fatal: bug: submit cannot fail: " & $e
```
tokyovigilante: io_uring_enter can legitimately fail: EAGAIN/EBUSY under CQ pressure, so this isn't a can't-happen branch; it turns load into a process abort. Suggest handling as backpressure (retry after draining CQEs) rather than quit.

lib/std/ioring/backends/iouring.nim
tokyovigilante: Actually this does nanosleep on idle so probably ok. However the worker loop's nanosleep is a fixed poll cadence, not an event-driven wake: under master, a sleeping worker is woken by epoll_wait the moment I/O lands; here a CQE that arrives mid-sleep waits out the tick. That's ~1 kHz idle wakeups per worker and up to a tick of added completion latency. 

Both points have the same fix: `io_uring_enter(GETEVENTS, min_complete = 1)` with a timespec as the idle block (and epoll's `epoll_wait(timeoutMs)` equivalently on that backend). Then the nanosleep can go away entirely and waitCompletions can lean on the same blocking poll?

PR #2186, opened 2026-07-25
propagate .closure properly

Araq:
refs https://github.com/nim-lang/nimony/pull/2178

### DIFF
--- a/src/nimony/sem.nim
+++ b/src/nimony/sem.nim
@@ -2088,7 +2088,7 @@ proc semWhenImpl(c: var SemContext; dest: var TokenBuf; it: var Item; mode: When
       let condStart = dest.len
       var phase = SemcheckBodies
       swap c.phase, phase
-      semConstBoolExpr c, dest, it.n, allowUnresolved = c.routine.inGeneric > 0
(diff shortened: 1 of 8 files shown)


src/nimony/sem.nim
```diff
@@ -4726,6 +4726,55 @@ proc semConv(c: var SemContext; dest: var TokenBuf; it: var Item) =
   it.typ = destType
   commonType c, dest, it, beforeExpr, expected
 
+proc closureProcType(c: var SemContext; typ: TypeCursor): TypeCursor =
+  ## `typ` with `.closure` added — the type of a `(toClosure X)` node. Mirrors
+  ## exactly what `typenav.getTypeImpl` computes for that node on the hexer
+  ## side, so frontend and backend agree on what the wrapper is worth.
+  var t = typ
+  let kind = t.typeKind
+  if kind notin {ProctypeT, ItertypeT}: return typ
```
Redict, line 4735: bug/assert would be more honest

Araq: No, instead commonType later will produce a real error message.

PR #2193, opened 2026-07-27
fixes #2192; report invalid set element types in type declarations instead of asserting

### LINKED ISSUE #2192: `type X = set[uint64]` crashes with an assertion instead of reporting an error
```nim
type
  X = set[uint64]
```

```sh
$ nimony c test.nim
[  0%] test.nim
Error: unhandled exception: src/lib/nifcore.nim(712, 3) `c.rem == 0` into: body did not consume all 17 children (left 5) [AssertionDefect]
```

Same for any invalid set element type in a type declaration, e.g. `type Y = set[string]`. Expected is a proper error message like the one the set constructor path produces (see `tests/nimony/errmsgs/twrongsetelem.nim`):

```
Error: type uint64 is too large to be a set element type
```

Cause: the `SetT` branch of `semLocalTypeImpl` appends the `(err ...)` node after the already closed `(set ...)` tree, so the enclosing type decl gets an extra child and a later fixed-arity `into:` walk over the decl asserts.

Found while working on #2098 (PR #2188); reproduces on current master (91a428ed) and is independent of forward declarations.

Redict:
fixes #2192

The `SetT` branch of `semLocalTypeImpl` appended the `(err ...)` node after the already closed `(set ...)` tree, giving the enclosing type decl an extra child; a later fixed-arity `into:` walk over the decl then asserted:

```
Error: unhandled exception: src/lib/nifcore.nim(712, 3) `c.rem == 0` into: body did not consume all 17 children (left 5) [AssertionDefect]
```

Now the `(set ...)` tree is replaced by the error node, with the set tree kept as the error's origin, so `type X = set[uint64]` reports:

```
tests/nimony/errmsgs/tsetdecltoolarge.nim(3, 10) Error: type uint64 is too large to be a set element type
tests/nimony/errmsgs/tsetdecltoolarge.nim(4, 10) Error: set element type must be ordinal
```

(Same replace-instead-of-append idea as `buildErrAt` from #2124, done inline since that PR is not merged yet; thanks @demotomohiro for the pointer.)

### DIFF
--- a/src/nimony/semtypes.nim
+++ b/src/nimony/semtypes.nim
@@ -1062,20 +1062,26 @@ proc semLocalTypeImpl*(c: var SemContext; dest: var TokenBuf; n: var Cursor;
     of SetT:
       if tryTypeClass(c, dest, n):
         return
+      let setStart = dest.len
       var elemTypeStart = 0
       dest.takeInto n:
         elemTypeStart = dest.len
         semLocalTypeImpl c, dest, n, InLocalDecl
       let elemType = cursorAt(dest, elemTypeStart)
+      var errMsg = ""
       if containsGenericParams(elemType):
         # allow
         discard
       elif not isOrdinalType(elemType, allowEnumWithHoles = true):
-        c.buildErr dest, info, "set element type must be ordinal"
+        errMsg = "set element type must be ordinal"
       else:
         let length = lengthOrd(c, elemType)
         if length.isNaN or length > MaxSetElements:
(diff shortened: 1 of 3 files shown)


Redict:
Rebased onto master and switched to the `buildErrAt` proc from #2194. The `SetT` branch now just calls `c.buildErrAt dest, setStart, errMsg`.

Retested on top of 886885af: errmsgs 38/38 (incl. the new `tsetdecltoolarge`), sets 4/4, enums 10/10, object 19/19, generics 29/29, types 3/3, basics 2/2, casestmt 4/4, consteval 5/5, concepts 20/20.

PR #2196, opened 2026-07-28
intrinsics: model them as declarations, not as call-tag misuse

Araq:
Replaces the `importc`-string sniffing for `__builtin_*` with two pragmas, one shared row table and a dedicated Leng node. This is phases 2 and 3 of `nativenif/doc/intrinsics.md`; the legacy path (mem*, atomics, syscalls) is untouched.

    proc ctz64(x: uint64): int32 {.intrinsic: Ctz.}
    proc bsf(x: uint64): uint64 {.instruction: bsf.}

New tags, appended to doc/tags.md (ids are positional): `(instruction X)` and `(intrinsic X)` for NimonyPragma + LengPragma, and `(instr SYM X*)` for LengExpr + LengStmt. `(instr …)` is typed exactly like `(call …)` — the callee's signature drives everything — but a distinct tag, so a consumer sees "not an ABI call" from the tag alone, with no symbol resolution.

`src/lib/intrinsics.nim` is the single source of truth: one row per opcode carrying operand roles, the two-address `tie`, target availability, effects and a signature shape. It is deliberately dependency-free (plain data, no NIF API) so nimony's nifstreams-based front end and arkham's nifcore-based back end can both read it. Note it bootstraps under nimony itself, hence the ordinal loop and the fully-spelled-out const rows.

The declaration is the typing contract: `sempragmas` unifies the row's shape against the signature ONCE, at the declaration, after which the symbol is an ordinary typed proc and no later pass needs a rule per opcode. The message lands in the routine's `effects` slot — the one slot that already accepts an `(err …)` — so the positional tree stays intact; sem re-reads what its signature phase wrote, so an `(err …)` found there is passed through rather than re-reported as "`effects` must be empty".

hexer resolves the callee once (`applicationTag`) and bakes the answer into the tag. The C backend lowers the portable rows to `__builtin_ctz`/`clz`/ `popcount`/`bswap` and rejects a target-pinned row by name; the LLVM backend rejects `(instr …)` outright rather than substituting something else, since `llvm.cttz`/`ctlz` need a `declare` plus an extra `i1` operand that `genCallWithType` cannot express.

Tests: tests/nimony/intrinsics covers the C backend and every declaration-site rejection; tintrinsics also joins hastur's NativeTestFiles, so the C and native backends are checked to agree on results they reach by different instructions.


src/lengc/codegen.nim
```diff
@@ -567,6 +575,13 @@ proc genProcDecl(c: var GeneratedCode; n: var Cursor; isExtern: bool) =
   var prc = takeProcDecl(n)
 
   let prag = parseProcPragmas(c, prc.pragmas)
+  if prag.flags * {InstructionP, IntrinsicP} != {}:
+    # No C declaration: an intrinsic has no definition to link against, and
+    # every application of it is an `(instr …)` lowered at the use site.
+    c.code.setLen signatureBegin
+    c.m.closeScope()
+    c.currentProc = oldProc
+    return
```
Redict, line 624: `c.inToplevel` stays false past this return; the normal exit restores it at 653. `genVar`'s `vk == IsLocal and c.inToplevel` at 542 then stops routing top-level locals into `c.init` for the rest of the module.

src/lengc/llvmcodegen.nim
```diff
@@ -561,6 +561,11 @@ proc parseProcPragmasLLVM(c: var LLVMCode; n: var Cursor): PragmaInfo =
       of SelectanyP:
         result.flags.incl pk
         skip n
+      of InstructionP, IntrinsicP:
+        # An intrinsic declares a machine instruction, not a callable: there is
+        # nothing to emit for the declaration itself.
+        result.flags.incl pk
```
Redict, line 583: flag is set but never read, so this falls through to the `LLFunc` branch at 676 and emits an empty unterminated function. C got the guard at codegen.nim:578.

src/lengc/llvmgenexprs.nim
```diff
@@ -1008,6 +1008,12 @@ proc genExprLLVM(c: var LLVMCode; n: var Cursor; result: var LLValue) =
     let retTypeCursor = getType(c.m, n)
     let retType = genTypeLLVMReadOnly(c, retTypeCursor)
     genCallWithType(c, n, retType, result)
+  of InstrC:
+    # The portable rows map onto `llvm.cttz`/`ctlz`/`ctpop`/`bswap`, but each
+    # needs a `declare` and (for cttz/ctlz) an extra `i1 is_zero_poison`
```
Redict: `genGccBuiltinCall` at 383 already emits the declare and the `i1` operand, plus the i64->i32 truncation. reachable from `IntrinsicNames[op]` + `bits` via the same `cBuiltinFor` switch.

src/lengc/genexprs.nim
```diff
@@ -68,6 +68,86 @@ proc genCall(c: var GeneratedCode; n: var Cursor) =
...
+    if n.substructureKind == PragmasU:
+      var p = n
+      p.loopInto:
+        let pk = p.pragmaKind
+        if pk in {InstructionP, IntrinsicP}:
+          var a = p; a = sub(a)
+          if a.kind == Ident:
+            result = intrinsicOpByName(c.m.pool.strings[a.strId],
+                       (if pk == InstructionP: icPinned else: icPortable))
+        skip p
```
Redict: re-derives what lengcgen.nim:822 resolved and :999 flattened to a name. arkham's `firstIntParamBits` is a third copy. `(instruction bsf 64)` would make all three a token read.

src/lib/intrinsics.nim
```diff
@@ -0,0 +1,220 @@
...
+  OperandRole* = enum
+    roIn         ## a pure source
+    roOut        ## a pure destination
+    roInout      ## read *and* written (needs the `var`-first spelling)
+
+  IntrinsicEffect* = enum
+    efPure       ## no memory effect, no trap: CSE / fold / DCE eligible
+    efReads      ## reads memory through a pointer operand
+    efWrites     ## writes memory through a pointer operand
+    efBarrier    ## no memory reordering across it
```
Redict, line 109: `roles` is `AllIn` on every row and `effects` is `{efPure}` on every row; neither has a reader in nativenif#85. `targets` and `tie` do.

PR #2201, opened 2026-07-28
fixes #1822; inline for-loops inside closure-iterator bodies

### LINKED ISSUE #1822: Closure iterators
still not sure should closure iterator change its api from what it is in nim2.0

first option is to preserve old api, which also makes cps transform easier for calls, but requires lambdalift and cps handle it together and looks it will work slower than in nim 2.0

second option is to transform iterators so they start returning next continutation and result at once via tuple, but it will require to transform each call site and it totaly changes logic of calling x twice and etc

```nim

import std/syncio

block:
    proc myiter1(): proc(): int {.closure, passive.} =
        var cont = Continuation()
        var iter_result = 0
        proc body() {.passive, closure.} =
            iter_result = 1
            cont = delay()
            suspend cont.env.caller
            iter_result = 2
            cont = delay()
            suspend cont.env.caller
            iter_result = 3
            cont = Continuation()
        cont = delay body()
        return proc(): int {.closure, passive.} =
            var this = delay()
            cont.env.caller = this
            suspend cont
            return iter_result

    type 
        EnvData = object of RootObj
            cont: Continuation
            iterResult: int64
        Env = object of RootObj
            d: EnvData
        Closure = tuple
            fn: proc ()
            env: ptr Env
        
    var x = myiter1()
    while cast[ptr Closure](x.addr)[].env[].d.cont.fn != nil:
        echo x()

block:
    proc myiter2(): (proc(): (Continuation, int) {.closure, passive.}) =
        var cont = Continuation()
        proc (): (Continuation, int) {.closure, passive.} =
            cont = delay()
            result = (cont, 1)
            suspend()
            cont = delay()
            result = (cont, 2)
            suspend()
            result = (Continuation(), 3)

    var x = myiter2()
    var iter = x()
    while true:
        echo iter[1]
        if iter[0].fn == nil:
            break
        iter[0].complete()
```

Redict:
A `.closure` iterator with a `for` loop in its body crashed hexer:

```nim
iterator countup(a, b: int): int {.closure.} =
  for i in a..b:
    yield i

for x in countup(1, 3):
  echo x
```

```
[Bug] could not find symbol: i.0
```

`transformStmt`'s `IteratorS` case routes `.closure`/`.passive` iterators to `rewriteClosureIter`, which only rewrote `yld` and copied everything else verbatim. Nothing ever descended into the body, so the inner `for` survived as a `(for ...)` into duplifier, which tripped over a for-var no pass had declared.

Fix: run `transformStmt` over the body first, then `rewriteYieldsAndCopy` over its output. The order matters. Inlining copies the loop body once per inner yield, and those copies can contain outer `yld`s that still need the rewrite.

Test: `tests/nimony/cps/tclosure_iter_nested_for.nim`, covering both a single and a nested `for`. It fails on master and passes here.

The nested case from the issue (an anonymous closure iterator returned from a proc) is a separate bug and still fails, on `duplifier.nim(841) fnType.substructureKind == ParamsU`.


src/hexer/iterinliner.nim
```diff
@@ -498,8 +498,23 @@ proc rewriteClosureIter(e: var EContext; dest: var TokenBuf;
...
+    # pass ever declared ("could not find symbol: i.0").
+    # Inlining must run before the yield rewrite: the inline iterator's
+    # expansion copies the loop body once per inner yield, and each copy may
+    # contain an outer `(yld ...)` that still needs rewriting.
+    var bodyBuf = createTokenBuf()
+    swap dest, bodyBuf
     while c.hasMore:
-      rewriteYieldsAndCopy(e, dest, c, synthResultSym)
+      transformStmt(e, dest, c)
+    swap dest, bodyBuf
```
Araq:
```suggestion
    var bodyBuf = createTokenBuf()
while c.hasMore:
      transformStmt(e, bodyBuf, c)
```

Redict: Applied, thanks. Dropped the trailing `swap` too, since it would have exchanged the buffers back and sent the `rewriteYieldsAndCopy` output into `bodyBuf`. `tests/nimony/cps` is 28/28.

PR #2209, opened 2026-07-29
fixes #2155; remove libc header imports from std/posix entirely

### LINKED ISSUE #2155: Compiling `os.getLastModificationTime` causes compile error from gcc on ARM64 CPU
`mytests/testos.nim`:
```nim
import std/os

try:
  discard getLastModificationTime("testos.nim")
except:
  discard
```

It compiles without errors on AMD64 CPU.
But generates compile errors on Raspberry Pi5:
```
$ nimony c mytests/testos.nim
nimcache/tesh6d999/oswd7dmm.c:534:12: error: conflicting types for ‘stat’; have ‘int(NC8 *, struct stat *)’ {aka ‘int(unsigned char *, struct stat *)’}
  534 | extern int stat(NC8* a1_6, struct stat* a2_5);
      |            ^~~~
In file included from nimcache/tesh6d999/oswd7dmm.c:427:
/usr/include/aarch64-linux-gnu/sys/stat.h:205:12: note: previous declaration of ‘stat’ with type ‘int(const char * restrict,  struct stat * restrict)’
  205 | extern int stat (const char *__restrict __file,
      |            ^~~~
nimcache/tesh6d999/pososrh1q1.c:452:5: error: conflicting types for ‘stat’; have ‘int(NC8 *, struct stat *)’ {aka ‘int(unsigned char *, struct stat *)’}
  452 | int stat(NC8* a1_6, struct stat* a2_5);
      |     ^~~~
In file included from nimcache/tesh6d999/pososrh1q1.c:429:
/usr/include/aarch64-linux-gnu/sys/stat.h:205:12: note: previous declaration of ‘stat’ with type ‘int(const char * restrict,  struct stat * restrict)’
  205 | extern int stat (const char *__restrict __file,
      |            ^~~~
[ 99%] sysvq0asl.o
```

On `lib/std/posix/posix.nim`, `Stat` object type is defined without header pragma on AMD64 CPU and OSX.
But on other platforms, `Stat` object type is defined with `header: "<sys/stat.h>"`.
And if `nimNativeIo` is defined (it is defined in default), `stat` proc is imported without header pragma.
So when compiling generated C code, signature of `stat` function declared in `<sys/stat.h>` and `stat` function in generated C code doesn't match and causes compile error.

Defining `Stat` object on ARM64 like `Stat` object on AMD64 can fix this issue, but the same compile error is probably still generated on other CPU.

Redict:
Fixes #2155. Per review, this became the full purge: std/posix no longer imports any libc header. Types and constants are hand-written per-arch ABI transcriptions, procs are bare `importc` of real libc/kernel symbols. `-d:useLibc` now only picks the allocator and stdio implementation, never which declarations exist.

The original bug class dies with it. #2155 was a header-based `struct stat` colliding with a bare `extern int stat(...)` in one TU; with no headers there is nothing to collide with. Header importc also can't work on the LLVM and native codegens, which is what motivated going all the way.

What keeps hand-transcribed ABIs honest is `tests/nimony/stdlib/tposixabi.nim`. It generates a C file with one `_Static_assert` per transcribed constant, struct size and field offset, expected values taken from the live Nim declarations, and compiles it with the system compiler. A wrong transcription is a red lane on the platform that can see it, not silent corruption. This already paid for itself: I first sized `pthread_mutex_t` at 40 bytes for both amd64 and arm64. That's right on amd64 and wrong on aarch64 glibc (48), and the probe caught it in CI before anything scribbled past a `SysLockObj`.

The hardcoded `struct stat` layouts:

- linux/amd64: 144 B
- linux/arm64: the kernel's asm-generic layout, 128 B, different field order. riscv64 and loongarch64 share it, so those targets come nearly free later.
- linux/i386: glibc `struct stat64` (LFS), 96 B, bound to the `stat64`/`open64` symbol family so `Off` stays 64-bit.
- macOS: `__DARWIN_64_BIT_INO_T` layout, 144 B, with the `opendir$INODE64` symbols.

Fields other modules don't read are correctly-sized private padding.

`posix_other_consts` went from ~680 header vars to ~125 real constants, per-arch where they diverge. arm64 overrides part of the generic fcntl table, which is how the pre-existing bug where "shared" `O_DIRECTORY` was actually `O_DIRECT` on arm64 slipped in; fixed here. errno codes are regenerated per-OS by `tools/gen_errorcodes`, probe output committed as `errno_cache.txt`.

Two more things the purge flushed out. macOS `gcc` is clang and doesn't know `-Wno-stringop-overflow`, so the flag is now gated to keep `-Wunknown-warning-option` out of tracked test output. And plugins never saw `-d:useLibc` because `compilePlugin` builds its command line from scratch; it now forwards `useLibc`/`useLibcIo`/`useMimalloc` by allow-list (test: `tests/nimony/valgrind/tmacrolibc.nim`).

Verification. All five CI lanes are green, including the linux-arm64 lane this PR adds. I also checked the transcriptions against real headers locally, without CI round-trips: i386 via `gcc -m32` with extracted 32-bit glibc headers, aarch64 via an extracted cross gcc-13 sysroot. Both probes compile clean. Net diff is -208 lines across 28 files.


blackmius:
the nimNativeIo is a full mess, its better have a refactoring because you cant have when defined(NimNativeIo) which goes for every statement separately

Redict:
Agreed. `hasNativeStat` is the shape I'd refactor toward: one capability const per declaration cluster (fd family, mmap, exec, time), so `nimNativeIo` ends up in a few `const` definitions instead of ~20 scattered `when`s.

Kept it out of this PR because the header fallback path had zero CI coverage until the arm64 job added here. Refactoring every cluster against untested fallbacks is a bigger gamble than fixing the one that's known broken. Once this is green I'll do the rest as a follow-up.

Araq:
This whole idea of importing struct types from C is moribund, it doesn't work with LLVM and the native codegens anyway and GCC keeps getting stricter with its strict-typing bullshit (C is not Ada and sucks at becoming one). We want ABI precise declarations in the core. `importc` is for user libraries, not for Nimony core libraries.

blackmius:
maybe then clean whole std/posix from libc imports to have pure nimony types, now it looks it trying to support both modes

Redict:
Fine with me. arm64 uses the kernel's asm-generic `struct stat` (riscv64 and loongarch64 share it) and the exposed fields sit at the same offsets as amd64 except `st_mode`, so it's about 20 lines. Reworking the PR that way.

Purging importc from all of std/posix is a lot more - posix_other_consts alone is a few hundred header vars. Follow-up, unless you want it in this PR.

By the way, the new arm64 lane already caught a real pre-existing bug. The native dir ops claim `O_DIRECTORY` is "shared amd64/arm64" but arm64 overrides the generic fcntl values and `0o200000` there is `O_DIRECT`, so opendir fails and the walkDir tests are red. Will fix here too.

Araq:
> Purging importc from all of std/posix is a lot more - posix_other_consts alone is a few hundred header vars. Follow-up, unless you want it in this PR.

Yup, I want that.

Redict:
Done, this PR is now the full purge; body rewritten for the new scope. All five lanes green, including the new linux-arm64 job.

PR #2210, opened 2026-07-29
optimizers produce properly typed temporaries

Araq:


src/lengc/shoggoth/induction_variables.nim
```diff
@@ -193,7 +200,18 @@ proc addPtrVarDecl(c: var Context; p: string; atCursor: Cursor): int =
   if info.isValid: buf.appendLineInfo info
   buf.addSymDef p
   buf.addDotToken()                          # pragmas
-  buf.addDotToken()                          # type — inferred from init
+  # Declared type: `(ptr typeof <atCursor>)`. Spelled out rather than left to the
+  # backend's re-inference, for the same reason as `cse.emitTempType`.
+  block:
+    var emitted = false
+    if c.m != nil:
+      let t = getNominalType(c.m[], atCursor)
```
Araq: What is this shit? typenav has an API to extract types out of position. If that fails, don't hack around it, let it crash&burn!

PR #2213, opened 2026-07-29
destroyer: don't re-emit a scope's destructors after a return

tokyovigilante:
The destroyer appended a scope's `=destroy` sequence at the end of every scope, including scopes whose statement list already ended in a `return`/`raise`/`break` — the jump statements walk the whole scope chain and emit that same sequence themselves. In straight-line output the tail is unreachable, so it has looked like harmless dead code.

It is not harmless under CPS. `eliminateJumps` runs AFTER the destroyer (pipeline: destroyer -> cps -> njvl) and rewrites `return` into `(jtrue ´r)` plus FALLTHROUGH; at the tail of a proc body it does not even guard what follows with `if not ´r`. The dead destructor sequence becomes live code, so in a `.passive` proc every escaping local is destroyed twice — and because arcopt pairs the now-dead first `=destroy` with the `=wasMoved` and elides both, the surviving destroy is applied to the coroutine-frame field the return value was just moved out of. The caller gets back a string whose buffer has been freed.

Minimised shape: string locals live across suspension points, a `while` holding both a suspension and an early `return`, a further suspension after the loop, and a success tail that builds a string. Before:

    *result = this->desugar_23;
    nimStrDestroy(this->userId_7);
    nimStrDestroy(this->clientId_6);
    this->´r_5 = NIM_TRUE;
    nimStrDestroy(this->desugar_23);   /* the buffer *result now owns */
    nimStrDestroy(this->userId_7);     /* second time */
    nimStrDestroy(this->clientId_6);   /* second time */

A second shape reaches the same duplication differently: a body ending `result = …` followed by an explicit `return` gets `(ret) (ret result)` from the frontend's implicit trailing return, so `trReturn` emitted the sequence twice itself. That variant aborts in glibc rather than merely returning a freed string.

Fix: track whether the statement just translated leaves its list unconditionally (`Context.terminates`; a plain nested `(stmts` is transparent and inherits it from its last child), skip `trScope`'s closing `leaveScope` when it does, and emit no destructors for a `return`/`raise`/`break` entered in already-unreachable code.

Found by a `.passive` proc double-freeing its coroutine frame on the success path in a real service. Scanning the generated C of that binary showed 89 duplicated-destructor sites before the fix and 0 after, so the defect was not confined to one unusual proc shape.

Regression test: tests/nimony/cps/tcps_dtor_epilogue.nim covers both exits (fall-off-the-end and explicit `return`).


src/hexer/destroyer.nim
```diff
@@ -410,30 +448,48 @@ proc trTry(c: var Context; n: var Cursor) =
...
+      c.terminates = true
     of RaiseS:
       trRaise(c, n)
+      c.terminates = true
     of BreakS:
       trBreak(c, n)
+      c.terminates = true
     of IfS:
       trIf c, n
+      c.terminates = false
```
Araq, line 474: Suspicious. Hopefully all these assignments to `false` are dead code?

tokyovigilante: Claude reckons they're live...

They're the half of the tracking that prevents the inverse bug. terminates means "the statement list I am translating has already run the enclosing scopes' destructors via a jump". The jump statements set it; the = false after each COMPOUND statement clears it because the jump that set it terminated the inner list, not the enclosing one. Take the line the comment is on: for
 
```
if cond: return
# ...more statements...
return
```

the `then`-branch's return sets the flag inside `trIf`. Without the clear after `trIf`, everything after the `if` would be treated as unreachable even though the `else`-path reaches it — the second `return` would take the early-exit in `trReturn` and skip destructor emission entirely (leaks instead of double-frees). Same for a `while` whose body breaks: the statements after the loop are reachable.
  
The clears are deliberately conservative in the other direction: an `if` whose every branch returns still clears, so `trScope` appends a tail sequence that is genuinely dead in straight-line output and guarded with `if not ´r` under CPS everywhere except the proc-tail position the fix targets. The 89-site scan of the affected service binary (89 duplicated destructor sites before, 0 after) was done with this conservatism in place.

If you'd prefer the flag not live in Context, `tr` could return it (or it could be scoped per statement list) so the clears collapse into the control flow, happy to restructure; the behaviour is identical.

Araq:
Follow-up: See if tracker.nim can be used to make the solution simpler.

PR #2216, opened 2026-07-29
Pr/ioring epoll oneshot

tokyovigilante:


lib/std/threadpool.nim
```diff
@@ -227,6 +227,25 @@ proc poolHelp*(): bool {.inline.} =
...
 
+proc perror(s: cstring) {.importc, header: "<stdio.h>".}
+  ## libc perror — report a residual epoll_ctl failure on stderr (see the
+  ## ADD/MOD fallbacks in registerFd/rearmFd).
+
+var cErrno {.importc: "errno", header: "<errno.h>".}: cint
+  ## the C errno, read immediately after a failing epoll_ctl (no intervening libc
+  ## call) to classify the failure.
+var epollErrPerm {.importc: "EPERM", header: "<errno.h>".}: cint
+var epollErrBadFd {.importc: "EBADF", header: "<errno.h>".}: cint
```
Araq: We want to become libc-free though.

blackmius:
i did the same thing in my red ioring refactoring branch

lib/std/threadpool.nim
tokyovigilante: I've reworked it to just use a bare write and constants without `stdio`/`errno`, but is there a better mechanism?

PR #2221, opened 2026-07-29
semos: resolve package-dir imports (foo -> foo/foo.nim)

tokyovigilante:
resolveFile only tried `<path>/foo.nim`, so a module shipped as a package directory (`<path>/foo/foo.nim`) was unresolvable — notably std/posix, which nimony ships as std/posix/posix.nim, so `import std/posix` failed unless another module happened to load it first. Add the standard Nim package-dir fallback: when `foo.nim` is absent under a search dir, try `foo/foo.nim`. Test: tests/nimony/modules/tpkgimport imports a deps/pkgmod package.


Araq:
I don't know about this one. If the import is `std/posix/posix` then so be it. Maybe it really helps otherwise though, not sure.

tokyovigilante:
No drama, just a cosmetic thing anyway.

PR #2223, opened 2026-07-29
fixes #1626; constant-fold literal-only float expressions at max precision

### LINKED ISSUE #1626: Automatic type conversion in expressions with different kinds of floating-point is not performed
documentation says

> The compiler always uses the maximum precision available to evaluate
> floating-point values during semantic analysis; this means expressions like
> `0.09'f32 + 0.01'f32 == 0.09'f64 + 0.01'f64` are true.

running 

```nim
import std/syncio
echo 0.09'f32 + 0.01'f32 == 0.09'f64 + 0.01'f64
```

```sh
blackmius@MacBook nimony % ./bin/nimony c -r test3.nim         
false
```

Redict:
`echo 0.09'f32 + 0.01'f32 == 0.09'f64 + 0.01'f64` printed `false` at runtime while the same expression as a `const` is `true`. doc/language.md promises `true`: "the compiler always uses the maximum precision available to evaluate floating-point values during semantic analysis". Runtime expressions were never folded, so the documented rule only applied in `const` context.

The fold lives in hexer, not nimsem: desugar's `tr` dispatcher routes `add/sub/mul/div/neg/eq/le/lt` through `trFloatArith`, which emits the subtree normally and then attempts a constant fold over the finished output. `tr` recursion is bottom-up, so a nested `(add (f32) 0.09 0.01)` folds before the enclosing `(eq ...)` sees it, and the whole comparison collapses to `(true)`.

Deciding what to fold is left to the eval engine. `floatOpBits` looks at the opcode's type child to answer "is this a float op, and how wide", and everything past that is `expreval.eval`'s call: it already reports "cannot evaluate" for anything non-constant instead of failing, so a fold that does not work out leaves the tree alone. The evaluator computes float arithmetic in float64 with no intermediate f32 rounding, which is exactly the documented semantics. Folded arithmetic keeps its static type via a `(suf <lit> "fNN")` wrapper; comparisons become `(true)`/`(false)`.

## `bits` is now part of `EvalContext`

`evalShiftOp` and `evalBitnot` need the target `int` width for masking and used to read it as `c0.c.g.config.bits`, which is a nil deref for any caller without a `SemContext`. It is now a field on `EvalContext`, defaulted from the `SemContext` when there is one and passed explicitly otherwise; `initEvalContext` takes it as a parameter. `evalOrdinal`/`evalBitSet` thread it through as well, and desugar supplies `Pass.bits`.

That path was already reachable on master, independent of float folding: `evalBitSet` calls `evalOrdinal(nil, ...)` from desugar today, and sem does not fold runtime expressions, so a `shl`/`not` inside a **runtime** set constructor arrives with no `SemContext`. On unmodified master (`7049ae2e`), full toolchain:

```nim
var a: set[uint8] = {uint8(1 shl 2), 9'u8}            # SIGSEGV in evalShiftOp
var b: set[uint8] = {uint8((not 0'i32) and 7), 9'u8}  # SIGSEGV in evalBitnot
```

Both compile on this branch. Make them `const` and master is fine, because sem folds the element first, which is why it went unnoticen. Filed as #2228, which also covers a second, separate bug: a folded bitnot element produces the wrong bitset, and that one is **not** fixed here.

The third read, `SizeofX`'s `c.c.semGetSize(c.c[], …)`, now folds to "cannot evaluate" without a `SemContext`, matching the `c.c == nil` arms already in that proc. It is reached by e.g. `float(sizeof(Obj)) + 1.0`. The shift case is also what made `tests/nimony/consteval/tconstreadfile.nim` SIGSEGV inside its sub-compile once folding started exercising it.

## Rest

Non-finite results (Inf/NaN) stay unfolded on purpose; the runtime computes the same value anyway. `!=`/`>`/`>=` need no handling since sem lowers them to `not`/swapped `<`/`<=` before hexer runs.

For the repro the generated C passes `NIM_TRUE` to `write`, no runtime float compare left. tests/nimony/basicarith/tfloatfold.nim covers the repro, const/runtime agreement, a var operand blocking the fold, the f32 type surviving a folded add, and the `sizeof`/`shl`/`not`/call operand shapes. `1 shl 40` folds and runs to the same value, confirming the width is threaded rather than defaulted. Suites green after a full toolchain rebuild: basicarith, const, consteval, typeconversions, sysbasics, generics, misc, sets, strings, errmsgs, basics, calls, overload, times, opt; validator clean on desugar.nim, sem.nim, controlflow.nim and semdecls.nim.


Araq:
I think const-eval should generally be hexer's job and nimsem should only do it on demand when `static-T` requires it.

Redict:
Okay, I will look into it in terms of this PR.

Redict:
Moved the fold into hexer's desugar pass; sem.nim and semcall.nim are untouched now. desugar routes the arith/cmp tags through `trFloatArith`, which folds the finished subtree via `expreval.eval` with a nil SemContext (the literal-only guard means it never needs one). PR description updated.

Araq:
We already have an eval engine, why not use it more? Also the `AddX` etc opcode come with a type as the first child you can inspect quickly for "is this a float op?"

Redict:
Dropped the hand-rolled guard, `floatOpBits` just reads the type child and `eval` decides the rest.

That turned up three nil-`SemContext` crashes in expreval (`evalShiftOp`, `evalBitnot`, `SizeofX`). They are not mine: on unmodified master `var s: set[uint8] = {uint8(1 shl 2), 9'u8}` already segfaults, since `evalBitSet` calls `evalOrdinal(nil, ...)` and sem never folds runtime exprs. `const` hides it because sem folds the element first.

So this PR is now a float-folding change plus three unrelated segfault fixes. Happy to split those out if you want them separately. There may also be more nil-reachable paths through `evalOrdinal(nil, ...)` that I have not enumerated.

src/nimony/expreval.nim
```diff
@@ -465,7 +465,10 @@ template evalUnOp(c: var EvalContext; n: var Cursor; opr: untyped) {.dirty.} =
 template evalShiftOp(c0: var EvalContext; n: var Cursor; opr: untyped) {.dirty.} =
   let orig = n
   var isSigned = false
-  var bits = c0.c.g.config.bits
+  # only a fallback: the IntT/UIntT arm below always overwrites it. `c.c` is
+  # nil for callers outside sem (hexer's folding), so read the target width
+  # from the config only when there is a SemContext to read it from.
+  var bits = if c0.c != nil: c0.c.g.config.bits else: 0
```
Araq: Why would 0 be valid for bits? That is terrible, instead make bits part of `EvalContext` and feed it in the constructor. And make the callers provide it.

PR #2230, opened 2026-07-30
fixes #1988; judge private field visibility by originating module

### LINKED ISSUE #1988: Untyped template instantations in plugins can bypass private field visibility
Not sure if this is a bug or a feature, but it seems wrong to me.

**bugtype.nim**
```nim
type
  Box* = object
    public*: int
    secret: int     # private field!

func accepts*(box: Box): bool =
  ## Always succeed; only used to wrap a constructor in a call expression.
  true
```

**mbugplugin.nim**
```nim
import plugins

proc transform(n: NifCursor): NifBuilder =
  result = createTree()
  var args = n
  if args.stmtKind == StmtsS:
    args = firstChild(args)
  let info = args.info
  let cond = args
  result.withTree StmtsS, info:
    result.withTree IfS, info:
      result.withTree ElifU, info:
        result.withTree NotX, info:
          result.addSubtree cond
        result.withTree StmtsS, info:
          discard

var inp = loadPluginInput()
saveTree transform(inp)
```

**tbugtest.nim**
```nim
import bugtype

template pluginCheck*(cond: untyped) {.plugin: "mbugplugin".}

pluginCheck accepts(Box(public: 1, secret: 2))    # should fail because `secret` is private
## Expected: ``undeclared field: 'secret' for type Box``
## Actual: compiles successfully
```

**tcontrol.nim**
```nim
import bugtype

let x = Box(public: 1, secret: 2)    # this is correctly rejected by the compiler
```

Redict:
`findObjFieldConsiderVis` skipped the visibility check whenever `c.routine.inInst != 0`. That counter is bumped by template expansion, macro plugin expansion *and* generic instantiation, but only the last one is re-checking a body already validated in the field's owner module. An `untyped` template argument is semchecked there for the first time, at a call site that may not see the field, so the check was off exactly where it was needed.

Visibility is now judged against the module the code was written in. The wrinkle is that an expansion mixes two provenances in one tree: the body belongs to the expanded routine's module, the substituted call-site arguments belong to the caller. So a frame carries `{module, file}` and `visibilityModule` walks the stack for the frame whose file matches the access's own line info. Body tokens match, arguments don't.

Two things fell out of testing:

- A routine body has to establish its own home module rather than inherit whatever frame is live. Without it, `widestrs.nim`'s `=destroy` lost access to its own private field when it happened to be semchecked during an unrelated generic instantiation.
- `semfields` now tags its synthesized dots with the `"x"` access token. `fields()` enumerates private fields deliberately, that is what makes generic `==` in `system` work across modules.

Also stores resolved field symbols for generic object constructors, per your note on the issue. `subs` already preserved field SymbolDefs across instantiation, so the `Ident` branch was dead and removing it caused no expectation churn.

**Breaking:** code that reached a foreign private field through an `untyped`/`{.dirty.}` template or a macro plugin now gets the same `undeclared field` error the direct form always gave. Export the field or move the access into the owner module.

Tests: `tfieldvis_expansion` (the bug, constructor + dot + generic forms), `tfieldvis_nested` (expansion inside instantiation across three modules, plus a doubly-nested template chain), `tfieldvisplugin` (the issue's plugin repro), and a `{.dirty.}`-in-generic pair in `mfieldvis` that mirrors `bitabs`'s `getOrInclImpl` and must keep compiling. 597/597 green across 54 suites.


Araq:
This is too subtle. Can we instead of the "x" access token use the string literal for the module suffix? Module suffixes are not ambiguous btw, by design. (Or rather we treat them this way given the unlikelihood of a hash collision.)

Redict:
> Can we instead of the "x" access token use the string literal for the module suffix?

Sure, I'll do that, could be a better carrier.

Redict:
Done, the token now carries the module suffix and the re-check compares it against the field's owner instead of taking the token as a blanket "accept". Reads better too: the emitting site says which module authorized the access rather than just that someone did.

One reserved value survives: `"*"`, for `exprexec`. Its sub-compile runs under a synthesized suffix (`thisModuleSuffix[0..2] & checksum`), so no real module name could ever match there, and `rewriteSymsToIdents` rewrites the token to `"*"` on the way in. Without it `tconsttable` fails on `Table.data`.

Expected `.nif` churn in `tests/nimony/nosystem`: `"x"` becomes `"tge7jvqqk1"` / `"tte5tld8n"`, nothing else.

626/626 across all 55 suites.

PR #2233, opened 2026-07-30
lengc/LLVM: name variant object branches in debug info

### LINKED ISSUE #2068: lengc/LLVM: union object debug info missing kind-to-branch mapping in .c.nif
The generated `.c.nif` for union (variant) objects does not carry the mapping between `kind` values and their corresponding `branch object`. Without this mapping, the generated `.ll` cannot emit the correct DWARF variant part (`DW_TAG_variant_part` / `DW_TAG_variant`) debug metadata. So at #2067 , print struct looks like:

```
* thread #1, name = 'hello', stop reason = step over
    frame #0: 0x00005555555764aa hello`X60Qini_0_helg27y181 at hello.nim:39:6
   36     e1.kind = ekInt
   37     e1.intVal = 42
   38     e1.floatVal = 3.14
-> 39     foo(int32(e1.intVal))
   40     echo e1.intVal, " ", e1.floatVal
(lldb) p e1
(Expr) {
  kind = ekInt
   = {
     = (intVal = 42, floatVal = 3.1400000000000001)
     = {
      strVal = {
        bytes = 42
        more = 0x40091eb851eb851f {...}
      }
    }
  }
}
```

the type of `e1` is:
```
type
  ExprKind = enum
    ekInt, ekFloat, ekStr

  Expr = object
    case kind: ExprKind
    of ekInt, ekFloat:
      intVal: int64
      floatVal: float64
    of ekStr:
      strVal: string
```

Redict:
fixes #2068

`p e1` on a variant object showed unnamed members instead of naming the active branch:

```
(Expr) {
  kind = ekInt
   = {
     = (intVal = 42, floatVal = 3.1400000000000001)
     = { strVal = {...} }
  }
}
```

`llvmdebug` renders what it is given, and the mapping from `kind` values to branches never reached it. `hexer/lengcgen.trObjFields` lowered a case object to a bare `(union (object ...)+)`, dropping each branch's `(ranges ...)` with a `skip n` and emitting nothing at all for a field-less branch. So which values select which branch, and which field is the selector, were both gone by the time the backend ran. That is why this touches two layers.

## Leng gains a discriminated union

```
UnionDecl   ::= (union [FieldDecl | AnonObjDecl | UnionDecl]*)   # {.union.}, unchanged
              | (union UnionBranch+)                             # case object
UnionBranch ::= (of BranchRanges [AnonObjDecl | Empty])
              | (else [AnonObjDecl | Empty])
```

```
(fld :kind.0 . ExprKind.0.)
(union
 (of (ranges 0 1) (object . (fld :intVal.0 ...) (fld :floatVal.0 ...)))
 (of (ranges 2)   (object . (fld :strVal.0 ...)))
 (of (ranges 3).))                                   # `of ekNone: nil`
```

No new tags. `of`, `ranges` and `else` are already `LengOther` and both backends already decode that shape, so `src/models/tags.nim` comes back byte-identical from `gen_tags` and only doc comments move. That also lines up with #1946, which keeps `case` "as existing (quite complex with range handling etc.)".

The discriminator stays positional, meaning the `fld` emitted immediately before the `union`. `trObjFields` establishes that by construction and every consumer already walks union siblings forward, so nothing needs a backward cursor or an extra slot. A `{.union.}` object can never contain a `case` (asserted at `sizeof.nim:87`), so the child tag alone separates the two forms.

Field-less branches now emit `(of RANGES .)` instead of vanishing, which keeps their discriminant value.

## Naming the branches

`llvmdebug` names each branch member and its payload struct after the enum value that selects it. `lengcgen` folds enum symbols to integers when it copies the ranges, so the name has to be recovered from the selector's own `(enum ...)` decl.

```llvm
!15 = !DICompositeType(tag: DW_TAG_structure_type, name: "ekInt", elements: !{!12, !14}, ...)
!16 = !DIDerivedType(tag: DW_TAG_member, name: "ekInt", baseType: !15, ...)
!32 = !DICompositeType(tag: DW_TAG_structure_type, name: "ekStr", ...)
!82 = !DICompositeType(tag: DW_TAG_structure_type, name: "else", ...)
```

An `else` branch has no ranges to name it after, so it gets the literal name `else`. Leaving it anonymous would reproduce the reported bug for every variant with a default branch.

## Why not `DW_TAG_variant_part`

That is the semantically correct DWARF and I built it first, but lldb 22.1.8 cannot parse the tag:

```
error: unhandled type tag DW_TAG_variant_part (DW_TAG_variant_part), please file a bug
(Expr)  (kind = '\0')
```

The payload disappears entirely, which is worse than the bug being fixed. It is not a fan-out limitation either, single-value branches fail the same way. gdb 15.1 does render it correctly, so a variant part is worth adding later behind a flag, but the default has to work on the toolchain this repo pins in its own `clang.args`.

## Consumers

Nine `src/lengc` walks learned to descend the new wrapper, through one shared helper in `nifcdecl.nim`. Two of them were failing silently rather than loudly.

`typenav.typeOfField` never advanced its cursor on an unrecognised node, and its caller loops `while n.hasMore and not done`, so an `of` child was an infinite loop rather than a missed field.

`aliasing.pointerBearing` skipped what it did not recognise and returned false, so a variant whose only pointer lives in a branch read as pointer-free. That is the unsound direction for alias analysis. The same function already skipped nested anonymous object/union children before this change, so that hole existed independently, and both now descend.

`addObjectFieldsLLVM` needed no change. Its callers only ever enter an object or union body, and branches route through `genUnionBodyLLVM` instead.

arkham consumes Leng too, from the sibling `nativenif` repo, and had four walks with the same assumption. That is nim-lang/nativenif#93, merged, and included in the `src/nativenif.commit` pin this branch builds against.

## Verification

Built the toolchain under WSL (Nim 2.2.10, clang 18) and ran a probe covering a multi-value branch, a single-value branch with a `string` field, a field-less branch and an `else`.

- generated C is unchanged, anonymous struct per branch as before
- runtime read-back of every branch field returns the right values, including mutate-then-read, which is the real guard on the GEP index change in `searchFieldIdx`
- LLVM layout `%Expr = type { i8, { { i64, double } } }`, branch access at `i32 0, i32 1`
- post-lengcgen validator over a 123-file `.c.nif` corpus: 280 violations before the grammar edit, 276 after, none new. The four that disappear are exactly the ones this lowering would otherwise have introduced.

Re-verified after rebasing onto current master:

- `gen_tags` leaves `src/models/tags.nim` byte-identical
- native suite **101 / 101**
- main suite **723 / 740**, failing set byte-identical to pristine master in the same sandbox, so no regressions (those 17 are pre-existing concept failures unrelated to this change)

## Not covered by a test

A `case` nested inside another branch is the one shape with no coverage anywhere, so I wrote a test for it. It crashes nimsem at `sem.nim:3808 buildObjConstrField` reaching `nifcore.nim:428`. A nimsem built from pristine master crashes identically, so it predates this PR, and construction is the trigger while a bare declaration is fine. That is #2234. I left the test out rather than add a red one.

### DIFF
--- a/doc/internals/leng-spec.md
+++ b/doc/internals/leng-spec.md
@@ -198,7 +198,10 @@ ProcDecl ::= (proc SymbolDef Params Type ProcPragmas [Empty | StmtList])
 
 FieldDecl ::= (fld SymbolDef FieldPragmas Type)
 
-UnionDecl ::= (union [FieldDecl | AnonObjDecl | UnionDecl]*)
+UnionDecl ::= (union [FieldDecl | AnonObjDecl | UnionDecl]*) | # `{.union.}` form
+              (union UnionBranch+) # case object form
+UnionBranch ::= (of BranchRanges [AnonObjDecl | Empty]) |
+                (else [AnonObjDecl | Empty])
 AnonObjDecl ::= (object Empty [FieldDecl | AnonObjDecl | UnionDecl]*)
 ObjDecl ::= (object [Empty | Type] [FieldDecl | AnonObjDecl | UnionDecl]*)
 EnumFieldDecl ::= (efld SymbolDef Number)
@@ -284,6 +287,17 @@ Notes:
 - `varargs` is modelled as a special type but it must be combined with a named parameter
   just like any other parameters.
 - The type `flexarray` can only be used for a last field in an object declaration.
+- `union` has two forms and the child tag tells them apart: the `{.union.}` form
+  holds `fld`/`object`/`union` children, the case-object form holds `of`/`else`
+  children. The two never mix, because a `{.union.}` object cannot contain a
+  `case`. A case-object `union` is a *discriminated* union: the `of` branches
+  carry the `BranchRanges` that select them, and the **discriminator is the `fld`
+  immediately preceding the `union` among its siblings**. That position is an
+  invariant of the lowering, so a consumer reads the selector by remembering the
+  previous field rather than by searching. A branch with no fields is
+  `(of BranchRanges .)` — the empty body still records that the discriminant
+  value is legal. Both forms have identical memory layout: only the largest
+  branch's size matters.
 - The pragma `selectany` can be used to merge proc bodies that have the same name.
   It is used for generic procs so that only one generic instances remains in the
   final executable file.
--- a/doc/tags.md
+++ b/doc/tags.md
@@ -89,7 +89,7 @@
 | `(if (elif X X)+ (else X)?)` | LengStmt, NimonyStmt, NiflerKind | if statement header |
 | `(when (elif X X)+ (else X)?)` | NimonyStmt, NimonyOther, NiflerKind | when statement header |
 | `(elif X X)` | LengOther, NimonyOther, NiflerKind | pair of (condition, action) |
-| `(else X)` | LengOther, NimonyOther, NiflerKind | `else` action |
+| `(else X)`; `(else .T)` | LengOther, NimonyOther, NiflerKind | `else` action, or the default branch of a Leng discriminated `union` |
 | `(typevars (typevar ...)*)` | NimonyOther, NiflerKind | type variable/generic parameters; after sem an entry may also be a `(staticTypevar ...)` |
 | `(break .Y)`; `(break)` | LengStmt, NimonyStmt, NiflerKind | `break` statement |
 | `(continue .Y)`; `(continue)` | NimonyStmt, NiflerKind, NjvlKind | `continue` statement |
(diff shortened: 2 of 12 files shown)


Redict:
Heads up on a cross-repo dependency I missed initially: this needs nim-lang/nativenif#93 to land first (or at the same time).

arkham consumes Leng too, and it lives in the sibling `nativenif` repo that `ci.yml` checks out at `master`. It has four walks that assume every union child is an `(object …)`, so the branch wrapper trips them and `tests/nimony/arc/tbasics.nim` fails the native job:

```
Error: unhandled exception: arkham x64 v0: type not supported: NoType
66 / 67 native tests successful
```

The other four platforms pass; only linux-amd64 runs the native backend.

With both branches applied, in the sibling layout CI uses: native `67 / 67`, main suite `628 / 644` with a failing set identical to pristine master in the same sandbox.

I have also rebased this onto current master (`2f4c0d60`) since the base had moved.

PR #2240, opened 2026-07-31
fixes #1987; show template calls as inlined frames when debugging

### LINKED ISSUE #1987: lengc/LLVM: template should appear as a function call during debugging

Redict:
Stepping over a template call jumps into the template's body in another file, with no frame to attribute it to. This carries the expansion's provenance to the debug backend, which turns it into DWARF `DISubprogram` + `inlinedAt`.

Rewritten to the filename encoding you asked for, so the `comesfrom` tag is gone: no node, no dispatcher branches, and #2255 is no longer needed. Expanded code gets a forged name:

```
__crucial\0setElem.0.foo\1foo.nim\115\0[]=.0.system\1system.nim\133\0system.nim
^prefix   ^--------- outermost ---------^ ^--------- innermost --------^ ^real
```

The chain is outermost-first, so its length is the inlining depth. Everything after the last NUL is the real file, which is what `realFile()` returns for every consumer that does not care about frames.

### Two things beyond the sketch

Each entry carries `<sym>\1<declfile>\1<declline>`, not just the symbol. The declaration site cannot be recovered downstream: a template decl does not survive into Leng, and the expanded code's own info points at wherever the body came from. Without it a frame for a template declared in the test file claimed to live in `system.nim`. This also makes the result better than the wrapper design, which could only reach the body's first statement and was one line off.

`\1` rather than `|` as the field separator, since Nim lets an operator be named `|`.

### Result

```nim
template setElem(x: ptr UncheckedArray[char]; i: int; elem: char) =
  x[i] = elem

proc run(p: ptr UncheckedArray[char]) =
  setElem(p, 0, 'H')
```

```llvm
!11 = distinct !DISubprogram(name: "setElem", file: !2, line: 15)
!12 = distinct !DISubprogram(name: "[]=", file: !7, line: 33)
!13 = !DILocation(line: 34, scope: !11, inlinedAt: !9)
!14 = !DILocation(line: 34, scope: !12, inlinedAt: !13)
```

`setElem` expands `[]=`, so the chain nests two deep, and each frame points at its own declaration.

### Cost

`system.nim`'s `.s.nif` grows 4.01%, from 63 distinct forged names covering every expansion in the module. Filenames are interned, so a chain is paid for once per distinct expansion rather than per token. I had assumed this would be much worse before measuring it.

### Four things had to learn about the encoding

- `visibilityModule` compares the **real** file. It judges private-field access by which module code was written in, and a forged name has its own FileId, so raw id comparison missed and `tfieldvisplugin` broke.
- `reporters.infoToStr` and the other path printers call `realFile`, or an error message leaks the encoding to the user.
- `(err ...)` subtrees are copied verbatim. Their dot tokens are the instantiation contexts behind `Trace: instantiation from here`; re-emitting them duplicated the trace for a template erroring inside another expansion (`tinvalidrecursion`).
- Dot tokens elsewhere keep their line info. `buildErr` stores the error contexts there, so dropping it silently lost every trace line.

The forge runs **after** the body is sem-checked, not before: a template called inside this one has already expanded and forged its own name by then, so the outer level prepends onto the existing chain and the order comes out outermost-first, which is what nesting `inlinedAt` needs.

One more constraint worth recording: nimony has no closures, so the pass cannot use captured locals. `hastur boot` is what catches that.

### Tests

`tests/llvmdebug`, a golden suite over the `.ll`'s debug metadata and the first tests for the debug backend. Two cases: the nested expansion above, and a `var` declared inside a template, where the `DILocalVariable` scope and its `DILocation` must agree or the verifier rejects the module. Both verified with `llvm-as`.

`hastur.mode = skip` since the LLVM backend cannot build the full stdlib yet; run with `hastur tests/llvmdebug`.

672/672 and `hastur boot`. `tvarargs.nif`, `tresemtype.nif` and `ttemplate.nif` are re-recorded: toplevel `echo` is a varargs template, so its expansion now carries a forged name.

### Not done

`intramodinliner` wraps inlined bodies in `(scope ...)` and has the same missing-`inlinedAt` bug. It can reuse this encoding as-is, since nothing about it is template-specific.


Araq:
Instead of `tmplbody` use the tag name `comesfrom`.

Redict:
Renamed, pushed as a fourth commit.

Better name than mine: it says the statements came from expanding something, without claiming what. `intramodinliner` has the same missing-`inlinedAt` bug and can reuse this tag as-is now, instead of choosing between a template-named tag and a second one.

Comments that read "the template's symbol" now say "the origin symbol" where the code isn't template-specific. `sem.nim`'s re-sem branch keeps the template wording, since that path genuinely is `semTemplateCall`'s.

Tag value stays 348. The `.ll` goldens are unchanged - the tag name never reaches DWARF. 644/644, llvmdebug 2/2.

Redict:
Rebased on master, resolving against my own `visOwner` change from #2230. It wraps `semExpr`; this one wraps the emitted statements, so they nest.

The rebase also surfaced a third pass with the same bug as `derefs`/`vl`: `shoggoth/cse` walked the wrapper as an expression and treated it as a hoist anchor. `useCpuRegisters` in `nifreader.nim` declares an `{.inject.} p` and loads `^p`, both inside the wrapper, so CSE hoisted the load above the declaration and stage 1 failed to compile itself with `'p_11' undeclared`. Fixed in the backend commit.

Worth noting `hastur boot` was the only thing that caught it - the 655-test suite was green.

CI then found a fourth: arkham, in the sibling `nativenif` repo, asserts on the tag in four passes of its own. Fixed in nim-lang/nativenif#95, which should land with or before this.

That is five passes across two repos, each found by breakage in a different suite, and all for the same reason: `(comesfrom SYM S*)` puts an operand where every generic walker expects a statement. `(pragmax pragmas body)` has the identical shape and paid the same tax years earlier - 9 hand-written special cases today. I have written up a proposal for a shared `OperandHeadedS` shape set plus a validator check so the next such tag costs nothing, and will file it separately rather than grow this PR further.

655/655 and boot pass locally.

Araq:
Sorry for this late change-of-mind but comesfrom should be a pure line-info carrying node that is the first child of the existing ExprX construct, this should simplify things quite a bit.

Redict:
A pure line-info leaf is clearly better than a wrapper, and it kills the whole operand-headed problem: nothing has to step over a leading symbol, so the 16 dispatcher branches and the four regressions all go away. #2255 becomes unnecessary.

Before I rewrite it, one thing does not fit and I would rather ask than guess.

`expr` is `NimonyExpr` only, not `NimonyStmt` (`doc/tags.md:278`, `(expr XS+ X)`), and there is no `ExprS`. But the expansions being marked are exactly the *void* ones, which sit in statement position. `semcall.nim` only wraps when `returnType.typeKind == VoidT`, precisely because a non-void expansion is an expression and `beforeCall` has to keep pointing at the value for `commonType`. In the current output that shows up as

```
(comesfrom@,e echo.0.
 (stmts@2,~6
  (stmts@2,1 (cmd write.1. (haddr@6 stdout.0.)"a"@1,5))
  ...
```

sitting directly among statements, holding `stmts`. So `(expr (comesfrom …) …)` cannot go where the marker is needed today without `expr` also becoming a statement kind.

Three readings, and I do not know which you mean:

1. **Make `expr` a statement kind too**, then a void expansion emits `(expr (comesfrom SYM) <stmts…>)`. Fits the "first child of the existing ExprX" wording literally. Touches every `of ExprX` site, though most already loop `while n.hasMore` over the children and treat only the last as the value.

2. **Keep it statement-position but as a leaf**, so `(comesfrom SYM)` is emitted as a sibling *before* the expanded statements rather than wrapping them. Line-info only, no nesting, nothing to descend into. This is the smallest change and gets the whole benefit, but it is not "first child of ExprX".

3. **Mark the call site instead of the expansion**, wrapping non-void expansions where an `(expr …)` already exists and dropping the void case for now.

I lean 2 if the goal is a pure line-info node, and 1 if `(expr …)` in statement position is something you want anyway. Which did you have in mind?

Worth flagging on cost either way: the tag's *value* is derived from its row index in `doc/tags.md`, so changing its shape re-records the `.nif` goldens regardless. That part is mechanical. What I want to avoid is rewriting it twice.

Separately, nim-lang/nativenif#95 is now guarded with `when declared(ComesfromS)` so arkham compiles against a nimony with or without the tag. That breaks the CI deadlock between the two repos independently of which shape wins here.

Araq:
There is a better design still. Capturing line information well was a first class criterion when designing NIF. All we need to do here is to forge a special filename that encodes the real filename as well as `const CrucialPrefix = __crucial\0` prefix (with a binary zero that filenames cannot contain otherwise)

Then LLVM etc can simply extract the comesfrom information by `n.info.filename.startsWith(CrucialPrefix)`. The rest of the compilation pipeline must keep line information as precise as possible but this is an existing requirement!

Redict:
This one removes the tag entirely, which takes #2255 with it. I checked the mechanism holds:

`needsEscape` is `c < ' ' or c in ControlChars` (`nifbuilder.nim:103`), so a `\0` in a filename serializes as `\00` and round-trips through text NIF. `bif.nim` writes filenames length-prefixed, so binary is fine too. The LLVM side already keys off `info.file`: `dbgLocationId` calls `getOrCreateDIFile(c, rawInfo.file)`, and `getOrCreateInlineSP` derives the frame's file and line from the expansion body's own info. Both would read the prefix instead of a pushed frame.

Two things a prefix alone does not carry.

The synthetic `DISubprogram` is named after the template and cached per `SymId`, so every call site of `echo` shares one frame instead of making a new one each time. A filename says the statements came from somewhere, not from what. Encoding the symbol into the forged name (`__crucial\0echo.0.\0<realfile>`) carries it, at one interned filename per expanded symbol per module. Otherwise the SP gets named after the file and loses per-template identity.

Nesting is the other. `setElem` expands `[]=`, so the goldens have `!13 -> inlinedAt !10 -> inlinedAt !7`, two levels, currently a stack of frames. With a flat filename the depth has to live in the encoding too or nested expansions collapse into one. `__crucial\0a\0__crucial\0b\0real.nim` would do it and is not pretty.

Neither is fatal, but both push toward encoding a small record in the filename, so I would rather have your call than my guess.

On keeping line information precise: it is an existing requirement, but it is not currently met on this path, and that is the real risk here. `hexer` and the Leng passes rewrite statements freely, and the info that reaches `lengc` for an expanded body is frequently the call site's, not the template body's. That is why the current patch carries provenance in a node instead of inferring it, and why `getOrCreateInlineSP` documents the body's own info as the only surviving source for the template's file and line. As a filename, every pass that reassigns info has to preserve the forged one, and a pass that substitutes the call site's info produces a frame pointing at the wrong file with nothing to catch it. The node form fails loudly instead.

Happy to build it either way. Since it is the third shape, I would rather pin the encoding down before re-recording the goldens again.

Araq:
Instead of `__crucial\0a\0__crucial\0b\0real.nim` use `__crucial\0a\0b\0real.nim` and then go for this solution because:

Statement-only inling expansion information is not good enough, we want to do this for every template expansion and expression producing templates are very common. This means that any explicit node-based solution would pay a heavy implementation price as expressions are more frequent than statements.

src/lib/nifcore.nim
```diff
@@ -673,6 +673,111 @@ proc lineInfoFile*(c: Cursor): string =
   let li = rawLineInfo(c)
   if li.file.isValid and c.pool != nil: c.pool.filenames[li.file] else: ""
 
+# ── expansion provenance encoded in the filename ─────────────────────────
```
Araq: Move all of this logic to a new module named `src/lib/comesfrom.nim`

src/nimony/templates.nim
```diff
@@ -98,6 +98,110 @@ proc expandTemplateImpl(c: var SemContext; dest: var TokenBuf;
...
+      # `reporters` prints as `Trace: instantiation from here`. Rewriting them
+      # duplicates the trace for a template that errors inside another
+      # expansion (`tests/nimony/templates/tinvalidrecursion.nim`).
+      dest.addSubtree src
+      skip src
+      return
+    dest.addParLe(cursorTagId(src), info)
+    src.into:
+      while src.hasMore: forgeReemit(f, dest, src)
+    dest.addParRi()
```
Araq:
```suggestion
     else:
      dest.addParLe(cursorTagId(src), info)
      src.into:
        while src.hasMore: forgeReemit(f, dest, src)
      dest.addParRi()
```

src/nimony/templates.nim
```diff
@@ -98,6 +98,110 @@ proc expandTemplateImpl(c: var SemContext; dest: var TokenBuf;
...
+  else:
+    # Idents and anything unstructured: copy verbatim, info and all.
+    dest.addSubtree src
+    skip src
+
+proc forgeExpansionInfo*(c: var SemContext; dest: var TokenBuf; start: int;
+                         origin: SymId; declInfo: NifLineInfo;
+                         callInfo: NifLineInfo) =
+  ## Rewrite the line info of everything `dest` gained from `start` onward so
+  ## it records that the code came from expanding `origin` (#1987).
```
Araq, line 171: What?! Why? Sounds terrible. Only the templates body's first token needs the comesfrom information.

Redict: I tried it. Forging only the body's first token gives zero inlined frames:

```
!5  = DISubprogram(name: "run2", ...)
!21 = DISubprogram(name: "X60Qini_0_ttewk27gl1", ...)
!49 = DISubprogram(name: "main", ...)
```

No `withTemp` subprogram, no `inlinedAt` anywhere, and `tests/llvmdebug` drops to 0/1. The forged name does reach the `.c.nif`, exactly one occurrence on the head, so nothing is losing it in the pipeline.

What breaks is the consumption side. The backend has no notion of a current node to inherit from: `setLoc` is called from 46 places, each reading `info` off whatever cursor that lowering happens to hold, and `emit` stamps `currentProc.dbgLoc` onto every instruction. An instruction lowered from the third statement of a template body therefore gets that statement's own info. With only the head marked, everything after the first instruction falls back to the enclosing proc and the frame collapses.

So head-only needs the backend to track the active expansion across a subtree walk, which is a frame stack in `lengc` - the wrapper node again, without the node. Your call which way to go; I did not want to guess a fourth time.

On the cost, since that is the part that sounded terrible: filenames are interned. All of `system.nim` comes to 63 distinct forged names covering 8897 token references, and the `.s.nif` grows 4.01%. The re-emit is a single pass over the expansion buffer at expansion time. Per distinct expansion, not per token.

Both review comments are in: the logic moved to `src/lib/comesfrom.nim`, and I took the `else:` branch. The committed version of it has `else:` one space too deep and does not compile - `templates.nim(145, 6) Error: invalid indentation` - so my commit fixes the indentation on top of yours.

src/lib/nifcore.nim
Redict: Done: `src/lib/comesfrom.nim`. It has no NIF dependency, just string handling, so `nifpools` and `nifcoreparse` re-export it and every consumer keeps reaching `realFile` without a separate import.

src/nimony/templates.nim
Araq: Make no sense to me, in the codegen `trExpr` before it recurses it checks for a comes-from line information, pushes it onto a stack or whatever and pops it later, bam, done, the backend has `DISubprogram` information.

Redict: Doing it that way. The stack in the walker means only the head of an expansion needs a forged name, so the per-token re-emit goes away and most of the size growth should go with it. I'll push once tests/llvmdebug passes again.

src/lib/comesfrom.nim
```diff
@@ -0,0 +1,114 @@
...
+  ## True when `fname` carries expansion provenance rather than being a plain
+  ## source path. Cheap enough to call per token: the `\0` at the end of the
+  ## prefix is checked first, and a real path never has one.
+  if fname.len <= CrucialPrefix.len: return false
+  if fname[CrucialPrefix.len - 1] != '\0': return false
+  for i in 0 ..< CrucialPrefix.len:
+    if fname[i] != CrucialPrefix[i]: return false
+  result = true
+
+proc forgeCrucialFile*(origins: openArray[CrucialOrigin];
```
Araq, line 56: Silly API causes a seq to be created and passed to this. Just export a `addCrucialInfo(dest: var string; ...)`  operation instead.

src/nimony/sem.nim
```diff
@@ -1002,8 +1002,17 @@ proc visibilityModule(c: SemContext; info: NifLineInfo): string =
...
+  ## The comparison is on the *real* file: an expansion's tokens carry a forged
+  ## filename recording where they came from (see `comesfrom`'s `CrucialPrefix`),
+  ## and a forged name has its own `FileId`, so matching raw ids would miss.
+  if c.visOwner.len == 0: return c.thisModuleSuffix
+  if not info.file.isValid: return c.thisModuleSuffix
+  let infoFile = realFile(pool.filenames[info.file])
   for i in countdown(c.visOwner.len-1, 0):
-    if c.visOwner[i].file == info.file.uint32:
+    let ownerId = FileId(c.visOwner[i].file)
+    if not ownerId.isValid: continue
```
Araq, line 1014: Use structured control flow.

src/nimony/templates.nim
```diff
@@ -98,6 +98,110 @@ proc expandTemplateImpl(c: var SemContext; dest: var TokenBuf;
...
+    return li
+  var forged = f.forgedOf.getOrDefault(li.file)
+  if forged == FileId(0):
+    let fname = pool.filenames[li.file]
+    # Prepend, not append: a template's body is sem-checked (and so any template
+    # *it* calls is expanded) before this outer expansion is forged, so the
+    # existing chain is always the inner levels. Outermost first is what the
+    # debug backend needs to nest `inlinedAt` correctly.
+    var origins = @[f.originEntry]
+    for o in crucialOrigins(fname): origins.add o
```
Araq, line 124: Stupid code only necessary because of the stupid `forgeCrucialFile` API.

src/nimony/templates.nim
```diff
@@ -98,6 +98,110 @@ proc expandTemplateImpl(c: var SemContext; dest: var TokenBuf;
...
+    # Info-carrying: `buildErr` records each instantiation context as a dot
+    # token whose line info is the call site, and `reporters` turns those into
+    # the `Trace: instantiation from here` lines. Dropping it loses them.
+    dest.addDotToken(info); inc src
+  else:
+    # Idents and anything unstructured: copy verbatim, info and all.
+    dest.addSubtree src
+    skip src
+
+proc forgeExpansionInfo*(c: var SemContext; dest: var TokenBuf; start: int;
```
Araq, line 167: Expensive logic for a feature nobody really benefits from.

Araq:
Expensive logic for every template expansion and templates are used frequently, even `>=` is one. I don't like slower compiles just because one backend, LLVM thinks it can significantly improve the debug experience when it doesn't. We don't even map closure environments properly yet to debug information.

Make the template forging opt-in, I don't want to pay the price for fantasy problems.

src/lib/comesfrom.nim
Redict: Gone. It is `addCrucialInfo(dest: var string; sym, declFile: string; declLine: int32)` now, appending one entry to the string being built, plus `crucialTail` for the part of an existing name that follows the prefix. Building a forged name is three appends onto `CrucialPrefix` with no seq anywhere, and prepending onto an existing chain no longer parses it back out first.

src/nimony/sem.nim
Redict: Done, single exit with a `while` and a `break`.

src/nimony/templates.nim
Redict: Gone with the API. The nested loop and the per-file `Table` were both there to feed the seq; with `addCrucialInfo` the whole thing is three appends onto a local string.

src/nimony/templates.nim
Redict: Opt-in now, `--inlineframes:on`, off by default. Without it `forgeExpansionInfo` returns before doing anything and a compile is byte-identical to one without this patch, which is what the three `nosystem` goldens went back to: they had been re-recorded when the marking was unconditional, and they are now the pre-patch versions again.

The cost when it is on also dropped a lot, because the walker stack meant only the head of an expansion needs a forged name instead of every token. On a file using `>=`, `..<` and `echo`, `.s.nif` grows 6.59% with the flag and 0% without it.

Fair on closures: they are not mapped, and I am not claiming this fixes debugging generally. It is one thing that was wrong, behind a flag that nobody pays for unless they ask.

Verified at 7362de38: `hastur all` 672/673 on Linux, `hastur boot` SUCCESS with stages 2 and 3 byte-identical, `tests/llvmdebug` 2/2. The single failure is `nosystem/ttemplate`, which fails the same way on the bare merge-base with none of this patch applied.

Redict:
Done, `--inlineframes:on`, default off. `forgeExpansionInfo` returns immediately without it, so a normal compile does no extra work and produces byte-identical output. The three `nosystem` goldens are back to their pre-patch versions to prove that.

Two things changed since the last push besides the flag.

Only the head token of an expansion carries a forged name now, not every token. The LLVM walker pushes the origin when it enters such a head and pops on the way out, so an instruction lowered from the middle of a template body still lands in the frame. That is what you described, and it works once the backend tracks the region rather than reading each token. `>=` is covered: the push is on `genExprLLVM` as well as `genStmtLLVM`.

`forgeCrucialFile` is gone, replaced by `addCrucialInfo` appending into a string, so the seq and the table that fed it are gone too.

Measured on a file using `>=`, `..<` and `echo`: `.s.nif` +6.59% with the flag on, 0% with it off.

Araq:
CI is red though, please take a look.

Redict:
> CI is red though, please take a look.

Fixed.

src/nimony/sem.nim
```diff
@@ -1018,10 +1018,20 @@ proc visibilityModule(c: SemContext; info: NifLineInfo): string =
...
   result = c.thisModuleSuffix
+  if c.visOwner.len > 0 and info.file.isValid:
+    let infoFile = realFile(pool.filenames[info.file])
+    var i = c.visOwner.len - 1
+    while i >= 0:
+      let ownerId = FileId(c.visOwner[i].file)
+      if ownerId.isValid and realFile(pool.filenames[ownerId]) == infoFile:
+        result = c.visOwner[i].module
+        break
+      dec i
```
Araq: Keep the cheap version too for when inlineframe are turned off.

src/nimony/templates.nim
```diff
@@ -119,6 +119,95 @@ proc expandTemplateImpl(c: var SemContext; dest: var TokenBuf;
   else:
     discard "ParRi/close (classic) or stray suffix (nifcore)"
 
+proc forgeExpansionInfo*(c: var SemContext; dest: var TokenBuf; start: int;
```
Araq, line 122: This whole machinery can be avoided or vastly simplified when the template expansion simply always produces a line information suffix for the first expanded node.

src/nimony/sem.nim
Redict: Done. The `FileId` compare is back as the default path; the `realFile` compare only runs under `--inlineframes:on`, which is the only way a forged name exists to miss on.

src/nimony/templates.nim
Redict: You are right, the loop was never needed. An expansion is a single value - a `(stmts ...)` for a statement template, one expression otherwise - so the first node's head already covers the whole body.

I instrumented it to be sure rather than reason about it: counting expansions that produced more than one top-level value across the whole test suite gives 0. Marking only the first node produces byte-identical `.s.nif` output, same 688 forged names on the benchmark file, and `tests/llvmdebug` still shows all three statements of a 3-statement template body inside the frame.

So the loop is gone. What is left is: read the head, decide whether to mark it, re-emit it once.

PR #2264, opened 2026-08-01
sigmatch: don't assert when a formal type symbol isn't a type decl

tokyovigilante:
Overload resolution crashed nimsem with an AssertionDefect when a template body was semchecked generically and a template parameter appeared in type position — e.g. `proc mk(x: typ): GVec2[typ] = gvec2[typ](x)` inside a `template genCtor(typ: untyped)`. Matching the (already error-typed) argument against the generic routine's formal reached `matchSymbol`'s final `else` branch via the concreteMatch typevar-rematch path, where the formal symbol `fs` resolved to the template parameter itself (symKind ParamY), not a type declaration. `typeImpl(fs)` then tripped `assert result.stmtKind == TypeS`.

Guard the branch: if `fs` doesn't load as a type declaration, treat it as a wildcard placeholder rather than asserting — the real type check happens when the template is instantiated with concrete arguments (where `fs` is a genuine type and this branch is never taken). The crash becomes nimony's normal "not a type" diagnostic.

This is the genVecConstructor pattern in vmath; it no longer crashes the compiler (a separate limitation — the eager template-body type check rejecting a parameter used as a type — still errors rather than deferring to instantiation, and is tracked separately).

(cherry picked from commit 7a8fbabfee0a4fc945f7430455e973ed71eeaf88)

### DIFF
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -1498,6 +1498,15 @@ proc matchSymbol(m: var Match; f: Cursor; arg: CallArg) =
     # fast check that works for aliases too:
     if a.isSymbol and sameSymbol(a.symId, fs):
       discard "perfect match"
+    elif (let fdecl = tryLoadSym(fs);
+          fdecl.status != LacksNothing or fdecl.decl.stmtKind != TypeS):
+      # `fs` names something that is not a type declaration — e.g. a template
+      # parameter left in type position (`symKind == ParamY`) while a template
+      # body is semchecked generically. It is not a valid formal type, so no
+      # argument can match it: reject with the normal diagnostic rather than
+      # asserting in `typeImpl`. (An instantiation with a concrete type binds
+      # `fs` to a real type and never reaches this branch.)
+      m.error InvalidMatch, f, a
     else:
       var impl = typeImpl(fs)
       if impl.isTagLit and impl.cursorTagId == nifpools.ErrT:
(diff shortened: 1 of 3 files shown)


src/nimony/sigmatch.nim
```diff
@@ -1498,6 +1498,15 @@ proc matchSymbol(m: var Match; f: Cursor; arg: CallArg) =
...
       discard "perfect match"
+    elif (let fdecl = tryLoadSym(fs);
+          fdecl.status != LacksNothing or fdecl.decl.stmtKind != TypeS):
+      # `fs` is not a type declaration — e.g. a template parameter (`symKind ==
+      # ParamY`) left in type position while a template body is semchecked
+      # generically, before instantiation (reached via the concreteMatch rematch
+      # path; the argument is typically already an error type). `typeImpl` would
+      # assert here. Accept it as a wildcard rather than crash: the real type
+      # check happens when the template is instantiated with concrete arguments.
+      discard "non-type placeholder (e.g. template param as type) — defer to instantiation"
```
Araq: Not good enough. Type checking shouldn't pray&hope it's a template parameter. What about these makes things more permissive anyway, templates are type-checked too unless declared as `.untyped` in which case they do not enter this whole sigmatch machinery anyway.

tokyovigilante: Ok, just reject non-types here rather than asserting?

PR #2269, opened 2026-08-02
hexer/xelim: bind aggregate values that read through a call

tokyovigilante:
… calls

An object/tuple field whose value reads a location that a LATER sibling field moves was corrupted — silent wrong code, no crash:

  WithSeq(flag: xs.len > 0, n: xs.len, items: xs)   # flag == false (xs emptied)

`trAggregateValue` binds a value in an aggregate position to a cursor temp so it evaluates before a sibling's hoisted `wasMoved`, but only did so for top-level `CallKinds`. `n: xs.len` (a bare call) was bound and read `xs` before the move; `flag: xs.len > 0` is a comparison whose OPERAND is the call, so it was left inline in the constructor and evaluated AFTER the `items: xs` move's `wasMoved`, reading the emptied seq. The check must also be goal-independent: xelim2 — the pass right after the duplifier, where the move is introduced — runs at `ElimExprs`, where `isComplex` does not flag a call.

Bind any value that transitively reads through a call (`readsThroughCall`, goal-independent). Pure atoms/literals/`nil` and structural `kv` nodes (named tuples / table entries) read nothing invalidatable and pass through unchanged.

Adds tests/nimony/arc/tmoved_seq_sibling_field. arc/object/stdlib green.

### DIFF
--- a/src/hexer/xelim.nim
+++ b/src/hexer/xelim.nim
@@ -288,12 +288,29 @@ proc trExprLoop(c: var Context; dest: var TokenBuf; n: var Cursor; tar: var Targ
       trExpr c, dest, n, tar
   tar.t.addParRi()
 
+proc readsThroughCall(n: Cursor): bool =
+  ## Goal-independent: does this value tree contain a call — i.e. a read (like
+  ## `xs.len`) that a sibling field's hoisted `wasMoved` could invalidate?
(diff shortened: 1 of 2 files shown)


Araq:
Better fix is here https://github.com/nim-lang/nimony/pull/2300

PR #2277, opened 2026-08-03
[DRAFT] Fork/pr/nifmake deps freshness

tokyovigilante:
This is an attempt to improve the nif cache freshness issues I've seen during incremental recompiles with improved hashing, but happy to drop if a different direction is wanted.


Araq:
basing the system on `mtime` instead of hashing was a design decision as it's O(1) vs O(file contents), let us keep it this way and instead find the remaining missing/wrong setModificationTime issues.

tokyovigilante:
Hopefully better with #2289

PR #2280, opened 2026-08-03
Change build instructions so it works with nim stable and makes a faster output

leiserfg:


Araq:
I don't think hastur being built in release mode matters. The warnings problem has been addresses upstream already.

PR #2284, opened 2026-08-03
std: build the Windows process vectors from GetCommandLineW/GetEnviro…

Araq:
…nmentStringsW

A PE entry point receives no argc/argv/envp, so the globals the generated `main` fills from its parameters — cmdCount/cmdLine/nimEnviron — stayed empty on the freestanding (-d:nimNativeIo) target: std/parseopt saw no arguments and std/envvars no environment. The C runtime that would normally transcode and split the OS's UTF-16 command line and environment block is exactly what that build does without, so system/winstartup does it itself, from system's module init — which runs before any other module's, and after `main` stored the placeholders it is correcting.

The command line's splitting rules are std/cmdline.parseCmdLine's, deliberately: that is what paramStr/paramCount use on Windows, and the two views must not disagree about where the arguments are. The environment block is transcoded WHOLE so its NUL separators survive into the UTF-8 copy, leaving only the entry starts to collect.

std/envvars takes its nimEnviron branch there too: its Windows branch walks the raw block with libc wcschr, and existsEnv reaches for libc getenv — neither exists on this target. That getenv exclusion has to be a when/else rather than an early return, or the reference still reaches the linker and the image dies at load with STATUS_ENTRYPOINT_NOT_FOUND.


ZoomRmc:
Consider porting the Nim's parseopt API fix from https://github.com/nim-lang/Nim/pull/25571, possibly deprecating/renaming the init routine for non-silent breaking for Nim2 code ports, while there's a window where it's not set in stone for Nimony too.

The concerns in my latest comment there show latent bugs are present in the wild due to use of the sentinel value for argv access.

Porting that API already leads to comments like the following appearing in Nimony's own code. https://github.com/nim-lang/nimony/blob/ac7dda0f8202a2dfbfc5a105b44d117d808b8eeb/src/lengc/shoggoth/patextract.nim#L117-L118

Araq:
@ZoomRmc happy for you to take over this problem. Nimony is the place where we do things right.

PR #2287, opened 2026-08-04
make the inter-module inliner actually inline

Araq:
`intramodinliner.readModuleAnalysis` was a stub returning an empty table. `loadForeignAnalysis` cached that and reported success, so `lookupInlineInfo` fell back to `DefaultInlineInfo` (threshold 100, no weights) for every cross-module callee and `shouldInlineCall` declined all of them: only same-module `.inline` procs were ever spliced. The hot cursor/pool accessors (`kind`, `load`, `isTagLit`, `cursorTagId`, `inc`) all live in `nifcore.nim` / `nifpools.nim` and are called from elsewhere, so none of them was ever inlined. The `.x.nif` carried the annotations all along; the `.dce.nif` the reader consults did not.

Three parts:

* dce1 writes one `(inlines SYM threshold w...)` tree per `.inline` proc into the `.dce.nif` sidecar and skips the section when reading it back for liveness. `intraModuleInline` returns the payloads through `optimizeLengOutput` -> `lengcgen` -> `writeDceOutput`; `readModuleAnalysis` parses them.

* Foreign bodies now come from the callee's post-DCE `.c.nif`, not its `.x.nif`. In a `.x.nif` every generic instantiation, hexer-minted type and string literal is written with the own-module shorthand (`seq.0.Ixdx2fh1.`, `strlit.0.I<hash>.`) and dce2's `resolveSymbolConflicts` later rewrites them all to one canonical owner, so a body spliced into another module resolves those names against a module that no longer declares them ("Symbol not found in NIF module"). deps.nim lists the imported modules' `.c.nif` as ordering-only inputs of the `optimize` node, the same arrangement `arkham` already used.

* A 100-token size cap on cross-module callees (`SHOG_INLINE_MAX` overrides). `.inline` is threshold 0, i.e. "always", so without a cap the depth-4 cascade builds basic blocks that exhaust arkham's register allocator.

Also fixes a latent bug this exposed: `trySpliceCond` folded `(var :t T (call f ...)) (if (elif t ...))` into the guard while `countSymUses` only counted uses inside that adjacent `if`. A plain user local (`var negative = isNegative(x)`) matches the same shape and can be read further down the scope, so dropping its definition left those reads dangling. It now requires the temp to be compiler-minted, the invariant its doc comment already claimed.


src/hexer/intramodinliner.nim
```diff
@@ -1380,7 +1443,10 @@ proc intraModuleInline*(moduleSuffix: string; buf: var TokenBuf) =
...
   ## per dceEmit process.
-  var ma = analyzeModule(buf)
-  var graphs = initTable[string, ModuleAnalysis]()
-  graphs[moduleSuffix] = ensureMove ma
-  annotateInlinePragmas(buf, graphs.getOrQuit(moduleSuffix).inlineInfo)
+  ##
+  ## Also writes each `.inline` proc's computed `InlineInfo` into its own
+  ## pragma (`(inline THRESHOLD w…)`). That annotation is what importers read
+  ## back — see `readInlinePragma` — so no sidecar has to carry it.
+  discard moduleSuffix
```
Araq:
```suggestion
```

src/hexer/intramodinliner.nim
```diff
@@ -1129,7 +1185,15 @@ proc trySpliceCond*(c: var InlinerCtx; dest: var TokenBuf; n: var Cursor;
...
-  if not isLocalName(pool.syms[tmpSym]): return 0
+  let tmpName = pool.syms[tmpSym]
+  if not isLocalName(tmpName): return 0
+  # …and it must be a *compiler-minted* temp (xelim spells them `` `x.<n> ``).
+  # The `countSymUses` check below only looks inside the adjacent `if`, which
+  # is sound for a temp xelim mints solely to hold that guard — but a plain
+  # user local (`var negative = isNegative(x)`) can match the very same shape
+  # and still be read further down the scope, and dropping its definition
+  # then leaves those reads dangling.
+  if tmpName.len == 0 or tmpName[0] != '`': return 0
```
Araq:
```suggestion
```

src/hexer/intramodinliner.nim
```diff
@@ -176,7 +239,10 @@ proc lookupBody(c: var InlinerCtx; calleeSym: SymId; outCur: var Cursor): bool =
   if not loadForeign(c, modul): return false
   let fm = c.foreign.getOrQuit(modul)
   if calleeSym notin fm.bodies: return false
-  outCur = cursorAt(fm.buf, fm.bodies.getOrQuit(calleeSym))
+  let cur = cursorAt(fm.buf, fm.bodies.getOrQuit(calleeSym))
+  if c.maxTokens > 0 and subtreeTokenCount(cur, c.maxTokens) > c.maxTokens:
+    return false
+  outCur = cur
```
Araq:
```suggestion
  outCur = cursorAt(fm.buf, fm.bodies.getOrQuit(calleeSym))
```

src/hexer/intramodinliner.nim
```diff
@@ -127,39 +186,43 @@ proc findForeignFile(c: InlinerCtx; modul, ext: string): string =
...
   result = true
 
+proc subtreeTokenCount(n: Cursor; limit: int): int =
+  ## Number of tokens the subtree occupies (the physical span; elided
+  ## closes do not count). `limit` is only relevant for the comparison the
+  ## callers do, the count is exact.
+  if not n.isTagLit:
+    return 1
+  result = subtreeWidth(n)
+
```
Araq:
```suggestion
  if not n.isTagLit:
    result = 1
  else:
    result = subtreeWidth(n)

```

src/hexer/intramodinliner.nim
```diff
@@ -127,39 +186,43 @@ proc findForeignFile(c: InlinerCtx; modul, ext: string): string =
...
+  ## (`seq.0.Ixdx2fh1.`, `strlit.0.I<hash>.`), and dce2's
+  ## `resolveSymbolConflicts` later rewrites all of them to ONE canonical
+  ## owner module. Splicing such a body into a different module resolves those
+  ## names against the *callee's* module, which by then no longer declares
+  ## them — the importer dies with "Symbol not found in NIF module". The
+  ## `.c.nif` has already been through that merge, so its symbols are
+  ## canonical and safe to copy anywhere. `deps.nim` lists the imported
+  ## modules' `.c.nif` as inputs of the `optimize` node so they are
+  ## guaranteed to exist by the time we look; a module we cannot find is
+  ## simply not inlined from.
```
Araq:
```suggestion
```

src/hexer/intramodinliner.nim
```diff
@@ -48,27 +58,73 @@ proc shouldInline*(info: InlineInfo; argScores: openArray[int]): bool =
...
     counterPrefix: string                   # disambiguates passes (hexer vs dce2)
     bodies: Table[SymId, int]               # same-module callee → offset in `src`
+    ownInfo: Table[SymId, InlineInfo]       # same-module `(inline …)` annotations
+    symUses: Table[SymId, int]
+      # How many times each symbol is *read* across the whole module (defs
+      # excluded). `trySpliceCond` drops a local's definition, which is only
+      # sound when nothing else reads it; a use count taken from the single
+      # statement it inspects cannot see the rest of the scope. Kept current
+      # as the pass splices: freshly-minted syms are tallied when their
+      # subtree is handed back to `trIntra`.
```
Araq: And all this machinery because you couldn't figure out how to create a new unique name for temporaries?! WTF...

src/hexer/intramodinliner.nim
```diff
@@ -176,6 +266,11 @@ proc lookupBody(c: var InlinerCtx; calleeSym: SymId; outCur: var Cursor): bool =
   if not loadForeign(c, modul): return false
   let fm = c.foreign.getOrQuit(modul)
   if calleeSym notin fm.bodies: return false
+  # No size cap: `.inline` is the programmer saying "inline this", and the
+  # only bodies that reach here are the ones that carry it (a proc without
+  # the pragma keeps `DefaultInlineInfo`, threshold 100 with no weights, and
+  # `shouldInlineCall` declines it). Refusing a body for being big would make
+  # `.inline` mean "inline if the compiler feels like it".
```
Araq, line 273:
```suggestion
```

PR #2292, opened 2026-08-04
hexer: closure values as callees/fields/elements (+ env-param primitive)

tokyovigilante:
Closure *values* (not just proc symbols) used as callees, fields, and seq elements
  currently trigger internal compiler errors in hexer (`could not find symbol: f.0`,
  `nifcore c.rem==0`, …). This lowers them.

  Structured as the start of a series so it can land in pieces:

  1. **hoist the env-param primitive to `coro_transform`** — moves the lowered-closure
     env param (`` `ep.0 `` + emitter) out of `lambdalifting`'s privates into the shared
     module, next to `RootObjName`/wrapper-shape. Pure relocation; it's the contract
     every closure-signature-emitting pass must share.
  2. **closure values as callees/fields/elements** (+ cursor-hoist, iterinliner) — the
     lowering, built on that primitive.
  3. Tests: 10 cases.

  Built against current `master`, not ported from an older branch — sits on top of
  #2251 (`c2d25cc9`) rather than fighting it (several pre-#2251 fixes turned out
  unnecessary and were dropped).

  Verified: `hastur build all` + `hastur all` **682/682** + self-host `hastur boot`, all green.

  Deferred (want to agree the shape first): the cross-module case needs a foreign-decl
  canonicalizer reusing the same `` `ep.0 `` primitive, but it overlaps #2251's
  named-type-body lifting — the iterator ctor-collision case rides with it.


src/hexer/duplifier.nim
```diff
@@ -436,6 +436,26 @@ proc callWasMoved(c: var Context; sym: SymId; info: NifLineInfo; typ: Cursor) =
...
+  ## `isCursorField` / trObjConstr and the lifter's `unravelObjField`. Without
+  ## this, a captured `{.cursor.}` local hoisted into a closure environment
+  ## (see lambdalifting) would get `=dup`'d into its env field, re-forming the
+  ## exact ref cycle the cursor was written to break.
+  if le.exprKind != DotX: return false
+  var f = le
+  inc f            # into the dot: object operand
+  skip f           # skip the object operand -> field symbol
+  if f.kind != Symbol: return false
+  let res = tryLoadSym(f.symId)
```
Araq: Fields are not looked up with tryLoadSym!!!! This has a special API in typenav, use it instead!

src/hexer/lambdalifting.nim
```diff
@@ -75,7 +80,16 @@ type
     procStack: seq[SymId]
     dest: TokenBuf
     closureProcs, createsEnv, escapes: HashSet[SymId]
-    localToEnv: Table[SymId, EnvField]
+    localToEnv: Table[(SymId, SymId), EnvField]
+      ## key: (env OBJECT type sym, captured local sym). Local syms are
+      ## only unique per proc (sem restarts numbering; iterinliner's
+      ## `ii temps reset per proc), so a bare-SymId key lets one proc's
+      ## capture leak into an unrelated proc that reuses the SymId.
```
Araq: Yes, and that means you got the `Context` lifetimes wrong! Maybe you need a ProcContext inside Context, see how the other passes deal with this issue.

src/hexer/lambdalifting.nim
```diff
@@ -175,16 +202,58 @@ proc localToField(c: var Context; n: Cursor; local, typ: SymId): SymId =
...
+
+proc maybeMarkForeignClosureUse(c: var Context; s: SymId) =
+  ## Pass 2 must run in modules that merely USE foreign closure values —
+  ## calling, storing or comparing them. The consumer side of the
+  ## closure ABI (tuple projections at call sites, closure-typed local
+  ## decls) is pass 2's rewrite, but every `hasClosures` trigger so far
+  ## keys on module-LOCAL closure declarations, so a pure consumer never
+  ## ran it and emitted a direct call against the producer's tuple ABI.
+  ## Foreign decls are canonicalized to the lifted tuple at load
+  ## (canonForeignDecl); match that or a still-raw `.closure` shape.
```
Araq: What is this shit? Looks like a gross hack on top of a phase ordering issue. Solve the phase ordering issue instead and remember that when implemented correctly, closures really are a local language feature, unlike methods which really are not.

src/hexer/lambdalifting.nim
```diff
@@ -905,7 +1012,19 @@ proc treProcLift(c: var Context; dest: var TokenBuf; n: var Cursor) =
...
+    # closure-typed params/returns lowered to `(fn, env)` tuples).
+    # Schedule the rewritten decl for republishing so downstream passes
+    # (xelim temp typing, lengcgen inlining) type calls against the
+    # LOWERED signature — the published copy is still sem's otherwise.
+    # Deferred like the iter `shouldPublish` flush: publishing mid-walk
+    # would let a later symbol-use of this routine read the rewritten
+    # type and lower it a second time.
+    var copy = createTokenBuf(lift.len)
+    copy.add lift
+    c.shouldRepublish.add (sym, ensureMove copy)
```
Araq: But xelim got fixed in the meantime so this whole machinery might not be necessary anymore.

src/hexer/lambdalifting.nim
```diff
@@ -1335,6 +1635,22 @@ proc elimLambdas*(pass: var Pass) =
     hooks: lambdaHooks(),
     nextTemp: pass.nextTemp         # nested njvl runs continue the xelim counter
   )
+  # Module-wide pre-scan: ANY `(closure)` pragma anywhere — a proc decl,
+  # a proctype alias body, an object field, a parameter type — means
+  # pass 2 must run. The per-shape triggers in pass 1 cannot see closure
```
Araq: Just run the pass always instead, this pre-scan is a terrible hack.

src/hexer/lambdalifting.nim
```diff
@@ -1371,6 +1687,10 @@ proc elimLambdas*(pass: var Pass) =
       # confuse the iter-sym-as-value check.
       for (sym, sig) in c.pendingIterSigs.mitems:
         publishSignature sig, sym, 0
+      # Republish every routine pass 2 rewrote — same timing rationale as
+      # the iter flush above.
+      for i in 0 ..< c.shouldRepublish.len:
+        programs.publish(c.shouldRepublish[i][0], move c.shouldRepublish[i][1])
```
Araq: Indicates a design flaw, we must rethink the compilation pipeline instead.

tokyovigilante:
Thanks, this is a bit sloppier than most. Will work through the feedback.

src/hexer/lambdalifting.nim
tokyovigilante: Noted, dropped.

src/hexer/duplifier.nim
tokyovigilante: `lookupField`/`typeOfField`? Got it.

Araq:
I know -- the last day they dumbed down Claude again...

src/hexer/lambdalifting.nim
```diff
@@ -1198,20 +1461,32 @@ proc tre(c: var Context; dest: var TokenBuf; n: var Cursor) =
...
+        if n.kind == SymbolDef:
+          typeSym = n.symId
         takeTree dest, n        # name
         takeTree dest, n        # exported
         takeTree dest, n        # typevars
         takeTree dest, n        # pragmas
-        if n.hasMore:
-          treType c, dest, n    # body
+        if n.kind == TagLit and n.typeKind == TupleT and isLiftedClosureTuple(n):
+          # already the stable lowered shape (pass 1 itertype rewrite)
```
Araq, line 1448: Not good enough, you cannot detect `isLiftedClosureTuple` reliably.

tokyovigilante: What about reserved field names for lowered tuples? Or a closure type kind?

Orthogonally, field lookup now goes through typenav.

Araq:
Merging it since `isLiftedClosureTuple` is pre-existing.

PR #2310, opened 2026-08-12
Fix Hexer: ExtractModule now support module starting with a number

Gesee-y:
The function `extractModule` in the file `src/lib/symparser.nim` failed to recognize module starting with a number

```nim
proc extractModule*(s: string): string =
  # From "abc.12.Mod132a3bc" extract "Mod132a3bc".
  # From "abc.12" extract "".
  var i = s.len - 2
  while i > 0:
    if s[i] == '.':
      # if the name was "abc.12.1mod" this will fail and output that no module was found
      if s[i+1] in {'0'..'9'}:
        return ""
      else:
        return substr(s, i+1)
    dec i
  return "" 
```

### DIFF
--- a/src/lib/symparser.nim
+++ b/src/lib/symparser.nim
@@ -34,13 +34,17 @@ proc genericTypeName*(key, modname: string): string =
 proc extractModule*(s: string): string =
   # From "abc.12.Mod132a3bc" extract "Mod132a3bc".
   # From "abc.12" extract "".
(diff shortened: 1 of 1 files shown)


Araq:
There are no modules starting with a number in Nim.

Gesee-y:
@Araq The program I was writing was starting with a number.
In fact any file starting with a number failed (things like `1file.nim` wouldn't compile)

PR #2311, opened 2026-08-12
Fix Monotime: Restored monotime arithmetic

Gesee-y:
Monotime arithmetic was commented in the code.

So I just uncommented it,
Imported the time module in it
And transformer side-effect free functions in the `times` module from `proc` to `func` so it could work in `monotimes`


Araq:
Unrelated CI failure, merging.

PR #2317, opened 2026-08-14
intramodinliner: don't substitute a bare nil literal — it loses its type

tokyovigilante:
nil is the one type-polymorphic literal. Splicing it raw into an inlined body (instead of binding it to a typed `(var :p <type> = arg)` copy) leaves downstream codegen with an untyped (nil): an inlined ref =dup hook's nil-guarded deref then renders as C `(*NIM_NIL).field`, which does not compile ("member reference base type 'void'"). Nothing folds the constant-false guard away (condVal doesn't model a bare-nil truth test), so the substitution bought nothing.

Introduced by #2303's inliner rewrite (bisected: 96bd8517 good, d5797970 bad). Reproducer: an inlinable proc taking a ref param, called with a literal nil, whose body assigns the param to a ref field — see the added test.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


Araq:
So give the nil a specific type via `ConvX`, do not lose this optimization.

Araq:
`(nil)` can also carry its type directly via `(nil T)`, nimsem.nim should do this as it knows all about the types. Then this whole set of problems collapses.

Araq:
I did it my way instead.

PR #2318, opened 2026-08-15
Read nimcache from nim.cfg if set

tokyovigilante:
nifler config already emits the legacy nim.cfg hierarchy's nimcache setting into the .cfg.nif dirs tree, but parseConfig had no case for it, so the value was silently dropped and the cache always landed in the project-relative 'nimcache' directory.

- parseConfig now consumes the 'nimcache' entry (compat mode)
- precedence: --nimcache > nim.cfg > "nimcache"


Araq:
Do not mix envvar-additions (which I hate, makes for unreproducible builds) with bugfixes.

tokyovigilante:
> Do not mix envvar-additions (which I hate, makes for unreproducible builds) with bugfixes.

Ta, updated to just pass the flag through.

PR #2336, opened 2026-08-19
hexer: statement-based passes, and lower and/or with lab/jmp

Araq:
Two steps towards the Final IR (nim-lang/nimony#1946, now written up in doc/final_ir.md, which xelim.nim and njvl/finalir.nim already referenced).

Drop the second xelim run
-------------------------

xelim was the only pass with a statement-insertion point, so every other pass that needed a temporary fabricated one in place as `(expr (stmts (var tmp ...)) tmp)` -- the very construct xelim exists to remove. The pipeline therefore alternated between passes that broke the normal form and xelim runs that repaired it, and `xelim2` did no lowering of its own at all: its whole job was flattening what the duplifier had just built.

Two normal-form rules make hoisting to the enclosing statement always sound: control flow is statement-based (xelim1 already did this), and an impure `while` condition becomes a leading body guard, so "in front of the statement" never means "outside the loop" (`xelim.mayBindToTemp`). Given that, each pass gets a `hoisted` buffer plus a flush at its statement boundary and emits statements directly. Applied to `eraiser` (which moves behind xelim1 -- it only ran first because it needed a repair pass after it) and `duplifier`. `xelim2` and the eraiser's `needsXelim` plumbing are gone. `xelim_final` stays: `LowerCasts` unnests calls and binds cast operands to variables, both real lowerings.

lab/jmp as Nimony statements, and the two-target condition compiler -------------------------------------------------------------------

xelim materialised a short-circuit condition as a bool temp plus an if/else diamond, followed by a second diamond re-testing the same slot. Recovering the branch the source wrote takes jump threading plus a proof that the temp is dead -- work the optimizers were asked to redo at every `and`.

`(lab D)` and `(jmp Y)` become NimonyStmt tags (LabS/JmpS, plus LabY), so the hexer passes see them in their ordinary `case n.stmtKind` traversal and lengcgen maps them one-to-one onto the Leng constructs they already were. `xelim.trCondJump` is then Appel's two-target condition compiler: `and`/`or` are exact duals, `not` is a target swap, and short-circuiting is the layout rather than a special case. `trIfFlat` lays the whole if/elif chain out flat with one shared `(lab)` per merge.

Note the CPS state machine also spells its states `lab`/`jmp`, with an integer operand; coro_transform now dispatches on operand kind. Cx runs only for the ElimExprs and LowerCasts goals: nj.nim keeps its cfvar lowering and finalir.nim its own condition compiler.

Flattening moves declarations out of the branch that contained them, so two scoping rules keep declaration, initialisation and destruction on the same set of paths: an arm's `(scope ...)` spans its condition and its body, and a conditionally-evaluated operand gets a scope of its own. Both need destroyer.nim to treat `(scope ...)` as a real destructor scope (it was transparent) and `trJmp` to unwind every scope a jump leaves.

destroyer: divergence via the shared Tracker
--------------------------------------------

`c.terminates` was hand-rolled join logic -- `trIf` cleared it after every `if`, which is `closeBranches` written out by hand, and less accurately (an if/else whose every arm returns should leave the path dead). It is now a read-only template over `lengc/shoggoth/trackers.nim`'s `Tracker.diverged`, the same substrate cse/copyprop use, with the branch protocol wired into trIf/trCase/trWhile/trBlock/trTry and markDiverged/gotoLabel/landLabel at the jumps and joins. 21 assignments deleted; the traversal no longer computes divergence anywhere, and the dead scope-exit destructor after a fully-returning if is gone.

Generated Leng over tests/nimony/stdlib/tjson.nim and its dependency closure is 5.3% smaller (syncio -36%, unicode and parseutils -5.5%). `hastur all` is green, including the self-hosting boot (stages 1/2/3 byte-identical), the nj/vl goldens, contracts and the validator.


doc/final_ir.md
```diff
@@ -0,0 +1,309 @@
...
+Two constraints came out of that:
+
+- **The CPS state machine also spells its states `lab`/`jmp`**, with an
+  *integer* operand (`coro_transform`, `togoto`). Those passes now dispatch on
+  the operand kind: `IntLit` is a coroutine state, a symbol is a structured
+  merge label they pass through untouched.
+- **`nj.nim` and `finalir.nim` keep their own condition lowerings.** `Cx` runs
+  only for the `ElimExprs` and `LowerCasts` goals (`CondJumpGoals`); `nj`
+  lowers a short-circuit to its cfvar (`mflag`/`jtrue`) form and its golden
+  outputs are that form, and `finalir` carries its own condition compiler.
```
Araq:
```suggestion
```

PR #2337, opened 2026-08-19
hexer: mangle passive-proc coro helpers with the defining module's suffix

tokyovigilante:
A passive proc defined in module A and called from module B failed hexer with `could not find symbol: <proc>init.0.<callerModule>`: `coroTypeForProc/coroWrapperProc/stateToProcName` mangled helper names with the transforming module's suffix, but the helpers are generated once, by the defining module. 

Use the proc's own module suffix (the rule `coroWrapperForExternIter` already applies to cross-module closure iters), and publish a placeholder signature for foreign passive wrappers the way `publishWrapperSignature` does for same-module iter trampolines — the caller's hexer process cannot load the defining module's hexer-generated wrapper from the sem index.

### DIFF
--- a/src/hexer/coro_transform.nim
+++ b/src/hexer/coro_transform.nim
@@ -191,24 +191,26 @@ proc coroTrSons*(c: var Context; dest: var TokenBuf; n: var Cursor)
 # Naming helpers
 # ---------------------------------------------------------------------
 
-proc coroNameStem*(procId: SymId): string =
-  ## Returns the symbol's full name minus its trailing module suffix.
-  ## Unlike `extractVersionedBasename`, this preserves any intermediate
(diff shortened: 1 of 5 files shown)


Araq:
Makes no sense to me, you only need to forge the name consistently for cross-module interop.

tokyovigilante:
I can't profess to fully understanding the whole pipeline, but my understanding of it is that the NIF at hexer stage doesn't carry any index, but the cross-module symbol loading is done by nimsem earlier? 

So then when module B calls A's passive proc, B's hexer computes <proc>`init.B while the helpers were generated in A as  <proc>`init.0.A, and so they don't (without this change) reconcile to the same function?

tokyovigilante:
Nice, way better solution. Closing in favour of #2344.

PR #2344, opened 2026-08-21
hexer: name passive-proc coro helpers for the defining module

Araq:
A `.passive` proc defined in module A and called from module B failed hexer with `could not find symbol: <proc>`init.0.<callerModule>`. `coroTypeForProc` / `coroWrapperProc` / `stateToProcName` mangled the helper names with the *transforming* module's suffix, but the helpers are generated once, by the module that defines the proc.

Forging the name consistently is necessary but not sufficient. With only the suffix fixed the call still dies, now on the correctly forged name: `constparams.trCall` does a `getType` on the callee to walk its formals in lockstep with the actuals (passByConstRef / mut-out-lent / varargs / typedesc) and to read `raises` off the pragmas, and the wrapper is hexer-generated — hexer-generated symbols never enter a module's sem index, which is the only thing `tryLoadSym` can read. So the caller's process has to reconstruct the signature locally, from the passive proc's own decl.

Cross-module closure iters escape this only because a foreign iter's wrapper sym appears solely inside a `tupconstr` that carries its type explicitly, so nothing ever types the symbol itself. A passive proc's wrapper is a call target. Same missing signature, different blast radius — so `publishWrapperSignature` now serves both callers rather than growing a second copy for passive procs:

  - one `coroHelperName` minter behind all five naming helpers; the `Extern` pair is now a context-free spelling of the same rule (which lambdalifting needs, holding its own `Context` type).
  - the module-suffix guard is gone. `tryLoadSym` on the wrapper already answers "do I need to publish this?", and a suffix test gets it backwards: foreign is precisely the case that needs publishing.
  - the placeholder copies the routine's own pragmas instead of synthesizing `(closure)`, matching what `generateCoroutineHelpers` does for the real wrapper. That is what would carry `.raises` across a module boundary. Untested in that direction: `{.passive, raises.}` with a result already fails single-module on `constparams.nim(207) n.kind == Symbol`, unrelated to this change.

Diagnosis and regression test from #2337 by @tokyovigilante.

### DIFF
--- a/src/hexer/coro_transform.nim
+++ b/src/hexer/coro_transform.nim
@@ -191,24 +191,32 @@ proc coroTrSons*(c: var Context; dest: var TokenBuf; n: var Cursor)
 # Naming helpers
 # ---------------------------------------------------------------------
 
-proc coroNameStem*(procId: SymId): string =
-  ## Returns the symbol's full name minus its trailing module suffix.
-  ## Unlike `extractVersionedBasename`, this preserves any intermediate
-  ## `I<hash>` segment for generic instances — necessary because two
-  ## different instantiations (e.g. `gen.12.Iaaaa.mod` and
-  ## `gen.12.Ibbbb.mod`) would otherwise produce identical stem
-  ## `"gen.12"` and collide on every synthesised wrapper / coro-type /
-  ## state-proc / field name.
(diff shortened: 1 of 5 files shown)


tokyovigilante:
Will take it for a spin.

PR #2348, opened 2026-08-21
[draft]: guard wraps vs else exploitation, labels, and CPS states

tokyovigilante:
As usual, probably a better way to do this but... Good news is this is about the only major chunk against nimony proper I'm carrying to build my client code.

From Claude:

Branch: fork/pr/njvl-guard-else-close (base 1f8e608b = current master)
Commits:
  1. dc43d87a njvl: keep guard wraps balanced across else exploitation and labels
  2. 09815ad8 cps: turn short-circuit jumps that cross a state boundary into transitions

## The bug chain (all three needed for the repro/consumer code)

1. nj.nim `maybeCloseGuard` emitted `) . )` into a still-open exploited else
   branch → exploited ite got a bogus 4th child (stray dot), guard ite lost
   its else slot (2 children). Silent for vl/togoto (dot parses as the
   optional join slot); the coroutine transform's bounded walker asserts
   `nifcore.nim(862) c.rem == 0 — into: body did not consume all N children
   (left 1)`.
2. nj.nim wrapped `(lab L)` join points inside guard ites / exploited else
   branches — hidden join edges into a branch.
3. coro_transform passed symbol jmp/lab pairs through untouched; when a
   suspension point sits between them, the state assembler splits them into
   different state procs → C error "label used but not defined"
   (`if a and b: <passive call>` is enough).

## Trigger (single-module, `.passive`)

- assert (bug 1+2): `if p == "/a" or p == "/b": return 1` then
  `if c: return <suspending call>` — the test file.
- C label error (bug 3): `if a and b: <passive call>` with the short-circuit
  merge label after the suspension point.

## History

- All three latent on master before #2344; exposed when #2344 routed
  cross-module `.passive` bodies through the coroutine transform (that is why
  consumer builds broke while the suite stayed green — no in-tree test had a
  short-circuit + guard + suspension).
- Red on 1f8e608b: nifcore 862 assert; with only commit 1: cross-state label
  C error; green with both.


Araq:
I'd rather see this effort spend on basing CPS on the new final-ir, nj.nim is moribund.

tokyovigilante:
Fair, do you see that as straightforward for an agent or a big change?

Araq:
It is hard and the agent needs my help.

PR #2367, opened 2026-08-25
the CPS transform reads the Final IR, not NJVL

Araq:
`treIteratorBody` ran `nj.eliminateJumps`, whose entire job is to REMOVE jumps: a monotone `mflag` guard per construct, a `jtrue` that sets every guard between a leaving statement and its target, and `(ite (not g) ...)` wrapped around everything that follows. `toGoto` then put the jumps back. Two inverse rewrites in a row — and the guard machinery's else-exploitation is what produced the unbalanced `ite` that `tests/nimony/cps/treturn_or_guard.nim` pins (it failed with "into: body did not consume all 31 children").

It now runs `finalir.toFinalIr`: `loop`/`ite`/`case`/`lab`/`jmp`, control flow entirely statement-based, not a single cfvar. `toGoto` only has to decide which of those transfers must become a state transition. `nj` leaves hexer's build with this.

Six pieces:

- `loop` has no condition slot and no prelude to rotate — `finalir.trWhile` already made the condition a leading guard whose `else` is `jmp loopExit` — so a suspending loop is one head state plus a back-edge, and `(continue .)` lowers to `jmp head` off a `loopHeads` stack. A non-suspending one becomes `while true`; the trailing back-edge marker is dropped and an earlier `continue` becomes the Final IR's own `jmp`/`lab` pair, because `lengcgen` rejects `ContinueS`.
- FLATTENING. The Final IR wraps every body in its own `(stmts ...)`. A `(lab k)` closes the current state proc, so it can only live at the body's top level; a label inside a nested list would close that list's parens instead. `nj` flattened by construction, so this had no counterpart. Body slots of KEPT constructs go through `trGotoScoped`, which preserves exactly one wrapper.
- `repairCrossStateJumps` (from #2348, now the primary mechanism, not a patch): a symbolic `jmp` whose `lab` lands in a different state proc becomes an integer state transition, with a fall-through `jmp` in front of the converted label. Pairs inside one state stay a `goto`.
- `try`/`except`/`fin`: `nj` lowered it, the Final IR keeps it. Lowered here to an error tracker plus one `(lab)` the body's `raise`s jump to. The condition is `(neq ErrorCode err Success)`, deliberately NOT `(failed err)`: that tag is the eraiser's contract for a raising call's own result temp, and `constparams.localsThatBecomeTuples` retypes every symbol under one to `(tuple ErrorCode T)`. A try whose body cannot raise gets no tracker — otherwise `try: ... finally: ...` fabricates an error path a void coroutine has no frame slot for.
- `case` is a construct in the Final IR ("Format as existing"); a suspending branch now gets the same label split as an `ite` arm.
- A non-suspending construct is WALKED, not copied: a `raise` inside one belongs to an enclosing `try`, and `takeTree` would leave it a proc exit.

`tfinalir_shapes.nim` covers what `nj` never handed this pass: a suspension inside one `case` branch, `try` inside a suspending loop with `break` and `continue` crossing a state boundary, and a nested `try` whose inner handler raises to the outer one. The old path dies on it with "cannot build a default value for type: .".

Typed `except SomeType:` is still not dispatched on; the handler runs for any error, exactly as `nj` had it. `except A, B:` does not get past sem either, so this pass is not where that gets decided.


Araq:
Merging it because it improves things but I need to go over it again later.

PR #2376, opened 2026-08-27
Added support for inheriting from generic concepts

### LINKED ISSUE #2375: Concept inheritance does not support generic concepts
Currently, `concept of` only works with non-generic parent concepts.

The following is rejected by semcheck:
```nim
type
  Foo*[T] = concept
    func foo(x: T): T

  Bar*[T] = concept of Foo[T]     # Error: concept can only inherit from other concepts
    func bar(x: T): T
```

## Workaround

Redeclare the requirements:
```nim
  Bar*[T] = concept
    func foo(x: T): T
    func bar(x: T): T
```

gmpreussner:
Fixes https://github.com/nim-lang/nimony/issues/2375

### DIFF
--- a/src/nimony/semtypes.nim
+++ b/src/nimony/semtypes.nim
@@ -205,15 +205,21 @@ proc declareConceptSelf(c: var SemContext; dest: var TokenBuf; info: NifLineInfo
     dest.addDotToken() # value
   publish c, dest, result, declStart
 
+proc semInvoke(c: var SemContext; dest: var TokenBuf; n: var Cursor)
+
 proc semOneConceptParent(c: var SemContext; dest: var TokenBuf; n: var Cursor;
-                         ownerSym: SymId; parents: var seq[SymId];
-                         hadError: var bool) =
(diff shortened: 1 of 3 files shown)


src/nimony/semtypes.nim
```diff
@@ -205,15 +205,21 @@ proc declareConceptSelf(c: var SemContext; dest: var TokenBuf; info: NifLineInfo
     dest.addDotToken() # value
   publish c, dest, result, declStart
 
+proc semInvoke(c: var SemContext; dest: var TokenBuf; n: var Cursor)
+
 proc semOneConceptParent(c: var SemContext; dest: var TokenBuf; n: var Cursor;
-                         ownerSym: SymId; parents: var seq[SymId];
-                         hadError: var bool) =
+                         ownerSym: SymId; parentHeads: var seq[SymId];
+                         parentTrees: var seq[TokenBuf]; hadError: var bool) =
```
Araq:
```nim
type
  ConceptParent = object
    head: SymId
    body: TokenBuf

...; parents: var seq[ConceptParent]; hadError: var bool) =
```

PR #2386, opened 2026-08-28
mm: a `--mm:NAME` memory management switch, plus `--mm:arc`

Araq:
`system.nim` no longer names a strategy: it says `include "$MM"` and the compiler expands `$MM` to `system/<name>` on the include path, right before it is resolved (in `semInclude` and in `deps.processInclude`, the two places that resolve includes). Adding a strategy is adding a file under `lib/std/system/` -- nothing enumerates them, so no `when` chain that every new strategy has to be threaded through.

`NifConfig.mm` holds the name `normalize`d, because the option is spelled in camelCase (`--mm:atomicArc`) but names a file, and files stay all-lowercase (`system/atomicarc.nim`). `--mm` values are validated to a bare `[a-z0-9]+` name so the option cannot become a way to `include` an arbitrary path.

`defined(gcArc)` / `defined(gcAtomicArc)` fall out generically: `isDefined`'s final fallback compares the normalized symbol against `"gc" & config.mm`, so a strategy this compiler has never heard of gets its define too. `--mm:` is part of `getOptionsAsOneString`, so switching strategies invalidates the cached build configuration and forces a rebuild.

The strategy module owns exactly three primitives (`arcInc`, `arcDec`, `arcIsUnique`). `GC_ref`/`GC_unref` are defined purely in terms of `=dup` and `=destroy` and are therefore strategy independent, so they move to `system/refops` instead of being copied into every strategy. `arcops.nim` becomes `atomicarc.nim` (the default), and the new `arc.nim` is that minus the atomics: cheaper counters, but a `ref` must stay on one thread.


src/nimony/nifconfig.nim
```diff
@@ -206,10 +228,18 @@ proc getOptionsAsOneString*(config: NifConfig): string =
...
   result.add " --bits:" & $config.bits
   result.add " --cpu:" & platform.CPU[config.targetCPU].name
   result.add " --os:" & platform.OS[config.targetOS].name
 
+proc expandMM*(config: NifConfig; path: string): string =
+  ## Expands the `$MM` placeholder in an `include` path to the module implementing
+  ## the selected memory management strategy: `"$MM"` -> `"system/atomicarc"`.
+  ## Applied before the path is resolved, so `resolveFile`'s own `$VAR` (environment
+  ## variable) rule never sees it.
+  result = path.replace(MmPlaceholder, MmDir & config.mm)
```
Araq:
```suggestion
  result = path
  if result.endsWith(MmPlaceholder):
    result.shrink result.len - MmPlaceholder.len
    result.add MmDir
    result.add config.mm
```

src/nimony/nifconfig.nim
```diff
@@ -206,10 +228,22 @@ proc getOptionsAsOneString*(config: NifConfig): string =
...
   result.add " --os:" & platform.OS[config.targetOS].name
 
+proc expandMM*(config: NifConfig; path: string): string =
+  ## Expands the `$MM` placeholder in an `include` path to the module implementing
+  ## the selected memory management strategy: `"$MM"` -> `"system/atomicarc"`.
+  ## Applied before the path is resolved, so `resolveFile`'s own `$VAR` (environment
+  ## variable) rule never sees it.
+  result = path
+  if result.endsWith(MmPlaceholder):
+    result.shrink result.len - MmPlaceholder.len
```
Araq:
```suggestion
    when defined(nimony):
      result.shrink result.len - MmPlaceholder.len
    else:
      result.setLen result.len - MmPlaceholder.len
```

PR #2390, opened 2026-08-28
ioring/threadpool: bugfixes and improvements

Araq:
`slotsForFd` and `hasPendingForFd` took `SlotArena` by value, and every call site passes a seq element (`gSlots[lane]...`), for which nimony materialises a copy. That deep-copied the whole arena -- MaxOps slots, each with a 128-byte Sockaddr_storage, plus the fdHeads table, ~1.6 MB -- on every I/O event, twice per event in `processFd`. They take `var` now, which also makes the aliasing real, and that turns two latent bugs live:

* `freeSlot` reset the links to `Slot()`, i.e. 0, which is a valid slot index -- a freed slot looked like it still pointed at slot 0. Reset to -1.
* `slotsForFd` read the successor *after* the body ran, and the body frees the slot it was handed. Read it before the yield. Without both, a single read completion spins forever (tioringrw hangs at 100% CPU).

Correctness:

* io_uring filled SQEs from the stack-local drain buffer, so an accept SQE handed the kernel `addr op.acceptAddr` into a frame that is gone by completion time. Fill from the arena copy.
* `sockaddr_in` was a header-less importc without sin_zero, so nimony emitted its own 8-byte struct and every `listenTcp` died on `bind failed` (Linux rejects addrlen < 16). `sockaddr_storage` had no fields at all, so `sizeof` was 0 and accept never returned a peer address. Both are proper per-OS `{.pure.}` transcriptions now, machine-checked by tposixabi.
* `submitForPoll` armed the fd with the newest op's direction only, but EPOLL_CTL_MOD *replaces* the interest set: submitWrite then submitRead left the fd on EPOLLIN and the write parked forever. Arm with the union.
* `fdNotPollable` compared epoll_ctl's return value against EPERM/EBADF, but epoll_ctl returns -1 and reports the reason in errno, so it never matched and every failure printed the ADD+MOD error. Read errno.
* kqueue registered a `{.noinit.}` KEvent, passing garbage fflags/data/udata to the kernel; a stray NOTE_LOWAT makes the registration fire only after N garbage bytes. Zero-initialise it (KEvent.udata is `nil pointer` now so that is expressible).
* `threadIdx` is a threadvar defaulting to 0, so the main thread and worker 0 shared one unlocked slot arena and one single-producer io_uring SQ ring. Per-thread ring state is indexed by `ioLane()` now: workers keep their index, non-workers get a trailing lane.
* `initIoRing` was not idempotent although the module inits the ring at import time and the docs tell callers to call it -- re-running it under live workers is a use-after-free. Guarded like initDefaultRing was.
* `shutdown` closed the descriptors workers were still polling. Stop the pool first.
* `parWait` polled `poolPollIo`, which is only the do-nothing default; std/ioring installs the real poller into `gReactor`. A join with every chunk parked on I/O spun forever.
* `parSubmit` dropped its `hint`, so every chunk of a `||` loop piled into one stripe; the plugin passes the chunk index for exactly that reason.
* `initPool`'s `or workerCount > 0` let a second caller return before the CAS winner had allocated the pool seqs.
* `create workers[i], ..., i` passed the affinity request as `stackSize`.
* `poll_multi` set the multishot flag before `poll_add`, which zeroes `len` through `prepRw` -- the flag lives in `len`, so it was always dropped.
* `nextPow2` shifted by 32 unconditionally, UB on 32-bit targets.
* `kqueueClose` closed one lane where epollClose closes all.

closeFd still only cancels ops in the calling thread's lane; the docstring says so instead of claiming otherwise. Doing it properly needs a cross-lane cancel request plus IORING_OP_ASYNC_CANCEL, since the kernel owns the slot's buffers until it acknowledges.


lib/std/parfor.nim
```diff
@@ -136,7 +140,7 @@ proc parSubmit*(c: Continuation; hint = 0) {.inline.} =
   ## runners — but the spread still avoids needlessly caller-running chunks on
   ## the submitting thread and balances load. Re-exported so the `||` plugin
   ## only needs symbols visible through `import std/parfor`.
-  submit(c, -1)
+  submit(c, hint)
```
blackmius: it stucks thread on queue locks

```nim
import std/monotimes
import std/syncio
import std/threadpool
import std/parfor
initPool()

proc dfs(depth, breadth: int): uint32 =
  if depth == 0:
    return 1
  var sums = newSeq[uint32](breadth)
  for i in `||`(0, breadth-1, workload = CpuBound):
    sums[i] = dfs(depth-1, breadth)
  result = 0
  for i in 0..<breadth:
    result += sums[i]

var
  depth = 9
  breadth = 9

let start = getMonoTime().ticks
let answer = dfs(depth, breadth)
let stop = getMonoTime().ticks
echo "answer: ", answer
echo "time (s) ", (stop-start).float32 / 1_000_000_000.0

for i, wm in workerMetrics.pairs:
  echo "WorkerMetrics id=", i, ", enqueueMiss=", wm.enqueueMiss, ", dequeueMiss=", wm.dequeueMiss,
    ", tasksSubmited=", wm.tasksSubmited, ", tasksHandled=", wm.tasksHandled,
    ", tasksStealed=", wm.tasksStealed
```

```
daniil@ubuntu24:~/projects/nimony$ ./bin/nimony c -r bench_dfs.nim 
[100%] done
[100%] done
answer: 387420489
time (s) 12.536649703979492
WorkerMetrics id=0, enqueueMiss=0, dequeueMiss=202262738, tasksSubmited=114136978, tasksHandled=114351419, tasksStealed=76208833
WorkerMetrics id=1, enqueueMiss=0, dequeueMiss=324013, tasksSubmited=98163, tasksHandled=113866, tasksStealed=75975
WorkerMetrics id=2, enqueueMiss=0, dequeueMiss=140558232, tasksSubmited=31037112, tasksHandled=30816672, tasksStealed=20570350
```

and with hint=-1

```
daniil@ubuntu24:~/projects/nimony$ ./bin/nimony c -r bench_dfs.nim 
[ 54%] done
answer: 387420489
time (s) 5.578644275665283
WorkerMetrics id=0, enqueueMiss=0, dequeueMiss=120584962, tasksSubmited=145090471, tasksHandled=145080342, tasksStealed=96721386
WorkerMetrics id=1, enqueueMiss=0, dequeueMiss=257570, tasksSubmited=67923, tasksHandled=80564, tasksStealed=53989
WorkerMetrics id=2, enqueueMiss=0, dequeueMiss=523550, tasksSubmited=112383, tasksHandled=121379, tasksStealed=79643
```

PR #2393, opened 2026-08-28
Reject orphan generic params during overload resolution

### LINKED ISSUE #2392: Orphan generic params are not rejected during overload resolution
## Minimal Repro
```nim
func foo[Z: static[int]](x: float) = discard
func foo(x: float) = foo[1](x)

foo[1](1.0)  # OK
foo(1.0)     # Error: ambiguous call: 'foo'
```

## Expected Result
`foo(1.0)` does not specify type parameter `Z`, so the first overload should be rejected, and the second overload should remain the only viable candidate.

## Observed Result
The smecheck for `Z` is happening too late. `CouldNotInferTypeVar` is running only later in `buildTypeArgs` when `pickBestMatch` is already finished. Because both overloads match `x: float`, they both remain valid candidates.

gmpreussner:
Generic routines whose type parameters never appear in the callable signature (params or return type) were still treated as viable matches when called without explicit instantiation. That left unbound static or regular typevars until buildTypeArgs, after pickBestMatch had already tied them with non-generic overloads and reported "ambiguous call".

Reject such candidates at the end of sigmatch, and when every overload fails only for that reason, report the direct "could not infer type" diagnostic instead of the generic overload-set type mismatch wrapper.

Add `tests/nimony/overload/tstaticgenoverload.nim`

Fixes https://github.com/nim-lang/nimony/issues/2392

### DIFF
--- a/src/nimony/semcall.nim
+++ b/src/nimony/semcall.nim
@@ -1303,6 +1303,12 @@ proc resolveOverloads(c: var SemContext; dest: var TokenBuf; it: var Item; cs: v
         dest.addSubtree erroredN
         dest.addStrLit("", cs.callNodeInfo)
       return
+    elif m.len > 0 and allFailedDueToCouldNotInfer(m):
+      # Orphan typevars are rejected in sigmatch, so no candidate survives
+      # pickBestMatch; keep the direct "could not infer type" diagnostic instead
+      # of the generic overload-set "Type mismatch" wrapper.
+      buildErr c, dest, cs.callNodeInfo, getErrorMsg(m[0])
+      return
     elif m.len > 0:
       errorMsg = "Type mismatch at [position]\n"
       errorMsg.add asNimCode erroredN
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -2430,6 +2430,45 @@ proc matchTypevars*(m: var Match; fn: FnCandidate; explicitTypeVars: Cursor) =
       m.error0 RoutineIsNotGeneric
       return
 
+proc typevarInType(n: var TypeCursor; v: SymId): bool =
+  case n.kind
+  of Symbol:
+    if n.symId == v: return true
+    inc n
+  of TagLit:
+    n.loopInto:
+      if typevarInType(n, v): return true
+  else:
+    inc n
+  false
+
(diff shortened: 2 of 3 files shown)


gmpreussner:
## Implementation Notes

`matchFailedOnlyDueToCouldNotInfer`, `allFailedDueToCouldNotInfer` and the change in `semcall.nim` are only needed to satisfy the existing `tunderspecifiedgeneric.nim` unit test (the generic semcheck error message did not match the message that the test expected). I did not want to modify the existing test, so I'm forwarding the expected error type instead. This is probably the cleaner approach anyway, but confirm if you want this @Araq

src/nimony/sigmatch.nim
```diff
@@ -2467,6 +2508,16 @@ proc hasUnboundTypevars*(m: Match): bool =
...
+proc matchFailedOnlyDueToCouldNotInfer*(m: Match): bool =
+  result = m.err and m.error.kind == CouldNotInferTypeVar
+
+proc allFailedDueToCouldNotInfer*(m: openArray[Match]): bool =
+  if m.len == 0: return false
+  for i in 0..<m.len:
+    if not matchFailedOnlyDueToCouldNotInfer(m[i]):
+      return false
+  true
+
```
gmpreussner, line 2520: Only needed to satisfy existing tunderspecifiedgeneric.nim test

src/nimony/semcall.nim
```diff
@@ -1303,6 +1303,12 @@ proc resolveOverloads(c: var SemContext; dest: var TokenBuf; it: var Item; cs: v
         dest.addSubtree erroredN
         dest.addStrLit("", cs.callNodeInfo)
       return
+    elif m.len > 0 and allFailedDueToCouldNotInfer(m):
+      # Orphan typevars are rejected in sigmatch, so no candidate survives
+      # pickBestMatch; keep the direct "could not infer type" diagnostic instead
+      # of the generic overload-set "Type mismatch" wrapper.
+      buildErr c, dest, cs.callNodeInfo, getErrorMsg(m[0])
+      return
```
gmpreussner, line 1311: Only needed to satisfy existing tunderspecifiedgeneric.nim test

Araq:
Test is great but I want to try it myself, I have the vague feeling that it can be done with some re-routing of the existing logic instead.

Araq:
I did it my way.

PR #2395, opened 2026-08-29
ioring: let submitPollAdd say which direction it is waiting for

tokyovigilante:
submitPollAdd armed both directions unconditionally: armMaskForFd returns EvRead or EvWrite for opPollAdd, and the io_uring backend hardcodes poll_add(fd, POLLIN or POLLOUT). A connected socket is writable nearly always, so a caller waiting to READ completes immediately — and the op is oneshot, so its re-arm spins: 300 completions in 300 poll cycles on a socketpair with nothing ever written to it, mask EvWrite every time. libcurl's multi-socket engine, which this docstring names, asks for CURL_POLL_IN for a whole download.

So carry the direction on the op: arm exactly what was asked for, and complete only on `firedEvents and pollMask`. A wake for an unrequested direction leaves the slot pending, and the existing trailing re-arm keeps watching for the one it did request.

The mask defaults to both directions, so existing callers are unchanged — verified byte-identical at the default and 0 spurious with EvRead, on epoll and io_uring. tpolladd asserts the negative without a timeout: arm read-only on the writable-but-not-readable end first, then on the readable end; exactly one completion comes back and it must be the latter.

Also: report errno and the fd from the epoll ADD+MOD failure — epoll_ctl reports every failure as -1, so "failed: -1" says nothing.

### DIFF
--- a/lib/std/ioring/backends/epoll.nim
+++ b/lib/std/ioring/backends/epoll.nim
@@ -67,7 +67,12 @@ proc epollReArm(fd: cint; mask: int, alreadyRegistered: bool) {.nimcall.} =
       # connection. (A regular-file/closed fd is skipped above; MOD can't help it.)
       res = epoll_ctl(epollFd, EPOLL_CTL_MOD, fd, addr ev)
       if res != 0:
-        stderr.writeLine("ioring: epoll ADD+MOD both failed: " & $res)
+        # Report the ERRNO, not the return value: epoll_ctl reports every
+        # failure as -1, so "failed: -1" says nothing about which failure it
+        # was — and this branch fires only when both verbs failed on a fd we
+        # believe is live, i.e. exactly when the reason matters.
+        stderr.writeLine("ioring: epoll ADD+MOD both failed on fd " & $fd &
+                         " (errno " & $errno() & ")")
 
 proc epollPoll(timeoutMs: int): bool {.nimcall.} =
   let lane = ioLane()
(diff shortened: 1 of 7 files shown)


lib/std/ioring/backends/epoll.nim
```diff
@@ -67,7 +67,12 @@ proc epollReArm(fd: cint; mask: int, alreadyRegistered: bool) {.nimcall.} =
       # connection. (A regular-file/closed fd is skipped above; MOD can't help it.)
       res = epoll_ctl(epollFd, EPOLL_CTL_MOD, fd, addr ev)
       if res != 0:
-        stderr.writeLine("ioring: epoll ADD+MOD both failed: " & $res)
+        # Report the ERRNO, not the return value: epoll_ctl reports every
+        # failure as -1, so "failed: -1" says nothing about which failure it
+        # was — and this branch fires only when both verbs failed on a fd we
+        # believe is live, i.e. exactly when the reason matters.
+        stderr.writeLine("ioring: epoll ADD+MOD both failed on fd " & $fd &
```
Araq, line 74: Btw, the stdlib shouldn't just dump shit to `stderrr`, I have said it before...

tokyovigilante: You have, sorry. How do you feel about swapping it for an error hook eg `var ioringOnError*: proc(msg: cstring) {.nimcall.}`, dropping it, or leaving it alone?

Araq: IO operations can always fail anyway, not sure why these things are not `.raises` to begin with. Though maybe convenience demands it.

tokyovigilante: Fair, have had another go in #2401.

tokyovigilante: Fair, #2401 hopefully an improvement.

PR #2396, opened 2026-08-29
nif: replace cache files atomically, never truncate a mapped one

tokyovigilante:
The `.nif`/`.bif` writers replace a cache file in place while the readers mmap it. `vfs.writeBytes` (both backends) is `writeFile`, and `bif.store` opened the target `fmWrite` — both truncate. Meanwhile `nifreader.open` mmaps and parses straight out of the mapping, and `bif.load` mmaps a `.bif` and BORROWS its token block via `adoptForeignTokens`, deliberately keeping the mapping for the process lifetime.

`ftruncate` invalidates the pages of an existing mapping immediately, so a second process rewriting a cache entry another has mapped gives that reader SIGBUS past the new EOF, torn bytes inside the write window, or `EINVAL` from mmap on the zero-length file that exists between the truncating open and the first write. Those are the three shapes parallel builds fail in — `Bus error` with no location, a compile-time error on plainly valid code, and `cannot open: <hash>.s.nif` naming no source file — and each names a different innocent module every run, because the victim is whoever was reading the entry the two processes collided on. `nifmake -j:1` makes all three go away.

Fix: write a sibling temp file and rename it over the target. POSIX rename is atomic, and an open mapping keeps the old inode alive until it is unmapped, so a reader mid-parse finishes safely against the bytes it started with while the next opener sees the new file whole. The temp is a sibling (same filesystem, so the rename really is a rename), carries pid + a per-process counter so two concurrent writers cannot collide on the temp itself, and is not suffixed `.nif` so output scans never pick a half-written one up. It is removed on the failure path too.

`bif.store` now renders with `storeToString` and hands the bytes to `vfsWrite`, which costs nothing extra — `storeToFile` already materialised the whole image before writing — and routes `.bif` through the same relay as every other write. `storeToFile` stays for API compatibility, with a note that a caller who opens the target itself owns the atomicity problem, the truncation having already happened at its open.

Windows: `MoveFileEx` refuses to replace a file with an open mapping, so the rename fails loudly there. Not a regression — the truncating `fmWrite` open it replaces fails just as loudly against a mapped target (see the "keeps the file locked on Windows" note in `nifindexes.createIndex`), so the set of failing situations is unchanged and only the error text moves. On the platforms where the current code does not fail, it corrupts silently, which is worse. A bounded retry around the rename would absorb the Windows case; that is left out rather than guessing a timeout for a lock-holder the writer cannot identify.

tests/nifcache is a custom runner covering both writers. No race is needed: truncation invalidates the mapping at once, so one process suffices — store, mmap, store something smaller, then touch the mapping. Against the truncating writer both cases die with SIGBUS every time; with this change they read their original bytes back. It also asserts no `.tmp.*` residue.

### DIFF
--- a/src/lib/vfs.nim
+++ b/src/lib/vfs.nim
@@ -76,8 +76,32 @@ type
     AlwaysWrite,
     OnlyIfChanged
 
+# --- atomic replacement ---------------------------------------------------
+#
+# Never truncate a .nif or .bif that a reader may have mmap'd.
+
+import std / atomics
+
+when defined(windows):
+  proc osProcessId(): int32 {.importc: "GetCurrentProcessId", stdcall,
+                              dynlib: "kernel32".}
+else:
+  proc osProcessId(): int32 {.importc: "getpid".}
+
+when defined(nimony):
+  var atomicWriteCounter: int = 0
+  proc nextTempSeq(): int = atomicFetchAdd(atomicWriteCounter, 1)
+else:
+  var atomicWriteCounter: Atomic[int]
+  proc nextTempSeq(): int = atomicWriteCounter.fetchAdd(1)
+
+proc atomicTempPath*(target: string): string =
+  ## A sibling temp path for an atomic replacement of `target`
+  result = target & ".tmp." & $int(osProcessId()) & "." & $nextTempSeq()
+
 when defined(nimony):
   import std / [os, dirs, paths]
+  import std / private / oscommons
   # Nimony's stdlib procs are `.raises`. Wrap them so vfs.nim stays
   # non-raising (the rest of the compiler is wired that way).
   proc unixModTime(p: string): int64 =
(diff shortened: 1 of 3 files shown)


src/lib/vfs.nim
```diff
@@ -76,8 +76,50 @@ type
...
+# `fmWrite` open it replaces fails just as loudly against a mapped target
+# (see the "keeps the file locked on Windows" note in `nifindexes.createIndex`)
+# — and a loud failure is in any case strictly better than the silent
+# corruption the POSIX truncation produces.
+
+when defined(windows):
+  proc osProcessId(): int32 {.importc: "GetCurrentProcessId", stdcall,
+                              dynlib: "kernel32".}
+else:
+  proc osProcessId(): int32 {.importc: "getpid", header: "<unistd.h>".}
```
Araq: We want to be libc-free.

src/lib/vfs.nim
```diff
@@ -76,8 +76,50 @@ type
...
+var atomicWriteCounter: int = 0
+
+proc atomicTempPath*(target: string): string =
+  ## A temp path for an atomic replacement of `target`. It is a SIBLING of the
+  ## target (same directory, hence same filesystem, hence the rename really is
+  ## a rename) and carries the pid plus a per-process counter, so two
+  ## concurrent writers of one cache entry cannot collide on the temp itself.
+  ## The suffix is deliberately not `.nif`, so directory scans that look for
+  ## outputs never pick a half-written temp up.
+  inc atomicWriteCounter
```
Araq: How about you use an atomcInc for your atomic counter?

tokyovigilante: That would be too sensible... updated.

PR #2399, opened 2026-08-29
strutils: formatFloat without snprintf on wasm

tokyovigilante:
WebAssembly has no variadic calling convention. An imported function has one type, so a `{.varargs.}` importc cannot be called with different argument tails, and `c_snprintf(buf, n, fmt, precision, f)` emits a call whose f64 argument contradicts the import's declared i32. The engine rejects that when it compiles the MODULE, not when it reaches the call — so a single `formatFloat` on a diagnostic path takes the whole program down. It is not hypothetical: it is why a browser viewer built on this would not load at all.

The wasm arm formats from the digits `$f` already produces. Rounding works on the decimal string rather than scaling by a power of ten, because `f * 10^p` introduces an error that decides the last digit at exactly the values where it matters.

The one case digits alone cannot settle is decided against the value. `$f` is the SHORTEST round-tripping decimal, so a trailing `5` is not necessarily an exact tie: 9.995 prints as "9.995" while the double is 9.99499999999999957, which C rounds DOWN at two places. A first cut rounded it up on the strength of the printed digits, and the differential fixture caught it. The arm now builds both candidates, parses them back, and keeps the nearer — a genuine tie leaves them equidistant and goes half-to-even, as C does.

Two things this costs, both measured rather than assumed:

A precision beyond the shortest form's digit count pads with zeros where glibc continues the exact binary expansion (`0.1` at 20 places is `0.10000000000000000000` here, `0.10000000000000000555` there). Every precision the shortest form covers agrees byte for byte, which is every display use.

And it adds one host import. The module's env imports become `nim_write`, `nim_exit` and `strtod` — the last from parsing the candidates back. That trades a variadic import that CANNOT be typed on wasm for a non-variadic one that can, and it is the same `strtod` any wasm program that parses a float already needs; but it is a new requirement for one that only formats, so it is stated here rather than discovered.

The C path is untouched — everything above is inside a new `when defined(wasm32) and defined(standalone)` arm.

Verified on this branch: the glibc leg and the armed leg render the grid byte-identically, 27/27 lines, built from one fixture with `nimony c` and `nimony w` and diffed. `hastur build all` clean, `tests/nimony/stdlib` 53/53, `hastur all` 761/761.

### DIFF
--- a/lib/std/strutils.nim
+++ b/lib/std/strutils.nim
@@ -1122,7 +1122,8 @@ when defined(wasm32) and defined(standalone):
         dec i
       if carry:
         result.carry = true
-        kept = "1" & kept.substr(0, kept.len - 2)
+        kept.shrink kept.len - 2
+        kept = "1" & kept
     result.digits = kept
 
   func fmtDecimal(f: BiggestFloat, neg: bool, digits: string,
--- a/lib/std/strutils.nim
+++ b/lib/std/strutils.nim
@@ -966,8 +966,9 @@ func unescape*(s: string, prefix = "\"", suffix = "\""): string {.raises.} =
   if not s.endsWith(suffix):
     raise ValueError
 
-func c_snprintf(buf: cstring, n: csize_t, frmt: cstring): cint {.
-  header: "<stdio.h>", importc: "snprintf", varargs.}
+when not (defined(wasm32) and defined(standalone)):
+  func c_snprintf(buf: cstring, n: csize_t, frmt: cstring): cint {.
+    header: "<stdio.h>", importc: "snprintf", varargs.}
 
 type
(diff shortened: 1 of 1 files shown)


lib/std/strutils.nim
```diff
@@ -976,6 +977,219 @@ type
...
+      while i >= 0 and carry:
+        if kept[i] == '9':
+          kept[i] = '0'
+        else:
+          kept[i] = char(int(kept[i]) + 1)
+          carry = false
+        dec i
+      if carry:
+        result.carry = true
+        kept = "1" & kept.substr(0, kept.len - 2)
```
Araq:
```suggestion
        kept.shrink kept.len - 2
        kept = "1" & kept
```

PR #2401, opened 2026-08-29
ioring: stop printing, fail the ops instead

tokyovigilante:
Removes the three `stderr.writeLine`s in `std/ioring`. Two of them were noise; the third was standing in for an error path that did not exist.

`iouring`'s "failed to get sqe" already re-enqueues the dequeued ops so the next poll picks them up — and the `sqe == nil` branch four lines below recovers identically without announcing it, so the module was inconsistent with itself. "failed to init io_uring queue" likewise: the caller falls back to the epoll backend, and that fallback IS the handling. Both prints just go.

`epoll`'s "ADD+MOD both failed" is the different one. There is no recovery there: the fd ends up unarmed, so nothing will ever make its pending ops ready, their continuations stay parked, and the caller waits on a read that will never return and never fail. No error, no timeout, no crash — just a task that never wakes while the process runs on normally. That message was the only evidence the failure had happened at all.

So report it instead. `reArmEvent` returns whether the fd is armed, and `submitForPoll`/`processFd` complete that fd's pending ops with `ArmFailed` when it is not — the same thing `closeFd` already does when it cancels, and the same value a failed read or write reports, so callers need no new case.

Why completion and not `raise`: the arm runs inside `poll`, under `gReactor`, so the throw would land on whichever worker (or whichever embedder frame calling `pumpIo`) happened to be polling — a thread that owns none of the affected ops and can do nothing with it. It would not release them either, the slot's continuation being parked regardless, so the completion is needed anyway. And since `submit` is caller-runs under saturation, user continuations execute inside `processFd`, so an exception would unwind through those frames. The per-op result already reaches the owner; this path simply was not using it.

Two behaviour changes worth calling out. A MOD that fails on an already-registered fd was previously ignored outright — same silent stall, now reported. And the EPERM/EBADF "not pollable" case is no longer skipped silently: skipping was safe only while nothing was waiting on the fd, and `submitForPoll` runs immediately after `allocSlot`, so something always is.

kqueue's registration failures were likewise discarded; it now reports them.

Verified: `hastur build all` clean, `tests/nimony/threads` 10/10 on io_uring, `tpolladd` matches its golden on the epoll arm too, `hastur all` 761/761.

### DIFF
--- a/lib/std/ioring/backends/iouring.nim
+++ b/lib/std/ioring/backends/iouring.nim
@@ -10,7 +10,7 @@
 # polled" hang.
 
 import std/[assertions, atomics, posix/posix, ticketlocks, threadpool]
-import std/syncio
+import std/syncio   # quit
 import ../../posix/io_uring
 import ../core/types
 import ../core/slots
@@ -33,9 +33,8 @@ proc tryInitLocalQueues(): bool =
   try:
     for i in 0..<ioLanes():
       localQueues.add newQueue(sqEntries)
-  except ErrorCode as e:
-    stderr.writeLine("ioring: failed to init io_uring queue: " & $e)
-    return false
+  except ErrorCode:
+    return false   # the caller falls back to the epoll backend
   return true
 
 proc fillSqe(sqe: ptr Sqe; op: ptr OpContext) {.inline.} =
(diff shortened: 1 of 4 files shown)


lib/std/ioring/backends/iouring.nim
```diff
@@ -33,9 +33,8 @@ proc tryInitLocalQueues(): bool =
   try:
     for i in 0..<ioLanes():
       localQueues.add newQueue(sqEntries)
-  except ErrorCode as e:
-    stderr.writeLine("ioring: failed to init io_uring queue: " & $e)
```
blackmius, line 37: why suppressing this info?

tokyovigilante: Writing to stderr from stdlib considered harmful... - https://github.com/nim-lang/nimony/pull/2395#discussion_r3885376918

But this seems a bit more helpful than the other log that just recovers next poll.  I'm actually getting this message now on my laptop with 7.2, which should work fine with io_uring, so maybe it should stay in.

PR #2402, opened 2026-08-29
ioring: wait for completions in the kernel, not in a nanosleep

tokyovigilante:
`iouringPoll` accepts a `timeoutMs` and ignores it: it drains the op queue, submits, then harvests only CQEs that are ALREADY ready and returns. A caller that asks for a bounded wait therefore gets a busy loop.

It shows up most plainly for an embedder driving the ring from its own loop — ours pumps from a render frame callback, because its main thread belongs to a rendering engine and can never call the pool's own loop:

    pumpIo(1) over ~92ms      io_uring as-is:  3,509,132 iterations
                              with this patch:            93

One per millisecond, as asked, rather than a spinning core. The epoll backend was already correct here, because its wait IS `epoll_wait(timeoutMs)`.

`submitAndWait` fuses the flush and the wait into ONE `io_uring_enter` with `IORING_ENTER_GETEVENTS | IORING_ENTER_EXT_ARG` and a timespec — the shape of liburing's `io_uring_submit_and_wait_timeout`. Submitting and then entering again to wait would be two syscalls per poll iteration on the path being fixed. `iouringPoll` calls it unconditionally, so a poll with nothing new to submit still spends its budget waiting on work already in flight.

### DIFF
--- a/lib/std/ioring/backends/iouring.nim
+++ b/lib/std/ioring/backends/iouring.nim
@@ -92,10 +92,21 @@ proc iouringPoll(timeoutMs: int): bool {.nimcall.} =
       # `addr op.acceptAddr`/`addr op.acceptLen` and the kernel writes through
       # those at completion time, long after this stack frame is gone.
       fillSqe(sqe, addr gSlots[lane].slots[idx].op)
-    try:
-      discard localQueues[lane].submit()
-    except ErrorCode as e:
-      quit "fatal: bug: submit cannot fail: " & $e
+  # ONE enter per poll: flush whatever was just filled and, when the caller
+  # gave us a time budget, wait for a completion inside it.
+  #
+  # Waiting here is the whole point. Peeking at the CQ and returning empty
+  # leaves the caller to spend its budget in `nanosleep` (see
+  # `threadpool.workerLoop`) — a sleep no completion can interrupt, so every
+  # completion picks up latency bounded by the poll interval rather than being
+  # delivered when the kernel has it. epoll never had that problem because its
+  # wait IS `epoll_wait(timeoutMs)`. Measured on a write-heavy WebSocket
+  # workload before this: ~3.2x the control-reply latency of the epoll backend,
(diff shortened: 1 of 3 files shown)


blackmius:
there is a problem that worker blocks waiting cqes from its loop, when it could steal other workers tasks

upd: ok i see it uses timeout, then it should not do nanosleep after that it already sleep enough

easy fix is to make io_uring backend poll always return true if sleep handled inside

lib/std/ioring/backends/iouring.nim
```diff
@@ -95,10 +95,21 @@ proc iouringPoll(timeoutMs: int): bool {.nimcall.} =
...
+  # `threadpool.workerLoop`) — a sleep no completion can interrupt, so every
+  # completion picks up latency bounded by the poll interval rather than being
+  # delivered when the kernel has it. epoll never had that problem because its
+  # wait IS `epoll_wait(timeoutMs)`. Measured on a write-heavy WebSocket
+  # workload before this: ~3.2x the control-reply latency of the epoll backend,
+  # uniformly across rates.
+  #
+  # Called unconditionally, not just when SQEs were filled: with nothing new to
+  # submit there is still a budget to spend waiting for work already in flight.
+  discard localQueues[lane].submitAndWait(1'u, timeoutMs.int64 * 1_000_000'i64)
```
blackmius: also you dont need to call submit if your sqe queue is empty. you can check sqeReady or something in posix/io_uring that there is sqe which is in-flight

tokyovigilante: Good spot, should be fixed now.

tokyovigilante:
Superseded by #2425 / the Pr2427 merges: iouringPoll now waits in the ring with a per-lane deadline, BackendRelays.waits retires the double nap, and an idle lane with a budget waits in the ring (our commit 3's end state). Nothing outstanding after rebasing onto 2ded3517.

Issue #755, opened 2025-03-12
Generic constraints not enforced


Araq:
This program compiles but shouldn't:

```nim
import std / [syncio]

type
  BinaryTree[T] = ref object
    le, ri: BinaryTree[T]
    data: T

proc newNode*[T](data: sink T): BinaryTree[T] = BinaryTree[T](data: data)

proc append*[Ty](root: var BinaryTree[Ty], n: BinaryTree[Ty]) =
  # insert a node into the tree
  if root == nil:
    root = n
  else:
    var it = root
    while it != nil:
      var c = cmp(n.data, it.data)
      if c < 0:
        if it.le == nil:
          it.le = n
          return
        it = it.le
      else:
        if it.ri == nil:
          it.ri = n
          return
        it = it.ri
```

As the `Ty` is not declared to be `Comparable` the call to `cmp(n.data, it.data)` is actually invalid.

Araq:
Maybe but then the compiler is picky (and correctly so!) in other cases.

gmpreussner:
This is not fixed yet

Minimal repro:
```
type
  Comparable = concept
    func `==`(a, b: Self): bool
    func `<`(a, b: Self): bool
    func `>`(a, b: Self): bool

  Foo = object
    x: int

func `==`(a, b: Foo): bool = true

type Box[T: Comparable] = object
  v: T

var x: Box[Foo]
```

`Foo` is accepted even though it does not implement `<` and `>`.

I think the problem is that we still have this stub in `sigmatch.nim` `matchConceptBody`:
```
  # Until concrete-type requirement matching is complete, standalone concepts
  # match any concrete type (legacy stub behaviour). Concept-to-concept
  # subsumption always checks requirements structurally.
    # An unconstrained typevar reaches us as an empty (`.`) constraint: it
    # provably fulfils no requirement, so it must not satisfy the concept
    # (issue #755). Genuine concrete types stay leniently accepted.
    return a.kind != DotToken
```

gmpreussner:
I can't seem to fix this without breaking https://github.com/nim-lang/nimony/blob/master/tests/nimony/concepts/tconceptiterator.nim

@Araq can you check whether `tconceptiterator.nim` is actually valid syntax? `array[2, int]` indexes with `0..1` while the iterator is defined for `int`. That only works with the stub.

Araq:
tconceptiterator.nim is valid afaict.

gmpreussner:
[...]

PR #2424, opened 2026-09-01
Fix false-positive concept satisfaction for standalone user types

### LINKED ISSUE #2423: Concept always satisfied for every type if body requires `typedesc[Self]` procs
## Minimal repro
```nim
type
  Foo = concept
    proc trait(_: typedesc[Self]): Self

type Plain = object
  x: int

when Plain is Foo:
  {.error: "BUG: Plain satisfies Foo but has no trait proc".}
```

## Expected

Compilation succeeds (Plain has no matching `trait` proc).

## Actual

Compilation fails with the custom error message.

gmpreussner:
Standalone concepts (no `of` parent) used a legacy shortcut in `matchConceptBody`: any concrete type that was not `distinct` was accepted without checking whether its requirements were actually satisfied. Only `distinct` types went through structural checking via `conceptTargetNeedsStrictCheck`.

That stub made `T is Concept` a false positive for most user-defined types. In particular, concepts whose bodies require `typedesc[Self]` procs — such as ordinal traits using `ordinalCardinality(_: typedesc[Self])` and `fromOrdinal(_: typedesc[Self], …)` — reported satisfaction even when no matching procs existed. A plain `object`, a sparse `enum`, or even `string` would pass checks like `Plain is HasTrait` while `compiles(Plain.trait())` remained false. Generic instantiation still failed later, but negative `assert not (T is Concept)` tests and concept-gated APIs were wrong at the membership site.

The existing strict path already handles `typedesc[Self]` correctly when it runs: `conceptSelfSyms` collects `Self`, `conceptRoutineAvailable` binds `m.inferred[selfSym] = a`, and `matchConceptRoutineSig` matches signatures. The bug was that this path was skipped for almost all concrete types.

Expand `conceptTargetNeedsStrictCheck` so standalone concepts structurally check requirements for user-defined types declared in the module currently being semchecked:

- `object`, `enum`, `holey enum`, and `distinct` in `c.thisModuleSuffix`
- exclude `system` types (`SystemModuleSuffix`) and types from other modules

Types outside that set — built-in scalars, `string`, stdlib types like `Table`, compiler-internal types — keep legacy lenient matching until generic-candidate constraint verification is complete (see `tdollar_char_loophole`). This avoids regressions in table/keyed APIs, `Stringable`/`Keyable` use of `string`, and concept-check blowup tests.

Pass `SemContext` into `conceptTargetNeedsStrictCheck` from `matchConceptBody` so module ownership can be tested. Update comments in `matchConceptBody` to document the remaining lenient path and its relation to issue #755 (dot-token constraints) and the dollar-char loophole.

Tests:

- `tconcepttypedescfalsepositive.nim`: negative `is` check for standalone concept with `typedesc[Self]` requirement on a plain object
- `tconceptstandalonebadobject.nim`: compile-fail when a plain object is used where a standalone `Addable` constraint is required (extends coverage beyond `tconceptstandalonebad.nim`, which only exercised `distinct`)

Fixes https://github.com/nim-lang/nimony/issues/2423


gmpreussner:
@Araq This is a minimally invasive bandaid.

The correct fix would be to refactor this and completely remove the legacy stub (see also Issue #755). I tried to do this myself again but didn't get any acceptable results. I think it will require your domain knowledge to do this properly.

PR #2425, opened 2026-09-01
http modules for the stdlib

Araq:
NIF-based design by me. No callbacks.


doc/internals/http.md
```diff
@@ -0,0 +1,552 @@
...
+- **The scheduler can use them.** Deadlines visible to the runtime let the
+  ready queue be earliest-deadline-first rather than FIFO, on both the ring's
+  completion side and the pool's run queue.
+- **Admission control becomes possible.** A loop that is behind can refuse a
+  connection rather than accept work it cannot finish in budget — but only if
+  the budget is visible at accept time.
+
+### Mapping down
+
+The kernel already speaks absolute deadlines. `TIMEOUT_ABS` and `link_timeout`
```
blackmius: usually it is better to have a timeouts heap and single timeout registered in kernel to nearest deadline and then scalp the heap top, same as for epoll/kqueue but let you avoid clock_gettime sysscall with io_uring, without heap you likely will clog io_uring queues with timeouts

doc/internals/http.md
```diff
@@ -0,0 +1,552 @@
...
+
+## 4. Connections and threading
+
+`ConnId` is a dense integer with a generation counter, like ioring's slot
+arena — not a pointer. That keeps `HttpEvent` POD and copyable, and makes a
+stale reference *detectable* rather than a use-after-free, which matters
+because connections die asynchronously and handlers will hold ids across
+suspensions.
+
+**One loop per thread, and a connection belongs to its lane for life.** ioring's
```
blackmius, line 385: to utilize that pinning we should bind 1 cpu core to 1 NIC queue

doc/internals/http.md
```diff
@@ -0,0 +1,552 @@
...
+   1. Entries are not removed when an op completes in time; they name the slot
+   *generation* they were armed for and are dropped as stale when they surface.
+   `submitTimeout` is a pure timer, and reaching its deadline is its success.
+
+   Still to do here: io_uring can attach an absolute `link_timeout` to an
+   individual SQE, which is better than one lane-wide bound. The lane-wide
+   bound keeps both backends behaving identically in the meantime.
+3. Multishot accept. `submitAccept` is oneshot, so a busy listener pays one
+   submission per connection; io_uring has `accept_multishot`.
+4. DNS. `getaddrinfo` blocks and io_uring has no resolver, so it needs
```
blackmius: or write dns resolver in nim, it is just one udp datagram out one udp datagram in + few parses for hosts config  (/etc/hosts /etc/resolver)

doc/internals/http.md
```diff
@@ -0,0 +1,552 @@
...
+valid on the other side. `result` has to be assigned before its address is
+taken, since the compiler will not hand out the address of something it cannot
+prove is initialised.
+
+The entry point matters, and it is the §8 hazard in practice: a chain is
+started with `submit(delay(chain()), lane)` onto the pool, not called from
+non-passive code. A regular proc that calls a parking passive proc gets its
+`result` written into a frame it has already left.
+
+### A compiler bug this layer flushed out (fixed)
```
blackmius, line 615: claude documented commit history

doc/internals/http.md
```diff
@@ -0,0 +1,552 @@
...
+  parser stores one line as one value — splitting only some headers on commas
+  (never `Date`, never `Set-Cookie`) is a decision worth making deliberately.
+- Structured header values — `Content-Type`'s parameters, `Accept`'s q-values —
+  are stored as one string today. They are the natural next thing to give
+  sub-structure to, and the tree already has room for it.
+- Client-side correlation: a `ReqId` returned by `send`, matched by
+  `ResponseEvent`, so pipelined or multiplexed requests can be told apart.
+- Whether one `HttpLoop` type serves both directions or client and server get
+  their own.
+- Whether cancellation stays a runtime concern that ioring implements, or wants
```
blackmius, line 734: it definitely needs language-level support, runtime can only close OS resources, user will need to close their app-specific. but at least now no one calls close except timeouts and they can be catched via try/except (it is good because error propagates from bottom to top) if at some point it will requires to give a user .close() lever it will be a problematic without language-support to correctly pass close to the last child so it again can be handled via try/except

doc/internals/http.md
Araq: sure, it's all work-in-progress.

lib/std/ioring.nim
```diff
@@ -221,6 +264,41 @@ when defined(posix):
     else:
       result = (x shl 8) or (x shr 8)
 
+  proc socketNonBlocking*(): cint =
```
blackmius, line 268: io_uring has op_socket and op_cmd_sock for setsockopt syscalls

doc/internals/http.md
blackmius: the problem here is that then every passive can raise CancelError and its signature should always have raises pragma

Araq: Changing the spec so that `.passive` always implies `.raises` is fine with me, but a last resort.

lib/std/http/httpconn.nim
```diff
@@ -0,0 +1,374 @@
...
+
+type
+  ChunkState = enum
+    csHeader      ## at a chunk-size line
+    csData        ## inside a chunk's data, `chunkLeft` bytes to go
+    csDataEnd     ## at the CRLF that closes a chunk's data
+    csTrailer     ## after the zero chunk, in the trailer section
+    csDone        ## the body has ended
+
+  HttpConn* = object
```
blackmius, line 41: claude was writing about some TcpTransport relays

blackmius: maybe better to add std/asyncstreams with Stream based on vtables and concepts Readable Writable Seekavle and then create TcpConn object of DuplexStream
and only then HttpConn based on it

lib/std/http/httpmsg.nim
```diff
@@ -0,0 +1,641 @@
...
+# The vocabulary is sealed and smaller than 512, so looking a spelling up
+# does not need the pool's hash table — and must not, because `getKeyId`
+# takes a `string` and building one per header per request is exactly the
+# allocation this layer exists to avoid. Bucketing the tags by name length
+# leaves a handful of candidates per bucket, which a byte compare settles.
+
+var
+  gByLen: seq[TagId] = @[]
+  gLenStart: array[MaxTagNameLen + 2, int32]
+  gIndexedUpTo = 0
```
Araq: What? In what world is using global variables for this a good idea?

lib/std/http/httpmsg.nim
```diff
@@ -0,0 +1,641 @@
...
+  result = c.sub()
+  if c.cursorTagId == tag(tReq):
+    result.skip                  # (METHOD)
+    result.skip                  # target
+    result.skip                  # (version)
+  else:
+    result.skip                  # status
+    result.skip                  # (version)
+
+iterator headers*(m: var HttpMsg): TagId =
```
Araq: Why is this a `var HttpMsg`?

lib/std/http/httpparse.nim
```diff
@@ -0,0 +1,464 @@
...
+# ------------------------------------------------------------- primitives --
+
+proc isTokenChar(c: char): bool {.inline.} =
+  ## RFC 9110 `tchar`. Notably excludes the separators, so a header name
+  ## cannot contain a colon, space or control character.
+  case c
+  of 'a'..'z', 'A'..'Z', '0'..'9',
+     '!', '#', '$', '%', '&', '\'', '*', '+', '-', '.',
+     '^', '_', '`', '|', '~': true
+  else: false
```
Araq: Use a set, this is Nim.

lib/std/http/httpparse.nim
```diff
@@ -0,0 +1,464 @@
...
+  ## here, and it is a response we are reading rather than one we are serving.
+  var j = i
+  while j < buf.len:
+    let c = buf[j]
+    if c == '\r' or c == '\n': break
+    if c != '\t' and (c < ' ' or c == '\x7F'): return ParseBad
+    inc j
+  if j >= buf.len: return ParseIncomplete
+  last = j
+  result = j
```
Araq: Huh? So what is the point of this `last: var int` parameter?

lib/std/http/httpparse.nim
```diff
@@ -0,0 +1,464 @@
...
+
+proc parseStatusLine*(buf: openArray[char]; i: int; m: var HttpMsg): int =
```
Araq: `buf: openArray[char]; i: int;` is redundant. Here and everywhere else. If you use openArray, embrace it and use its toOpenArray()/whatever to slice it in O(1)

lib/std/http/httpparse.nim
```diff
@@ -0,0 +1,464 @@
...
+    ## that. It bounds the *line*, so a peer cannot send a megabyte of `0`.
+  MaxChunkExtLen* = 256
+    ## Chunk extensions are parsed only to be skipped; nothing reads them.
+  MaxTrailerCount* = 16
+
+proc hexDigit(c: char): int {.inline.} =
+  if c >= '0' and c <= '9': int(ord(c) - ord('0'))
+  elif c >= 'a' and c <= 'f': int(ord(c) - ord('a')) + 10
+  elif c >= 'A' and c <= 'F': int(ord(c) - ord('A')) + 10
+  else: -1
```
Araq: Use a `case` statement. This is Nim.

lib/std/http/httpparse.nim
```diff
@@ -0,0 +1,464 @@
...
+    if buf[j] != ';': return ParseBad
+    var stop = 0
+    let e = parseUntilEol(buf, j, stop)
+    if e < 0: return e
+    if stop - j > MaxChunkExtLen: return ParseBad
+    j = stop
+  let after = parseCrLf(buf, j)
+  if after < 0: return after
+  size = n
+  result = after
```
Araq: So in other words, you might as well have used `result = parseCrLf(buf, j)`...

lib/std/http/httpwire.nim
```diff
@@ -0,0 +1,330 @@
...
+  dest[i + 1] = '\n'
+  result = i + 2
+
+proc digitCount*(x: int): int =
+  ## Digits in the decimal form of a non-negative `x`.
+  result = 1
+  var n = x
+  while n >= 10:
+    n = n div 10
+    inc result
```
Araq, line 56: Seriously? And then you use that in writeInt which does this slow `div` loop all over again?

lib/std/http/httpwire.nim
```diff
@@ -0,0 +1,330 @@
...
+
+proc chunkOverhead*(size: int): int =
+  ## Bytes `writeChunkHeader` + `writeChunkEnd` add around `size` bytes of
+  ## data, so a caller can size a buffer without writing twice.
+  var digits = 1
+  var v = size
+  while v >= 16:
+    v = v shr 4
+    inc digits
+  result = digits + 2 + 2
```
Araq: Use `result` directly and initialize it with `5`...

lib/std/ioring/backends/epoll.nim
```diff
@@ -70,11 +74,26 @@ proc epollPoll(timeoutMs: int): bool {.nimcall.} =
...
+        continue          # nothing to arm on: the deadline heap is the wait
+      if buf[i].kind == opNop:
+        complete(idx, 0)  # nothing to wait for either
+        continue
+      if buf[i].kind == opConnect:
+        # Start the attempt here, on the polling thread, so the fd is already
+        # connecting by the time we watch it. A connect that finished at once
+        # has completed the slot and there is nothing left to arm.
+        if not startConnect(buf[i].fd, idx):
+          continue
```
Araq: At what point are you gonna acknowledge agents.md ?

lib/std/posix/io_uring.nim
```diff
@@ -526,6 +526,20 @@ const
...
+# spells out at length: under the default `-d:nimNativeIo` it reads a
+# module-local variable that only posix.nim's own freestanding wrappers
+# maintain, while every call below — `syscall`, `mmap`, `munmap` — is a bare
+# `importc` into libc and sets *libc's* errno. Reading the wrong one is not a
+# wrong message, it is a wrong control flow: every failure here reported `0`,
+# which maps to `Success`, so `raiseOSError` raised nothing at all and
+# `setup`'s `-1` fd travelled on into `mmap`. `newRing` then wrote through the
+# `MAP_FAILED` it got back, and the ioring init that is supposed to fall back
+# to epoll segfaulted instead.
+proc errnoLocation(): ptr cint {.importc: "__errno_location", sideEffect.}
```
Araq: But we don't want to use any libc, we use the Linux kernel directly which thankfully has no errno bullshit whatsoever at all, instead functions return error codes, how amazing is that...

tests/nimony/http/httptags.nim
```diff
@@ -0,0 +1,18 @@
...
+# tests mean.
+#
+# An imported module initialises once, before any of its importers, so
+# registering here makes the group independent of that order. It is also the
+# shape the design prescribes for an application: register the headers you
+# index on in one place during init, then seal.
+import std/http/httpmsg
+
+let hTrace* = registerHeader("x-trace-id")
+sealHttpTags()
```
Araq: An entire new module for these 3 lines? This is stupid. What even is x-trace-id anyway, never heard of it.

Araq:
Merging it now cause it contains useful fixes, will clean it up further later.

PR #2436, opened 2026-09-03
sem: generic `.passive` procs that suspend or delay-spawn

tokyovigilante:
Branch: fork/pr/async-generic-passive (3 fixes + 1 tests commit, base 8270a152)

Generic `.passive` procs fail on instantiation re-sem today:

- a generic `.passive` proc that calls `suspend()` is rejected with
  "expression of type `Continuation` must be discarded": first-pass sem
  types the magic as void, the instantiation re-sem routed it through
  semSuspend, which typed it Continuation.
- a generic `.passive` proc whose body does `delay(call)` is rejected with
  "`delay` takes a call expression or no argument": the first pass already
  flattened the call into `(delay fn args)`, and re-sem did not recognise
  the flattened form; the delayed callee was also never instantiated for
  the current type args.

Fixes: semSuspend types the call as void; semDelay is idempotent (a
sole-child probe keys the unflattened form, a flattened form copies its
children through) and re-sems the delayed callee so it gets instantiated.
The same commit keeps the compiler's own `src/lib` on the search path
unconditionally so macro-plugin sub-compiles agree with the outer build on
module suffixes outside the repo, and adds an on-disk fallback for macros
absent from the importer's compiledMacros so imported proc-pragma macros
work.

Adapted from aoughwl/nimony (author savannt, MIT), with attribution in the
commit messages. Three tests in tests/nimony/cps.


src/nimony/macro_plugin.nim
```diff
@@ -8,7 +8,9 @@
 ## `nimony s` (which picks up `.p.nif` input — same machinery `executeExpr`
 ## uses for CTFE), and exec the resulting binary at every call site.
 
-import std/[syncio, os, osproc, tables, hashes, assertions]
+import std/[syncio, os, osproc, tables, hashes, assertions, strutils]
+import std/[dirs, paths]
+include ".." / lib / compat2
```
Araq: All unrelated changes that have nothing to do with this PR.

Araq:
Clean it up so that the PR does only what you claim it does.

Araq:
My bad: I meant to undo:

> The same commit keeps the compiler's own src/lib on the search path
unconditionally so macro-plugin sub-compiles agree with the outer build on
module suffixes outside the repo, and adds an on-disk fallback for macros
absent from the importer's compiledMacros so imported proc-pragma macros
work.

tokyovigilante:
Ok, just the three sem fixes now.

Araq:
[nif ci / windows-amd64-nim-devel (master) (pull_request)](https://github.com/nim-lang/nimony/actions/runs/33747149281/job/100622205337?pr=2436) was stuck, let's see if it works now.

tests/nimony/cps/tpassive_generic_suspend.nim
```diff
@@ -0,0 +1,14 @@
...
+# as Continuation and the discard check rejected it.
+var log = ""
+proc worker[T](x: T) {.passive.} =
+  log.add "w"
+  suspend()
+  log.add "W"
+proc driver() {.passive.} =
+  let c = delay(worker(3))
+  complete(c)
+  complete(c)
```
Araq:
```suggestion
```

tests/nimony/cps/tpassive_generic_suspend.output
```diff
@@ -0,0 +1 @@
+wWwW
```
Araq:
```suggestion
wW
```

PR #2437, opened 2026-09-03
hexer: lower foreign closure decls at load time

tokyovigilante:
Branch: fork/pr/crossmodule-closure-abi (1 commit, base 8270a152)

A module that uses a closure declared elsewhere loads the symbol's decl as
sem wrote it (a `.closure` proctype), while the owner already lowered it to
the {fn, env} tuple. The consumer emits a raw function pointer against a
tuple definition: wrong code, no diagnostic.

The fix adds `programs.declLoadTransformer`, a hook applied to every decl
`tryLoadSym` parses from a foreign index. hexer installs `canonForeignDecl`
there, which lowers closure types in the decl the way lambdalifting lowers
local ones (closure tuple plus env param). sem and the other frontends leave
the hook nil; bodies and generic decls are untouched.

Two tests: a hook variable assigned in one module and called through a
template from another, and a call through a foreign object's closure field
from a module that declares no closure itself.


Araq:
Wrong idea, the invariant we seek is:

Hexer uses the semchecked results for type queries. These keep the `.closure` and `.passive` and `.raises` annotations and are precise. Only forged symbols, introduced in Hexer's own passes are problematic and require a publish mechanism that is compatible with semchecked/Nimony's type system.

PR #2443, opened 2026-09-03
Fix overload resolution to prefer candidates with fewer unbound generic parameters

### LINKED ISSUE #2442: Overload resolution fails when several overloads share unconstrained type param and generic return value
I stumbled upon an odd overload resolution failure that goes away when introducing concepts.

## Minimal Repro

Compile (should fail):
```
nimony c foo.nim
```

Compile with workaround:
```
nimony c -d:fix foo.nim
```

```nim
const broken = not defined(fix)

type
  Matrix*[R, C: static[int], T] = object
    x: int


when broken:
  # Error: could not infer type for R.2.foorbtsyq
  proc matrix*[R, C: static[int], T](elems: array[R, array[C, T]]): Matrix[R, C, T] {.noinit.} = discard
  proc matrix*[R, C: static[int], T](fill: T): Matrix[R, C, T] {.noinit.} = discard

else:
  # Workaround: add an arbitrary concept constraint to every overload.
  type
    Additive = concept
      func `+`(a, b: Self): Self

  proc matrix*[R, C: static[int], T: Additive](elems: array[R, array[C, T]]): Matrix[R, C, T] {.noinit.} = discard
  proc matrix*[R, C: static[int], T: Additive](fill: T): Matrix[R, C, T] {.noinit.} = discard


let data: array[2, array[2, float64]] = [
  [1.0'f64, 2.0'f64],
  [3.0'f64, 4.0'f64],
]

let m = matrix(data)  # should resolve to the array-overload of `matrix`
discard m
```

The two `matrix` overloads share an unconstrained type parameter `T` and generic return type `Matrix[R, C, T]`. Inferring the return type seems to prevent `R` and `C` from being deduced from an `array[R, array[C, T]]` argument.

Interestingly, the problem goes away after adding an arbitrary - even unrelated - type constraint to `T`.

gmpreussner:
Fixes https://github.com/nim-lang/nimony/issues/2442

### DIFF
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -70,6 +70,7 @@ type
                       ## direction only makes `sigmatch` MISS a rejection —
                       ## never invent one — and `buildTypeArgs` still catches it
                       ## the way it always did.
+    unboundStaticTvars: int ## like `unboundTvars`, but only static parameters
     fn*: FnCandidate
     args*, typeArgs*: TokenBuf
     err*, flipped*: bool
@@ -109,7 +110,10 @@ proc bindTypevar(m: var Match; fs: SymId; a: Cursor) =
   ## `InferActualTypevar` — are bound but not counted, exactly as
   ## `matchTypevars` did not count them.
   m.inferred[fs] = a
-  if fs in m.tvars: dec m.unboundTvars
+  if fs in m.tvars:
+    dec m.unboundTvars
+    if isStaticTypevar(fs):
+      dec m.unboundStaticTvars
 
 proc addMissingConstraint*(m: var Match; routine: Cursor) =
   let key = asNimCode(routine, {renderNoBody})
@@ -789,7 +793,10 @@ proc adoptInferred(m: var Match; probe: Match; adopted: var seq[SymId]) =
 proc undoInferred(m: var Match; adopted: openArray[SymId]) =
   for k in adopted:
     m.inferred.del k
-    if k in m.tvars: inc m.unboundTvars
+    if k in m.tvars:
+      inc m.unboundTvars
+      if isStaticTypevar(k):
+        inc m.unboundStaticTvars
 
 type
   ConceptProbe = object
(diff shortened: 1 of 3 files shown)


gmpreussner:
## Root cause

This wasn’t a failure to infer `R`/`C` from `array[R, array[C, T]]`. That works fine when the array overload is chosen.

The bug was in **overload resolution** in `cmpMatches` (`sigmatch.nim`). When comparing two successful matches, Nimony preferred the candidate with **fewer** entries in `m.inferred`:

- `matrix(elems: array[R, array[C, T]])` binds `R`, `C`, and `T` → 3 entries
- `matrix(fill: T)` binds `T` to the whole array → 1 entry

The catch-all overload won, leaving `R` and `C` unbound, and `buildTypeArgs` then reported `could not infer type for R`.

The `Additive` workaround in `matrix.nim` worked because it **rejected** the `fill: T` overload, not because it helped array-length inference.

## Fix

In `sigmatch.nim` `cmpMatches`, rank by **`unboundTvars` first** (fewer wins), then fall back to the existing `inferred.len` tie-breaker.

Added regression test: `tests/nimony/overload/tstaticinferoverload.nim`

src/nimony/sigmatch.nim
```diff
@@ -2713,6 +2713,14 @@ proc cmpMatches*(a, b: Match; preferIterators = false): DisambiguationResult =
...
     elif aInh > bInh:
       return SecondWins
+    # Prefer the candidate that leaves fewer of the routine's own generic
+    # parameters unbound. Binding `T` to the whole argument in `foo(x: T)`
+    # must not beat `foo(x: array[N, T])` just because it records fewer
+    # entries in `inferred` while orphaning the static `N`.
+    if a.unboundTvars < b.unboundTvars:
+      return FirstWins
+    elif a.unboundTvars > b.unboundTvars:
+      return SecondWins
```
Araq: But why are there matches with unboundTvars at all...

Araq:
This is more of the same, some adhoc filtering. But a match with unbound generic parameters is not a match at all, the matching process should fail for these. Static-T or not does not matter.

Araq:
Succeeded by my PR.

PR #2453, opened 2026-09-04
sigmatch: prefer concept requirements on scope ties in generic bodies

### LINKED ISSUE #2452: Overload resolution fails when concept-constrained func calls unconstrained generic func
## Minimal Repro
```nim
type
  Foo* = concept
    func bar(x: Self)

func bar[T](x: T) = discard
func doBar[T: Foo](x: T) = bar(x)
```

## Expected Result
Compilation succeeds without errors.

## Observed Result

The `bar(x)` call in `doBar` generates the following error:
```
Error: ambiguous call: 'bar'
proc (x: T) (declared in tests/tmp/foo1.nim(5, 1))
func (x: Self) (declared in tests/tmp/foo1.nim(3, 5))
```

## Notes

The concept stub appears as a candidate because Nimony injects concept requirement signatures into the overload set when resolving calls involving a type variable constrained by that concept (`T: Foo`). That’s why the compiler lists the declaration from line 5 as a real candidate, not just a static check. This is expected behavior.

The problem here is that the two resulting candidates have the same priority, so Nimony cannot disambiguate them.

gmpreussner:
Model concept requirement stubs as ConceptScope so they beat same-module and imported overloads when all else ties, deferring the call to instantiation. Add regression test and document the scope hierarchy.

Fixes https://github.com/nim-lang/nimony/issues/2452

### DIFF
--- a/doc/language.md
+++ b/doc/language.md
@@ -1189,8 +1189,9 @@ No new ranking category is introduced; the existing rules do the work:
 
   func doBar[T: Foo](x: T): int = bar(x)
 
-  doBar(1.0'f64) # 2, not ambiguous: deferred at definition time,
-                 # then `bar(x: float64)` wins at instantiation
+  doBar(1.0'f64) # 2, not ambiguous: inside `doBar`, `bar` is bound to the
+                 # Foo concept requirement (ConceptScope wins the tie), which
+                 # defers resolution; at instantiation, `bar(x: float64)` wins
   ```
 
   ```nim
--- a/doc/language.md
+++ b/doc/language.md
@@ -1143,15 +1143,26 @@ Conceptually, every routine has one extra implicit parameter of type
 `Scope`. The compiler synthesizes the corresponding argument at every
 call site:
 
+- when the routine is a **concept requirement** — a phantom declaration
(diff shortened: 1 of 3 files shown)


doc/language.md
```diff
@@ -1162,7 +1173,25 @@ category is introduced; the existing rules do the work:
...
+  type
+    Foo = concept
+      func bar(x: Self): int
+
+  func bar[T](x: T): int = 1
+  func bar(x: float64): int = 2
+
+  func doBar[T: Foo](x: T): int = bar(x)
+
+  doBar(1.0'f64) # 2, not ambiguous: deferred at definition time,
```
Araq: Misleading comment. The `bar` that is bound in `doBar` is the one coming from the concept which then enablees the deferral at instantiation time.

gmpreussner: How about this?
```
  doBar(1.0'f64) # 2, not ambiguous: inside `doBar`, `bar` is bound to the
                 # Foo concept requirement (ConceptScope wins the tie), which
                 # defers resolution; at instantiation, `bar(x: float64)` wins
```

gmpreussner: I can also remove it altogether if the semantics seem clear

PR #2455, opened 2026-09-04
sem: merge non-routine imports for lone local routines under InnerMost

### LINKED ISSUE #2454: Overload resolution between proc and enum fails if declared in separate modules
The following is fine:
```nim
# foo.nim
type
  Foo = enum
    foo
    bar

func foo() = discard
func test(f: Foo) = discard

test(foo)  # foo is an enum value
```

## Minimal Repro

The following fails with Nimony:
```nim
# foo1.nim
import foo2

func foo() = discard
func test(f: Foo) = discard

test(foo)  # foo is now interpreted as a proc
```
```nim
# foo2.nim
type
  Foo* = enum
    foo
    bar
```

Classic Nim sees both candidates and disambiguates by the expected type at the call site (or reports "ambiguous identifier" without one).

## Observed Result

`foo` is matched to the `proc` instead of the enum value even though `test` takes an enum parameter:
```
Error: Type mismatch at [position]
test(foo)
[1] expected: Foo but got: proc () (declared in tests/tmp/foo1.nim(9, 1))
```

## Workaround
Qualify the enum value at the callsite:
```nim
test(Foo.foo) 
```

gmpreussner:
When InnerMost finds a single local routine, also consider imported non-routine bindings (enum fields, consts, types) so expected-type disambiguation can pick the right symbol. Imported routines are still shadowed.

Fixes https://github.com/nim-lang/nimony/issues/2454

### DIFF
--- a/src/nimony/sembasics.nim
+++ b/src/nimony/sembasics.nim
@@ -30,7 +30,7 @@ template buildTree*(dest: var TokenBuf; kind: StmtKind|ExprKind|TypeKind|SymKind
   dest.addParRi()
 
 proc considerImportedSymbols(c: var SemContext; dest: var TokenBuf; name: StrId; info: NifLineInfo;
-                             option = FindAll): int =
+                             option = FindAll; skipRoutines = false): int =
   result = 0
   let ignoreStyle = IgnoreStyleFeature in c.features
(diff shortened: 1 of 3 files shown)


Araq:
Feels adhoc. Must think it through.

Issue #2460, opened 2026-09-05
Feature: Add support for folding values involving static[T] parameters


gmpreussner:
The following expressions already work:
```nim
# basic static[T] arithmetic is supported
type Storage[R, C, T] = object
  data: array[R * C, T]
```
```nim
 # literal values are supported
type Storage[T] = object
  data: array[binomial(4, 2), T]  
```
```nim
# sizeof, low, high etc. are built-in "magics" that already fold
type Storage[T] = object
  data: array[sizeof(T), T] 
```

For a wide range of mathematical and other use cases, as well as backward compatibility with Nim, it would be nice to support `static[T]` parameters in user-defined functions for value expressions, for example:
```nim
type Storage[N, K, T] = object
    data: array[binomial(N, K), T]
```

PR #2461, opened 2026-09-05
WIP: Support user-defined compile-time calls in static generic type positions

gmpreussner:
Generic type bodies and routine signatures could already defer symbolic array bounds built from built-in operators, e.g. `array[M * N, T]` with `M, N: static[int]`, and fold them when the type is instantiated or when overload resolution compares array lengths. The same deferral did not work for user-defined routines such as `array[bump(N), T]` or `array[bump(R * C), T]`.

Changes:
- phase_validator: add `"call"` to `typeCtxTagsLiteral`, treating deferred routine applications in type slots like other compile-time value expressions (`mul`, `add`, …).

- sigmatch: extend `foldValueExpr` with top-level fallbacks (depth == 0) that run after the fast literal/operator path:
    1. `lengthOrd` for `rangetype` lengths
    2. `evalOrdinal` for concrete expressions, including user func calls
    3. `substituteTypevars` + `evalOrdinal` once static params are bound

  This keeps the logic generic — no array-specific helpers — while letting
  `matchArrayType` fold formal bounds such as `bump(R * C)` against
  concrete argument types.

Add tests/nimony/generics/tstaticfunc.nim covering:
- generic object types with unary and binary user funcs in array bounds
- arithmetic nested inside call arguments (`bump(R * C)`, `mix(R + 1, C * 2)`)
- proc signature matching with explicit static generic arguments
- range bounds built from a user func (`range[0 .. bump(N) - 1]`)

Instantiation-time folding in semArrayType/evalOrdinal was already sufficient for type definitions; the validator and sigmatch gaps were the missing pieces.

Implements https://github.com/nim-lang/nimony/issues/2460

### DIFF
--- a/src/nimony/sigmatch.nim
+++ b/src/nimony/sigmatch.nim
@@ -550,11 +550,11 @@ proc matchesConstraint*(m: var Match; f: var Cursor; a: Cursor): bool =
   result = matchesConstraintAux(m, f, a)
 
 proc foldValueExpr(m: var Match; a: Cursor; depth = 0): xint =
-  ## The tiny fixed-opcode evaluator for compile-time *values* in type
-  ## positions: folds `+ - *` over integer literals and rewrites an
-  ## array-index `rangetype` to its length. Anything else — in particular a
-  ## still-symbolic expression — yields NaN and is compared syntactically
-  ## instead; matching never solves for a value parameter backwards.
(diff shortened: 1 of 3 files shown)


gmpreussner:
We decided to not support user-defined functions in type parameters, so this would be a spec change.

We may retain this feature for backward compatibility with Nim, but gated behind a feature flag.

PR #2470, opened 2026-09-05
coro_transform: nil-compare fn-slot projection for any closure-value style

tokyovigilante:
The HconvX rewrite (closure tuple vs nil -> compare the fn slot) only matched Symbol/Tupat/Dot inners. A closure value can equally arrive as '(hderef (call get... seq i))' from an inlined seq-element loop -- callback lists hit exactly that -- leaving a cast-struct-to- pointer the C backend rejects:

```
    for hook in engine.hooks:
      if hook != nil: hook(x)      # error: cannot convert to a pointer type
```

Gate on the destination being a pointer cast and let typenav type ANY Symbol or tag inner; unknown shapes type as auto and fall through unchanged.

On nifcore a node is a TagLit and there is no ParLe kind, so the check is `inner.isTagLit`; an `inner.kind == ParLe` test is silently never true there, which is how the first version of this change passed a rebase without doing anything.

Adds the regression test (both the iterator element and the indexed form, single module); a let-bound copy of the element already worked.


src/hexer/coro_transform.nim
```diff
@@ -2538,7 +2538,13 @@ proc coroTr*(c: var Context; dest: var TokenBuf; n: var Cursor) =
...
         var dstType = inner
         skip inner           # past target type
-        if inner.kind == Symbol or inner.exprKind in {TupatX, DotX}:
+        # Any value shape can carry the closure tuple — a symbol, a
+        # field, but also `(hderef (call get… seq i))` from an inlined
+        # seq-element loop. Gate on the DESTINATION being a pointer cast
+        # and let typenav decide; unknown shapes type as auto and fall
+        # through.
+        if dstType.typeKind in {PtrT, PointerT} and
+            (inner.kind == Symbol or inner.isTagLit):
```
Araq, line 2547: This logic never made any sense to begin with. It's the comparator's job to compare closure values properly.

Araq:
It was a hack before, now it's a worse hack. Do it properly.

tokyovigilante:
Pulled in favour of #2475.

PR #2475, opened 2026-09-05
Compare closure values properly, instead of casting them to a pointer

tokyovigilante:
Supersedes #2470. Three commits; the first and third are prerequisites for the second.

**The problem.** There is no `==` for proc types at all, so `h == nil` on a closure   matches `==(x, y: nil pointer)` and sigmatch coerces the whole value to `pointer`. A  `.closure` value is an (fn, env) pair, so the cast can't typecheck. `coro_transform` compensated by intercepting the conversion and matching the operand against a list of value shapes (`Symbol`, `TupatX`, `DotX`) to project the fn slot — a list that grew every time a closure turned up in a new shape. #2470 widened it again.

**1. `fix(hexer): the proc typeclass on a value parameter crashed procHasPragma`**

`proc eqp[T: proc](a: T): bool` crashed before instantiation — the bare typeclass is `(proctype)` with no children, so `procHasPragma` walked off the end of the node and `extractPragma`'s `into` asserted. Guarded in `procHasPragma`, not `extractPragma`: a cursor past the end of a node is a bug everywhere else, and swallowing it there would hide the next one. Only the value-parameter form was affected, which is why `ttypeclassmatches` (typedesc) never caught it.

**2. `fix(sem): give proc values a comparator instead of casting them to pointer`**

`==` over the `proc` typeclass, magic `EqProc`, mapped like every other Eq magic to the typed `EqX`. `desugar` lowers `(eq|neq <closure> a b)` to a comparison of both the fn and env slots — the same function over a different environment is a different closure, and against `nil` it still reads correctly. The type is on the node, so nothing inspects operand shape, and the `coro_transform` block is deleted (`ConvX`/`HconvX` fold back into ordinary recursion, −44 lines there).

**3. `fix(hexer): the iterator typeclass crashed trProctype, and closure iters get a comparator`**

Same root cause as 1, one layer along: `(itertype)` with no children was classified as a closure iterator and walked to its params tag, past the end (`nifcore.nim:795`). With `[T: iterator]` available, closure iterators get the same comparator and
  `isClosureValueType` covers them — so `tryClosurePtrConv`, the last conversion-lowering path, is **deleted**: 49 lines replaced by a 9-line predicate. `HconvX`/`ConvX` are now ordinary nodes handled by `trSons`. A `.passive` iterator is excluded by the type predicate, not a special case: it lowers to a bare function pointer.

**Tests.** `tproctypeclass` (plain proc, `.closure`, iterator; one- and two-param) and `tclosure_seq_elem_nil` (from #2470, passing with the hack deleted). Both mutation-checked: the third guard is emphatic at 5/31, since the comparators live in the stdlib.

`tdistinctinfos.msgs` gains two lines: the new overloads as rejected candidates. The comparators are declared last in `comparisons.nim` so the errmsgs line citations don't shift.

Suite: **822/822**.


src/nimony/nimony_model.nim
```diff
@@ -477,6 +477,9 @@ proc skipToReturnType*(n: var Cursor) =
 proc procHasPragma*(typ: Cursor; kind: PragmaKind): bool =
   var typ = typ
   if typ.typeKind in RoutineTypes:
+    # A childless routine type is the bare `proc` typeclass, as written in
+    # `[T: proc]`: nothing to walk to, and no pragmas to find.
+    if cursorJump(typ) == 0: return false
```
Araq:
```suggestion
    if not hasMore(typ): return false
```

src/hexer/cps.nim
```diff
@@ -332,7 +332,10 @@ proc trProctype(c: var Context; dest: var TokenBuf; n: var Cursor) =
     # proctype function pointer, NOT a tuple.
     let isClosureIterType = nk == ItertypeT and not procHasPragma(n, PassiveP)
     let isPassiveIterType = nk == ItertypeT and procHasPragma(n, PassiveP)
-    if isPassiveProc or isClosureIterType or isPassiveIterType:
+    # A childless routine type is the bare typeclass, as written in `[T: proc]`
+    # / `[T: iterator]`: no signature to lower, so copy it through.
+    let isTypeclass = nk in {ProctypeT, ItertypeT} and cursorJump(n) == 0
```
Araq:
```suggestion
    let isTypeclass = nk in {ProctypeT, ItertypeT} and not hasMore(n)
```

src/hexer/desugar.nim
```diff
@@ -1292,17 +1338,33 @@ proc tr(c: var Context; dest: var TokenBuf; n: var Cursor; isTopScope = false) =
         genStringConcatChain(c, dest, n)
       else:
         trSons(c, dest, n)
-    of AddX, SubX, MulX, DivX, NegX, EqX, LeX, LtX:
+    of HconvX, ConvX:
+      trSons(c, dest, n)
```
Araq:
```suggestion
```

src/hexer/desugar.nim
```diff
@@ -1292,17 +1338,33 @@ proc tr(c: var Context; dest: var TokenBuf; n: var Cursor; isTopScope = false) =
...
         InfX, NeginfX, NanX, FalseX, TrueX, AndX, OrX, XorX,
         NotX, SizeofX, AlignofX, OffsetofX, OconstrX,
         AconstrX, BracketX, CurlyX, CurlyatX, OvfX,
         ModX, ShrX, ShlX, BitandX, BitorX, BitxorX,
-        BitnotX, NeqX, CastX, ConvX,
+        BitnotX, CastX,
         CchoiceX, OchoiceX, PragmaxX, QuotedX, HderefX,
         HaddrX, NewrefX, NewobjX, TupX, TupconstrX, TabconstrX,
-        AshrX, BaseobjX, HconvX, DconvX,
+        AshrX, BaseobjX, DconvX,
```
Araq:
```suggestion
        AshrX, BaseobjX, DconvX, HconvX, ConvX,
```

PR #2499, opened 2026-09-10
Dns resolver

blackmius:
added UdpSocket, asyncio file api, dns server and resolver


lib/std/ioring/backends/iouring.nim
```diff
@@ -69,6 +92,40 @@ proc fillSqe(sqe: ptr Sqe; op: ptr OpContext) {.inline.} =
...
+    # `IORING_OP_OPENAT` is the ring's own open: nothing waits on it and this
+    # backend never makes the syscall itself — the only syscalls an io_uring
+    # backend may make are the io_uring ones. The path is the caller-owned
+    # buffer carried in the op context; it stays alive until the op completes,
+    # the same contract `submitRead`'s buffer has.
+    discard sqe.openat(AtFdCwd, cast[pointer](op.buf), cint(op.openFlags), op.openMode)
+  of opRecvFrom:
+    # IORING_OP_RECVMSG has no recvfrom form: the source address travels in the
+    # msghdr (`msg_name`), which the kernel reads on submission and writes back
+    # into at completion, so it lives in the per-slot `gMsgs` arena — never on
```
blackmius, line 105: i am not sure ioring is a right place to provide msg arena, it looks like it should live in socket.nim, need some research how to orginize it properly

lib/std/ioring/core/types.nim
```diff
@@ -147,15 +148,23 @@ type
     fd*: FileHandle
     result*: int
       ## Op-dependent: a byte count for `opRead`/`opWrite`, the accepted fd for
-      ## `opAccept`, -1 on failure — and for `opPollAdd` the fired directions
-      ## encoded as a bit mask, which `readyEvents` decodes into `IoEvents`.
+      ## `opAccept`, the opened fd for `opOpen`, -1 on failure — and for
+      ## `opPollAdd` the fired directions encoded as a bit mask, which
+      ## `readyEvents` decodes into `IoEvents`.
 
   OpContext* = object
```
blackmius, line 155: it need to be a sum type

lib/std/asyncio.nim
```diff
@@ -0,0 +1,475 @@
...
+    stdcall, importc: "SetFilePointerEx", dynlib: "kernel32", sideEffect.}
+
+  proc seekImpl(h: OsFileHandle; off: int64; whence: cint): int64 {.raises.} =
+    var np: int64 = 0
+    if setFilePointerEx(h, off, addr np, DWORD(whence)) == WINBOOL(0):
+      raise IOError
+    result = np
+
+else:
+  proc posixOpen(path: cstring; flags: cint; mode: Mode): cint {.importc: "open", sideEffect.}
```
blackmius, line 176: syscalls are not used here

lib/std/asyncio.nim
```diff
@@ -0,0 +1,475 @@
...
+    of fmReadWrite:
+      desiredAccess = GENERIC_READ or GENERIC_WRITE; disposition = CREATE_ALWAYS
+    of fmReadWriteExisting:
+      desiredAccess = GENERIC_READ or GENERIC_WRITE; disposition = OPEN_EXISTING
+    of fmAppend:
+      # FILE_APPEND_DATA makes every write land at end-of-file — the Win32
+      # counterpart of POSIX `O_APPEND` (no initial seek needed).
+      desiredAccess = FILE_APPEND_DATA; disposition = OPEN_ALWAYS
+    var tmpFilename = filename
+    result = createFileW(newWideCString(tmpFilename).toWideCString,
```
blackmius, line 148: why it does not use ioring at all?

lib/std/dns.nim
```diff
@@ -0,0 +1,458 @@
...
+      seg = 0
+      val = 0
+    else:
+      return false
+  result = dots == 3 and seg > 0
+
+# ------------------------------------------------------------------ client ---
+
+type
+  Resolver* = object
```
blackmius, line 287: module should have defaultResolver relay, and likely for that we also need something like golang sync.Once so it will look like this
```nim
var gDefaultResolver: Resolver
var gInitResolverOnce: Once[proc(): Resolver {.passive.}]
proc initDefaultResolver(): Resolver {.passive.} =
  if gDefaultResolver != nil:
    return gDefaultResolver
  return gInitResolverOnce.run(proc() {.passive.} =
    var resolver: Resolver
    resolver.initResolver() # passive reads /etc/hosts /etc/resolv.conf /etc
    gDefaultResolver = resolver
    return gDefaultResolver
  )
proc resolve(host: string): peerAddr {.passive.} =
  var resolver = initDefaultResolver()
  return resolver.resolve(host)
```

blackmius: maybe it is easier to have thread local resolver

lib/std/ioring.nim
```diff
@@ -391,6 +447,24 @@ when defined(posix):
     setNonBlocking(fd)
     result = fd
 
+  proc createUdp*(port: uint16): cint =
+    ## A non-blocking UDP socket bound to the wildcard address on `port`;
+    ## `0` asks the kernel to pick one (`boundPort` reports it), mirroring
+    ## `listenTcp` but for datagrams — there is no listen, so what parks on
+    ## the ring is the send/recv, not a connection queue.
+    let fd = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP)
```
blackmius, line 455: socket/setsockopt is just a syscalls so createUdp createSocket can live in socket.nim and ioring just provide submitSocket submitSetSockOpt

lib/std/dns.nim
```diff
@@ -0,0 +1,458 @@
...
+
+proc add*(s: var DnsServer; name, ip: string) =
+  ## Serve `name` from `ip` (a dotted quad). Questions are matched
+  ## case-insensitively and without regard to a trailing dot.
+  assert isDots4(ip), "dns: server addresses must be dotted quads"
+  s.names[canon(name)] = ip
+
+proc boundPort*(s: var DnsServer): uint16 {.inline.} = boundPort(s.sock)
+
+proc serveOnce*(s: var DnsServer; dl = never) {.passive, raises.} =
```
blackmius: it should work like httpserver it loops forever in server.next() to receive a request and then have a response function that will pack data passed and send udp datagram back to requester
``` nim
var udpServer = newDnsServer(53)
var r: DnsRequest
while udpServer.next(r):
  for q in r.questions:
    q.answer(name=q.name, type=q.type, ttl=30, data="127.0.0.1")
    q.answer(name=q.name, type=q.type, ttl=30, data="127.0.0.2")
  r.respond()
```

lib/std/dns.nim
Araq: Use a lock, no need for fancy `Once`.

lib/std/ioring/backends/iouring.nim
blackmius: i have been thinking and since msg headers is only io_uring specific zero copy optimisations it should be in io_uring backend, poll backends does not use it

lib/std/ioring/core/types.nim
blackmius: variadic type fits better with minimal changes

lib/std/ioring/backends/iocp.nim
```diff
@@ -468,13 +566,31 @@ when defined(windows):
         gSlotAux[lane][slotIdx] = NoAux
         armDeadline(lane, slotIdx)
         # An fd-less op (nop, timer) has no socket to associate, and a readiness
-        # probe is served by WSAPoll rather than by the port.
-        if buf[i].kind != opNop and buf[i].kind != opTimeout and
-            buf[i].kind != opPollAdd and not ensureAssociated(buf[i].fd, lane):
-          complete(slotIdx, ECancelled) # closed or foreign handle: never issued
-        else:
+        # probe is served by WSAPoll rather than by the port. A FILE op (opOpen,
```
blackmius, line 569: well it worked somehow before

blackmius: i think i will drop windows support as i dont know where to run it

Araq: Just let me handle the windows stuff but don't remove it.

PR #2510, opened 2026-09-12
contracts: check `.requires` at compile time

Araq:
A `.requires` is discharged at the *call site*: the contract is read with each parameter standing for the argument written there, and handed to the same `inferle` fact base every other obligation uses. Three answers, and only the middle one is silent:

* proven — nothing more happens, the contract cost nothing;
* undecided — the run-time guard hexer emits into the callee stays;
* violated (the facts imply the *negation*) — a compile-time error.

`proveFact` is what keeps the last two apart; conflating "not proven" with "false" would make `not` unsound. `{.feature: "staticContracts".}` demands the proof of every call site in a module, `{.feature: "runtimeContracts".}` judges none and wins if both are given.

The old `checkReq` was dead code behind `when isMainModule` and a hello-world proved 0 of its 261 contract sites. What makes the ordinary shapes provable:

* Derived locations. A `VarId` was always a `SymId`; `s.len` is a location now, allocated from the negative range. A *transparent accessor* is looked through, so `len(s)` and the `s.len` written inside `seqimpl` are one location; a pre-pass indexes the module first, because a generic instance is created here rather than imported.
* `for i in 0 ..< s.len` takes the loop variable's range from the iterator's own `.ensures`, which `finalir.trFor` turns into `(assume …)`. Inline iterators are not inlined until hexer, so the range has to be stated.
* Loop induction. `scanLoopWrites` classifies every location a body writes as increasing/decreasing/preserves-non-negative, and only the facts no such movement can break survive: in `a <= b + c`, `a` may only fall and `b` may only rise. `inc`/`dec` are recognised by shape, not name, and *shift* what is known rather than erasing it.
* `if a and b:` and `if a or b:` arrive as one materialized boolean; both sides of the diamond are recorded, which is also what `assert a and b` needs.

`Natural` and `Positive` are real range types now, as the TODO in `system.nim` asked. A range-typed parameter or field *states* `0 <= i`, so that half of every index obligation is discharged by the type — no guard and no `.requires` at any call site. Four things were needed to make that land: `semConvArg` looks through a range to its base (`uint64(x)` with `x: Natural` is an ordinary integral conversion again); an obligation against a bound that *is* the base type's own is vacuous; `checkRangeAssign` answers `a - b`, `a + b`, `a shr k` and `a and k` structurally; and a declared range is consulted when the flow facts have been widened away, since no write can invalidate one.

Bugs this turned up, each of which could have reported a *correct* index as a violation:

* `rightHandSide` ignored `add` vs `sub`, so `i <= s.len - 2` was recorded as `i <= s.len + 2`;
* `not (a == b)` was modelled as `a >= b + 1` and "proved" `x > 3` from `if x != 3` — a disequality is not representable, and what knowledge there is sits on the false path;
* a `for` over an un-inlined iterator had no exit label, so everything after it was unreachable — and a join on a dead path keeps *both* arms of the next `if`, so `le >= 0` and `le < 0` held at once;
* `traverseLoop` invalidated what the body writes *after* the checkpoint the rollback restores;
* a `var` parameter forwarded to another takes no `(haddr …)` and so got no `(unknown …)`: `proc f(s: var string) = s.setLen(n); s[0] = 'x'`;
* a rejected initializer seeded facts, and `var a: range[0..10] = 20` left a contradiction from which everything was provable;
* `semProposition` clobbered `c.routine.resId` — invisible for procs, which re-declare it, fatal for iterators;
* `template default*[T: range]` was missing outright, so an object with a `Natural` field could not be `default`ed.

`sigmatch.orderArgs` and `anArgIsStillUntyped` genuinely fell through without initializing `result`; the dead-code bug had been hiding it.

Measured over the whole standard library: 560 proven of 4645 obligations before, 1636 of 5074 after, and the unique source sites that cannot be proven fall from 839 to 758 — of which only 8.6 % (was 20.5 %) are missing the lower bound. Strict is not the default yet: flipping it fails 116 of 818 test files, blocked now by length-tracking through construction rather than by loops. `newSeq[T](n)` cannot promise `len == n`; the OOM path sets it to zero.


src/finalir/finalir.nim
```diff
@@ -633,6 +639,176 @@ proc extractForBorrow(c: var Context; forStmt: ForStmt; info: NifLineInfo): Toke
...
+  var fn = call
+  fn = sub(fn)
+  if not fn.isSymbol: return
+  let sym = tryLoadSym(fn.symId)
+  if sym.status != LacksNothing: return
+  if not isRoutine(sym.decl.symKind): return
+  let r = asRoutine(sym.decl)
+  if cursorIsNil(r.pragmas): return
+  let ens = extractPragma(r.pragmas, EnsuresP)
+  if cursorIsNil(ens) or ens.exprKind != ExprX: return
```
Araq: Adhere to agents.md, use structured programming.

src/finalir/finalir.nim
```diff
@@ -633,6 +639,176 @@ proc extractForBorrow(c: var Context; forStmt: ForStmt; info: NifLineInfo): Toke
...
+    dest.shrink mark
+    return
+  skip r
+  if not emitContractOperand(dest, r, subst, info, 0):
+    dest.shrink mark
+    return
+  dest.addParRi() # close the comparison
+  dest.addParRi() # close `assume`
+
+proc forRangeAssumes(c: var Context; forStmt: ForStmt; info: NifLineInfo): TokenBuf =
```
Araq:
```suggestion
proc forRangeAssumes(c: var Context; dest: var TokenBuf; forStmt: ForStmt; info: NifLineInfo) =
```

src/finalir/finalir.nim
```diff
@@ -641,10 +817,19 @@ proc trFor(c: var Context; dest: var TokenBuf; n: var Cursor) =
   let forStmt = asForStmt(n) # peek at structure before advancing
   let forStart = n
   n = sub(n)
-  let borrowBuf = extractForBorrow(c, forStmt, info)
+  var borrowBuf = extractForBorrow(c, forStmt, info)
+  borrowBuf.add forRangeAssumes(c, forStmt, info)
```
Araq:
```suggestion
  forRangeAssumes(c, borrowBuf, forStmt, info)
```

src/nimony/contracts_fir.nim
```diff
@@ -1343,6 +2436,97 @@ proc leaveToReturn(c: var FirContext) = gotoReturn(c.tr, c.flow)
...
+      var tv = t
+      tv = sub(tv)
+      let v = tv
+      skip tv
+      if extractSymId(tv) == target:
+        value = v
+        result = true
+    skip t
+
+proc diamondTarget(thenArm, elseArm: Cursor; value: var Cursor;
```
Araq: Oh god no, just fix `xelim` / finalir instead not to emit this nonsense anymore.

PR #2534, opened 2026-09-17
Pr/macro plugins outside repo

tokyovigilante:
A macro declared in one module and used as a proc pragma in another — the `{.async.}` shape every async or effect library takes — fails to compile from any project directory outside this repo:

```
Error compiling macro plugin for 'traced':
[Error] cannot open: <nimcache>/<mod>.s.deps.nif
```

Two things conspire, one commit each:

1. **The outer build and the plugin sub-compile disagree on module suffixes.** The sub-compile always injects `--path:<repo>/src/lib` (nifbuilder/nifreader, pulled in by `std/macros`); the outer build only had it when the caller's `nimony.paths` happened to list it — which is why every suite under `tests/nimony/macros` carries `../../../src/lib` in its paths file. `semos.setupPaths` now adds the compiler-internal `src/lib` unconditionally, so the two agree everywhere.

2. **An imported macro is absent from the importer's `compiledMacros`** — its declaration is never re-semchecked there — so the plugin the dependency build already compiled was not found. `macro_plugin` gains an on-disk lookup for that case, an isolated host-plugin nimcache (`<name>.host/`), and strips any `--bits:NN` the outer compile carries: a plugin is always a host-native executable, and forwarding a 32-bit target (as the wasm harness does) fails the plugin's own pointer-size assert.

The new test `tests/nimony/plugins/timportedmacro.nim` (+ `deps/mimportedmacro.nim`) is the shape above with an identity macro. It lives in the `plugins` suite on purpose: that suite's `nimony.paths` does not list `src/lib`, so the test fails on master with the error above and passes with these commits. Existing suites unaffected.


src/nimony/macro_plugin.nim
```diff
@@ -302,6 +346,31 @@ proc compileMacroPlugin*(nifcachePath: string; macroDecl: Cursor; macroSym: SymI
...
+
+  # Pre-populate the isolated plugin cache with a HOST-bits stdlib. `nimony s`
+  # only READS its imports' `.s.nif` — it does not build them — so in a fresh
+  # per-plugin cache we must first materialise the stdlib the plugin imports.
+  # The plugin scaffold imports exactly `std/[syncio, macros]` (see
+  # `emitImportStdMacros`); seeding those pulls in the whole host-bits stdlib
+  # closure the plugin needs (incl. the NimNode/NIF-reader machinery). The
+  # native link of this seed may fail (harmless — we only need the `.s.nif`/
+  # `.c.nif`), and `nimony c` is incremental so repeat calls are cheap.
+  let seedFile = pluginCache / "macro_seed.nim"
```
Araq:
```suggestion
  let setupFile = pluginCache / "macro_setup.nim"
```

src/nimony/macro_plugin.nim
```diff
@@ -302,6 +346,31 @@ proc compileMacroPlugin*(nifcachePath: string; macroDecl: Cursor; macroSym: SymI
...
+  try:
+    writeFile(seedFile, "import std/[syncio, macros]\n")
```
Araq:
```suggestion
    writeFile(setupFile, "import std/[syncio, macros]\n")
```

src/nimony/macro_plugin.nim
```diff
@@ -302,6 +346,31 @@ proc compileMacroPlugin*(nifcachePath: string; macroDecl: Cursor; macroSym: SymI
...
+  except:
+    echo "Macro plugin: failed to write ", seedFile
```
Araq:
```suggestion
    echo "Macro plugin: failed to write ", setupFile
```

src/nimony/macro_plugin.nim
```diff
@@ -302,6 +346,31 @@ proc compileMacroPlugin*(nifcachePath: string; macroDecl: Cursor; macroSym: SymI
...
+    return ""
+  let seedCmd = quoteShell(nimonyExe) & hostifyPluginArgs(commandLineArgs) &
```
Araq:
```suggestion
  let setupCmd = quoteShell(nimonyExe) & hostifyPluginArgs(commandLineArgs) &
```

src/nimony/macro_plugin.nim
```diff
@@ -302,6 +346,31 @@ proc compileMacroPlugin*(nifcachePath: string; macroDecl: Cursor; macroSym: SymI
...
+                " --path:" & quoteShell(srcLibPath) &
+                " --nimcache:" & quoteShell(pluginCache) &
+                " c " & quoteShell(seedFile)
```
Araq:
```suggestion
                " c " & quoteShell(setupFile)
```

src/nimony/macro_plugin.nim
```diff
@@ -302,6 +346,31 @@ proc compileMacroPlugin*(nifcachePath: string; macroDecl: Cursor; macroSym: SymI
...
+  try:
+    discard execCmdEx(seedCmd)   # ignore exit: a failed native link still leaves the .s.nif/.c.nif
```
Araq:
```suggestion
    discard execCmdEx(setupCmd)   # ignore exit: a failed native link still leaves the .s.nif/.c.nif
```

src/nimony/macro_plugin.nim
```diff
@@ -302,6 +346,31 @@ proc compileMacroPlugin*(nifcachePath: string; macroDecl: Cursor; macroSym: SymI
...
+  except:
+    echo "Macro plugin: failed to seed host stdlib: ", seedCmd
```
Araq:
```suggestion
    echo "Macro plugin: failed to setup host stdlib: ", setupCmd
```

src/nimony/macro_plugin.nim
```diff
@@ -270,6 +272,31 @@ proc getMacroPluginPath*(nifcachePath: string; macroSym: SymId): string =
   when defined(windows):
     result.add ".exe"
 
+proc hostifyPluginArgs(args: string): string =
+  ## A macro plugin is always a HOST-native executable (it is exec'd at compile
+  ## time), so it must be built with the host word size. Strip any `--bits:NN`
+  ## the outer compile carries — e.g. the JS test harness forces `--bits:32` to
+  ## short-circuit its own native link, but forwarding that to the plugin build
+  ## makes its C backend fail a pointer-size static assert. All other args
+  ## (notably `--cc`, for nifmake signature matching) are preserved.
```
Araq: That is a bug though, there is also `--cpu` and `--os` that are affected and must be removed too.

Araq:
hostifyPluginArgs  is a hack, instead the caller should provide the command line stripped off dangerous switches. For this the main argument forwarding logic needs to distinguish between "forward to ordinary child processes" vs "forward to compile-time-eval processes".

tokyovigilante:
Done in aa381acf. hostifyPluginArgs is gone: parseCommonOption now has a third out-flag, forwardArgHost, false for --cpu, --os and --bits, and nimony/nimsem build a hostCommandLineArgs line beside commandLineArgs from it (the wasm triplet is target-only). 

Apparently semos.runProgram/prepareEval (const-eval) would need their own cache if we move them to the host line, so have left that there for now.

PR #2538, opened 2026-09-17
ioring: submitPollRemove — cancel in-flight ops without closing the fd

tokyovigilante:
`closeFd` is the only way to cancel an armed op, and it closes the descriptor too. 

A library that owns its sockets and wants to stop watching one (libcurl on `CURL_POLL_REMOVE`, before it closes the fd itself) can only `forgetFd`, which cannot free a slot whose op is in flight: the completion never comes and the slot is gone for the life of the process.

    submitPollAdd(fd, {evRead})   # armed on the next poll
    # ...curl decides to close fd itself: nothing here may cancel it
    closeFd(fd)                    # cancels, but closes what curl still owns

`submitPollRemove(fd)` drains the lane, cancels every op on `fd` via `cancelPendingOps` (which now returns the count), drops the backend registration and leaves the fd open. 

The drain matters: an op is enqueued by `submitPollAdd` and only becomes a slot on the next poll, so cancelling without it finds nothing and the op arms a moment later. 

IOCP cannot cancel without closing (the kernel owns the OVERLAPPED until it acknowledges), so it drops the registration and returns 0. 

Test tpollremove: fd still readable after cancel, 0 when nothing armed, 2 for two ops; green on io_uring and epoll.


lib/std/ioring.nim
```diff
@@ -263,7 +263,7 @@ proc waitCompletions*(comps: var openArray[IoCompletion]): int =
     result = pollCompletions(comps)
     if result > 0: return
 
-proc cancelPendingOps(fd: cint) =
+proc cancelPendingOps(fd: cint): int {.discardable.} =
```
Araq:
```suggestion
proc cancelPendingOps(fd: cint): int =
```

PR #2539, opened 2026-09-17
contracts: a not-nil pointer has no default value, and neither has an…

Araq:
…ything holding one

Zeroed storage reads as `nil`, which is the one value a not-nil pointer rules out -- so `array[8, NotNilRef]` states eight non-nil refs and produces eight nils. Two checks in `contracts_fir`:

`isConvertedNil`, in `wantNotNil`: a `T(nil)` conversion no longer vouches for itself on the strength of the CONVERSION's type. That is how `system/defaults` spells a pointer's default (`template default[T: nil (ref)](x: typedesc[T]): T = T(nil)`), and the `nil (ref)` constraint binds nothing today, so `default(T)` for a not-nil `T` handed back a nil in silence -- `default(array[8, T])`, whose `T: HasDefault` asks precisely this question, included.

`lacksDefaultValue` + a check in `traverseLocal`: a GLOBAL declared without an initializer whose type contains a not-nil pointer is an error. A global carries no initialization proof -- `traverseExpr` tracks locals only, because a global is readable from another module and another thread -- so its storage is simply zeroed. A local owes nothing at its declaration: the flow analysis lets `var r: NotNilRef` stand as long as a write reaches it before any read. An `importc`'d global is exempt: it BINDS storage C already defines.

Errors are also attributed better. A template body carries its own file's line info and `subsGenericProc` gives an instantiated routine the call site as its head info, so an error inside expanded code used to name `lib/std/system/defaults.nim` at someone who wrote `default(array[8, Foo])`. `FirContext.stmtContext` follows the last statement written in the enclosing routine's own file, and `buildErr` moves an error from another file onto it, stating the expansion's origin in the message. Comparing against the enclosing CONTEXT rather than the module is what keeps an `include`d file reporting in its own name.

What the checks found, all of it a real not-nil violation:

- `typenav.TypeScope.parent`/`current`, `symtabs.Scope.up`, `semdata.SemRoutine.parent`: nilable chains -- the outermost link has none. `{.feature: "lenientnils".}`, as 15+ sibling modules already carry.
- `lengc/nifmodules`: `MainModule(...)` omitted its not-nil `config`.
- `sem`/`semcall`: `default(Match)` -- `Match` holds a not-nil `ptr SemContext`.
- `std/os`: `execShellCmd`'s `argv` needs `UncheckedArray[nil cstring]`; slot 3 is the NULL terminator `execve` demands.
- `std/ioring`: `gCancelInFlight` is documented "the readiness backends leave it nil" but was not typed `nil proc`; `BackendRelays.forgetFd` likewise (its call sites already nil-check it). `poll`/`close` are `unchecked`: the global cannot be filled at its declaration, since a backend's relays open kernel objects sized by `ioLanes()`.
- the tests: recursive tree links that are compared `== nil`, object constructors leaving a not-nil field nil, `newSeq`/`setLen` on a seq of not-nil refs (`shrink` is the unconstrained one).


doc/lenientnils.md
```diff
@@ -14,11 +14,65 @@ type
...
+default-initializable again:
+
+```nim
+type Seeded = object
+  n: Node = Node(data: 0)
+  count: int
+
+var s: Seeded           # fine
+```
+
```
Araq:
```suggestion
```

doc/lenientnils.md
```diff
@@ -83,6 +137,31 @@ if optCallback != nil:
   optCallback(42)
 ```
 
+An iterator TYPE — the type of a first-class closure-iterator value — is a
```
Araq:
```suggestion
An iterator type is a
```

doc/lenientnils.md
```diff
@@ -83,6 +137,31 @@ if optCallback != nil:
...
+says "this representation, on my head" and stands — the same standing `addr`
+has. That is how a NULL reaches C:
+
+```nim
+sqe.prepRw(OP_POLL_ADD, fd, cast[pointer](nil), 0, 0)  # liburing's NULL addr
+```
+
+Prefer `nil pointer` on the parameter where you can say it; reach for `cast`
+where the declaration is not yours to change.
+
```
Araq:
```suggestion
```

PR #2540, opened 2026-09-17
sem: fold deferred plugin calls in overload resolution

gmpreussner:
Teach expreval to evaluate parked `(pluginCall …)` nodes via executeExpr once static bindings make their arguments concrete, and delegate from foldValueExpr so array lengths like `array[twicePlugin(N), T]` match during sigmatch. Add tstaticparams coverage for plugin vs template overloads.


src/nimony/expreval.nim
```diff
@@ -1108,6 +1108,34 @@ proc evalImpl(c: var EvalContext; n: var Cursor): Cursor =
...
+        skip n
+      else:
+        var callBuf = createTokenBuf(16)
+        callBuf.addParLe(CallS, n.info)
+        var ch = n
+        inc ch # past the `pluginCall` tag; the template symbol and its args follow
+        while ch.hasMore:
+          callBuf.addSubtree ch
+          skip ch
+        callBuf.addParRi()
```
Araq, line 1127: Instead make `c.c.executeExpr` deal with `PluginCallX` and treat it as a call? But I haven't checked its implementation.

src/nimony/sigmatch.nim
```diff
@@ -585,6 +585,27 @@ proc foldValueExpr(m: var Match; a: Cursor; depth = 0): xint =
...
+      # still symbolic the call stays parked and is compared structurally.
+      if m.context == nil or m.context.executeExpr == nil: return
+      var subBuf = createTokenBuf(16)
+      substituteTypevars(subBuf, a, m.inferred)
+      var sub = cursorAt(subBuf, 0)
+      if containsGenericParams(sub): return
+      var ec = initEvalContext(m.context, noExecute = false)
+      # A length is an ordinal; give the sub-compile a concrete result type so it
+      # can serialise the folded value (an `auto` result cannot be serialised).
+      ec.expectedType = m.context.types.intType
```
Araq, line 604: How do you know its a `length`?!

tests/nimony/generics/tstaticparams.nim
```diff
@@ -166,10 +166,27 @@ echo ritem(rb, 2)                                     # 30
 # the same shape as an instance's and `system.high[I, T](x: array[I, T])` binds
 # `I` to a *type* rather than to the value `N` (nim-lang/nimony#2485).
 template twice(n: int): untyped = 2 * n
+template twicePlugin(n: int): untyped {.plugin: "deps/mtwice".}
```
Araq, line 169: Instead of guessing `int` in sigmatch, use `int` as the return type here and reject template return types that are untyped/auto in sigmatch.

PR #2544, opened 2026-09-18
io_uring: cqNeedsFlush wants either SQ flag, not both

tokyovigilante:
`{SQ_CQ_OVERFLOW, SQ_TASKRUN} <= flags` is a subset test, so it asked for both bits at once; liburing ORs them. Harmless with no setup flags, where TASKRUN rarely appears alone, but with DEFER_TASKRUN it is the normal signal, and GETEVENTS was never passed: every operation stalled.


lib/std/posix/io_uring.nim
```diff
@@ -837,9 +837,13 @@ proc sqNeedsEnter(queue: var Queue; submit: int; flags: var EnterFlags): bool =
...
 proc cqNeedsFlush(queue: var Queue): bool =
+  ## Whether the CQ needs an `io_uring_enter` with GETEVENTS first: the kernel
+  ## sets SQ_CQ_OVERFLOW when completions did not fit and SQ_TASKRUN when
+  ## deferred completion work is pending. Either flag alone is the signal, as
+  ## in liburing; with DEFER_TASKRUN, TASKRUN on its own is the normal case.
   var sqFlags = atomicLoad(cast[ptr uint32](queue.sq.flags)[], moRelaxed)
-  return {SQ_CQ_OVERFLOW, SQ_TASKRUN} <= cast[ptr SqringFlags](sqFlags.addr)[]
-  # {SqCqOverflow, SqTaskrun} <= atomic_load_explicit(queue.sq.flags, moRelaxed)
+  let f = cast[ptr SqringFlags](sqFlags.addr)[]
+  return SQ_CQ_OVERFLOW in f or SQ_TASKRUN in f
```
Araq:
```suggestion
  return {SQ_CQ_OVERFLOW, SQ_TASKRUN} * f != {}
```

PR #2554, opened 2026-09-20
nimony: one web back end binary for both `j` and `w`

Araq:
Pins `src/nativenif.commit` to 79d9e2a, where the wasm code generator `ithaqua` is folded into `src/web`: one code generator up to the web IR, two renderers behind it, and one binary — `jorogumo j` writes the `.js` program, `jorogumo w` the `.wasm` module. There is no second tool to build, pin or keep in step any more.

- deps.nim: the whole-program command is `jorogumo` for both backends, with the renderer as its first argument. `--target:browser` stays JS-only; jorogumo rejects it under `w`.
- builders.nim: `buildIthaqua` is gone, `buildJorogumo` is what `nimony j`, `nimony w` and `hastur wasmdiff` all ask for.
- tests/ithaqua/run_wasm.js takes upstream's synchronous `nim_write`, so a `nim_exit` cannot cut the pending bytes on a pipe.
- doc/internals/jorogumo-plan.md is deleted: it said it would be superseded on landing by the reference in nativenif, which is now `nativenif/doc/web.md`.


src/hastur/hastur.nim
```diff
@@ -29,6 +29,8 @@ Usage:
 
 Commands:
   build [all|nimony|nifler|nifler2|hexer|lengc|shoggoth|nifmake|validator|dagon|pnak|arkham|nifasm|jorogumo|native|nifbench]   build selected tools (default: all).
+                       `jorogumo` is the web back end — ONE binary serving both
+                       `nimony j` (JavaScript) and `nimony w` (wasm32).
```
Araq:
```suggestion
```

tests/ithaqua/wasmgaps/README.md
```diff
@@ -1,13 +1,13 @@
-# Wasm-backend (ithaqua) gap repros
+# Wasm-backend gap repros
 
-Each file here is a MINIMAL repro of a
-construct the WASM leg (`nimony w` -> ithaqua) cannot compile yet, while the
-native oracle handles it fine — so it is quarantined out of `tests/ithaqua/`
-(wasmdiff walks only the top level) until ithaqua grows the lowering.
+Each file here is a MINIMAL repro of a construct the WASM leg (`nimony w` ->
```
Araq:
```suggestion
Each file here is a minimal repro of a construct the WASM leg (`nimony w` ->
```

